PR #4, opened 2021-07-20
CVL-16: Add home page tiles for various top-level workflows

timharrison-moj:
Adds ability for a tiled home landing page.
Example top-level workflows added.

### DIFF
--- a/assets/js/card.js
+++ b/assets/js/card.js
@@ -1,0 +1,12 @@
+window.onload = function () {
+  // Loops through dom and finds all elements with card--clickable class
+  document.querySelectorAll('.card--clickable').forEach(card => {
+    // Check if card has a link within it
+    if (card.querySelector('a') !== null) {
+      // Clicks the link within the heading to navigate to desired page
+      card.addEventListener('click', () => {
+        card.querySelector('a').click()
+      })
+    }
+  })
+}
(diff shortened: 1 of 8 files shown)


assets/sass/components/_card.scss
```diff
@@ -0,0 +1,100 @@
+/* ==========================================================================
```
yasinmustafa, line 1: Is this 'card' component a pattern they use in other services?

assets/js/card.js
```diff
@@ -0,0 +1,12 @@
+window.onload = function () {
```
yasinmustafa, line 1: presumably they use this card component elsewhere and  it requires this jquery?

yasinmustafa (approved):
LGTM - A couple of questions

assets/sass/components/_card.scss
timharrison-moj: Yes - I stole it from book a video link, but its in several others.

assets/js/card.js
timharrison-moj: yes. Its a way of making the whole tile clickable, and for the click to follow the link defined in the card header.

timharrison-moj: Actually, just checking that, it doesn't. Either it doesn't do that or the script is not active. I'll dig a bit more in next PR.

PR #7, opened 2021-07-29
CVL-28: Community API remove non-Delius username from client token.

timharrison-moj:
Small, speculative change to test whether the username is important when connecting to the community api.
I suspect that where we supplied a username it will try to validate it as a Delius user, and fail.


yasinmustafa (approved):
lgtm

PR #11, opened 2021-08-02
CVL-12 - add a Gotenberg container for HTML template to PDF conversions

timharrison-moj:
* Configured the Gotenberg service as a generic-service container related to the UI service.
* Added Gotenberg and properties to the Helm deployment files
* Added configuration so the UI service can reference the new Gotenberg container
* Updated the README with all dependent services, Gotenberg and generated types.
* Tested running the service locally against the DEV environment APIs and created a `docker-compose-dev.yaml` for convenience.


yasinmustafa (approved):
LGTM

PR #12, opened 2021-08-02
CVL-2 to fix Gotenberg deployment as a generic service

timharrison-moj:
Correcting attribute names to match generic service expectations.


yasinmustafa (approved):
LGTM

PR #15, opened 2021-08-12
CVL-12 Render a pdf document from a HTML template and data using Gotenberg

timharrison-moj:
Includes:-

* Version updates for various npm packages
* Configuration to allow the service to find the Gotenberg container
* The Gotenberg API client
* A licence template for PDF's (very rough for now)
* SCSS styles for PDF printed view
* New licence routes to preview the HTML and render as PDF
* Some tests for the licence routes
* An integration test for producing PDF from HTML


server/routes/index.ts
```diff
@@ -27,6 +33,13 @@ export default function Index({ userService, prisonerService, licenceService, co
     get('/prisoner/:nomsId/image', otherAccessRoutes.getPrisonerImage)
   }
 
+  // Routes for creating a licence
+  const licenceRoutes = () => {
+    get('/licence/staffId/:staffId/caseload', licenceAccessRoutes.getUserCaseload)
```
smcveigh941, line 38: We are going to end up having a lot of routes, and a lot of route handlers inside one class. I was thinking of splitting those out into classes which just have one GET and one POST method (if applicable).. just a thought

timharrison-moj: In my mind I was thinking that we would group the related routes for one `workflow` into the same class. So maybe `licenceRoutes` is not the right classification - might be better as `createLicenceRoutes` and `varyLicenceRoutes` etc..

timharrison-moj: ^ the comment in the code about is right, but the name isn't :-)

server/views/pages/licence/preview.njk
```diff
@@ -0,0 +1,52 @@
...
+            <p class="govuk-!-font-weight-bold govuk-!-margin-bottom-1">Name</p>
+            <p>{{ licence.roLastName }}, {{ licence.roFirstName }}</p>
+            <p class="govuk-!-font-weight-bold govuk-!-margin-bottom-1">Email</p>
+            <p>{{ licence.roEmail }}</p>
+            <p class="govuk-!-font-weight-bold govuk-!-margin-bottom-1">Email</p>
+            <p>{{ licence.roTelephone }}</p>
+        </div>
+    </div>
+</div>
+
```
timharrison-moj, line 49: I'm going to be tidying this template up quite a bit so it resembles an existing licence document, making Nunjucks macros for the titles, offender data, standard conditions, additional conditions, signatures etc..

smcveigh941 (approved):
LGTM

joelstobart-moj (approved):
LGTM -> (On behalf of Stephen)

PR #16, opened 2021-08-13
CVL-12: Use correct reference to gotenberg service

timharrison-moj:
The Gotenberg container is referenced by its service name within the cluster, so its URL from within the create-and-vary-a-licence service should have been `http://create-and-vary-a-licence-gotenberg:3000` - now corrected.


joelstobart-moj (approved):
LGTM

PR #18, opened 2021-08-18
CVSL-48: Created initial route handlers and view templates for v7 of …

smcveigh941:


server/views/pages/create/additionalConditions.njk
```diff
@@ -0,0 +1,123 @@
...
+                    name: "additional-conditions",
+                    fieldset: {
+                        legend: {
+                            text: "Restriction of residency",
+                            isPageHeading: true,
+                            classes: "govuk-fieldset__legend--l"
+                        },
+                        classes: "govuk-!-margin-top-6"
+                    },
+                    items: [
```
timharrison-moj, line 105: This is good for now (having the list of options in the template). Longer term, I wonder about pulling these options into a JSON config file and supplying the options objects into the templates. We could also version these - i.e. increment a licence version number whenever we change one of the configuration items. Discuss!

server/routes/licenceRoutes.ts
```diff
@@ -68,4 +68,137 @@ export default class LicenceRoutes {
...
+
+  public postInitialMeetingContact: RequestHandler = async (req, res): Promise<void> => {
+    const { id } = req.params
+    res.redirect(`/licence/create/id/${id}/initial-meeting-time`)
+  }
+
+  public getInitialMeetingTime: RequestHandler = async (req, res): Promise<void> => {
+    const offender = {
+      name: 'Adam Balasaravika',
+    }
```
timharrison-moj, line 119: Good for now, and we can extend this to call the licence service to supply the data rather than stub in the routes.

server/views/pages/create/additionalConditions.njk
smcveigh941: I had a think about this, and I only added in a subset of the options into html template for now since we will eventually be pulling them out into some kind of field config. Not sure how we want this to work, since each checkbox can have different kinds of input nested inside of them. I came across this which allows html-as-config, but not sure this is the route you want to go down https://json2html.com/

server/views/pages/create/additionalConditions.njk
```diff
@@ -0,0 +1,123 @@
...
+                        {
+                            value: "restriction-of-residency-condition-id-0",
+                            text: "Not to reside (not even to stay for one night) in the same household as [ANY / ANY FEMALE / ANY MALE] child under the age of [INSERT AGE] without the prior approval of your supervising officer.",
+                            conditional: {
+                                html: restrictionOfResidencyHtml
+                            }
+                        }
+                    ]
+                }) }}
```
timharrison-moj, line 114: Are we still missing a lot of the questions? I think there were a lot more in the prototype, but a good start to get some of them in.

smcveigh941: Included a subset of the conditions as hardcoded html with an example of what the payload could look like. We would eventually be migrating this to html-as-config I think? So didn't hard code the whole thing for now but happy to do so in the short term

timharrison-moj (approved):
Looks good to me. This is the initial workflow for the creation of a licence - its using stubbed data for now, so we've agreed to add some basic tests to this as a follow-on PR and then revisit each page to add the real data, integration & unit tests in their own PRs.

PR #20, opened 2021-08-23
CVSL-48: Configure Nunjucks to return a context environment for use in cheerio tests

smcveigh941:
- Configure Nunjucks to return an environment context for use in unit tests
- Write a unit test using Cheerio to test view templates in isolation


server/routes/testutils/appSetup.ts
```diff
@@ -45,7 +44,7 @@ function appSetup({
 
   app.set('view engine', 'njk')
 
-  nunjucksSetup(app, path)
+  nunjucksSetup(app)
```
timharrison-moj, line 48: I'm trying to think of any consequence of this - like will it alter the path for running in any of the modes (test, integration test, docker)? I think its neater and less complicated, and if this works in all the modes then I'm not sure why the path module was originally there!

server/views/pages/create/caseload.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+    const $ = cheerio.load(compiledTemplate.render(viewContext))
+    expect($('tbody .govuk-table__row').length).toBe(2)
+    expect($('#name-1').text()).toBe('Adam Balasaravika')
+    expect($('#crn-1').text()).toBe('X381306')
+    expect($('#release-date-1').text()).toBe('03 August 2022')
+    expect($('#name-2').text()).toBe('John Smith')
+    expect($('#crn-2').text()).toBe('X123456')
+    expect($('#release-date-2').text()).toBe('01 September 2022')
+  })
```
timharrison-moj, line 42: Yes, I like this way - a detailed check of the rendered view using stubbed information. As you mentioned, probably most useful for conditional view templates.

server/utils/nunjucksSetup.ts
```diff
@@ -49,4 +52,6 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
   njkEnv.addFilter('concatValues', (object: Record<string, unknown>) => {
     return Object.values(object).join(', ')
   })
+
+  return njkEnv
```
timharrison-moj, line 56: Is the nunjucks environment only need to cater for the cheerio tests? Any other use?

timharrison-moj (commented):
LGTM. A couple of questions, but I reckon we should merge this and check its all fine in DEV.

server/routes/testutils/appSetup.ts
smcveigh941: I cant immediately see why it would break in any case.. Integration tests passed in the pipeline. I guess we should just check that it works in Dev

server/utils/nunjucksSetup.ts
smcveigh941: The nunjucks environment is normally attached to the express app on startup. However in the case of cheerio tests, we would like to create and return the nunkucks environment to be able to render nunjucks templates without instantiating an express app

timharrison-moj (approved):
LGTM

PR #27, opened 2021-08-31
CVSL-54: Added validation middleware and flash messages

smcveigh941:
- Added validation middleware
- Add error messages to flash to be retrieved on the following GET
- Map errors to GOV.uk format for display https://design-system.service.gov.uk/components/error-summary/
- Add `REFERER` header to allow `res.redirect('back')` to work. Same origin only.
- Add cheerio tests to test the nunjucks filters in isolation
- Errors are displayed in the order they are validated (i.e. annotated)

screencapture-localhost-3000-licence-create-id-1-initial-meeting-time-2021-09-02-21_11_59

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,8 @@
     "body-parser": "^1.19.0",
     "bunyan": "^1.8.15",
     "bunyan-format": "^0.2.1",
+    "class-transformer": "^0.4.0",
+    "class-validator": "^0.13.1",
     "compression": "^1.7.4",
     "connect-flash": "^0.1.1",
     "connect-redis": "^6.0.0",
@@ -106,6 +108,7 @@
     "http-errors": "^1.8.0",
     "jquery": "^3.6.0",
     "jwt-decode": "^3.1.2",
+    "moment": "^2.29.1",
     "nocache": "^3.0.1",
     "nunjucks": "^3.2.3",
     "passport": "^0.4.1",
@@ -158,6 +161,7 @@
     "path-parse": "^1.0.7",
     "pdf-parse": "^1.1.1",
     "prettier": "^2.3.2",
+    "reflect-metadata": "^0.1.13",
     "ts-jest": "^27.0.3",
     "typescript": "^4.4.2"
   }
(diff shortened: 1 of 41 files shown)


package.json
```diff
@@ -106,6 +108,7 @@
     "http-errors": "^1.8.0",
     "jquery": "^3.6.0",
     "jwt-decode": "^3.1.2",
+    "moment": "^2.29.1",
     "nocache": "^3.0.1",
```
timharrison-moj, line 112: Oh good. I was going to add moment to the dependencies.

server/routes/creatingLicences/types/time.ts
```diff
@@ -0,0 +1,14 @@
...
+
+class SimpleTime {
+  @Expose()
+  hour: string
+
+  @Expose()
+  minute: string
+
+  @Expose()
+  ampm: string
```
timharrison-moj, line 11: Could this need an enumerated type for validation i.e. 'am' or 'pm'?

timharrison-moj: e.g. @IsIn(['am,'pm'])

server/views/layout.njk
```diff
@@ -31,10 +31,12 @@
...
   {{ govukPhaseBanner({
     tag: {
-      text: "alpha"
+      text: "beta"
     },
     html: 'This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it.'
   }) }}
+
+  {% include 'partials/formError.njk' %}
 {% endblock %}
```
timharrison-moj, line 40: I wonder if this might fail accessibility testing being, like the phase banner, outside of HTML landmarks. I'll try it and pick this up in my PR.

timharrison-moj (dismissed):
I like this PR very much! Great work on validation classes and tests, and particularly around the date and time types.

server/views/layout.njk
smcveigh941: Wave has no issue with this, but thats not to say Axe wont :) 

Screenshot 2021-09-03 at 11 05 35

server/routes/creatingLicences/types/time.ts
smcveigh941: Added this Enum, I still kept the validation of the enum together with the validation of the rest of the SimpleTime class though

timharrison-moj (approved):
LGTM

PR #28, opened 2021-09-01
CVSL-55: Accessibility testing with Axe and Cypress

timharrison-moj:
Added accessibility testing with Axe and Cypress.
Small issue with phase banner not being within landmark tags - solved by placing within <nav></nav> for now.


server/views/layout.njk
```diff
@@ -29,12 +29,14 @@
 {% endblock %}
 
 {% block beforeContent %}
+  <nav>
```
smcveigh941, line 32: The phase tag isn't really a navbar, can we include it in the `header.njk` instead?

smcveigh941 (changes requested):
Looks good, just one request

server/views/layout.njk
timharrison-moj: yes, I'll take a look at that first thing.

timharrison-moj: We already have a role of `banner` in the main title bar, and can't have two of these, so made this an aria role of `complementary` with a div id of `phase-banner`.

server/views/layout.njk
```diff
@@ -29,13 +29,14 @@
...
 
 {% block beforeContent %}
-  {{ govukPhaseBanner({
-    tag: {
-      text: "beta"
-    },
-    html: 'This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it.'
-  }) }}
-
+  <div id="phase-banner" role="complementary">
```
smcveigh941, line 32: instead of adding this wrapper, you should be able to add the role attribute to the phase banner macro, for example,

```
    {{ govukPhaseBanner({
      attributes: {
        role: 'complementary'
      },
      tag: {
        text: "beta"
      },
      html: 'This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it.'
    }) }}
```

smcveigh941 (changes requested):

server/views/layout.njk
timharrison-moj: just looking now, thanks.

timharrison-moj: Yes. That works better. Pushing shortly.

smcveigh941 (approved):
LGTM

PR #30, opened 2021-09-06
CVSL-94 Add a caseload view for discussion with the team

timharrison-moj:
* Adds a caseload template
* Adds stubbed data for paginated caseload list view
* Adds status tags and configuration for status values/colours and descriptions.
* Add responsive view which reduces detail as the screen width shrinks for desktop, iPad and mobile widths.
* No tests - just a spike for discussion with the team - it may be throw-away

image


smcveigh941 (approved):
LGTM

PR #31, opened 2021-09-08
CVSL-96: Updated alert channel to public channel and updated dependencies

timharrison-moj:
In this PR
* Use farsight-alerts in place of farsight-devs for alerts (public channel)
* Update sass version
* Update govuk-frontend version - fails the check_outdated circle job overnight


smcveigh941 (approved):
LGTM

PR #32, opened 2021-09-10
CVSL-107: Add additional conditions config and form builders + validation

smcveigh941:
- Licence conditions in JSON
- conditionsProvider util functions to provide the conditions in various structures (e.g. grouped by group name)
- formBuilder which takes the names of partials as params to decide which partials to display
- AdditionalConditions form validation - this will probably change in future if split out into separate pages
screencapture-localhost-3000-licence-create-id-1-additional-conditions-2021-09-10-11_04_45

### DIFF
--- a/server/config/conditions.ts
+++ b/server/config/conditions.ts
@@ -1,0 +1,38 @@
+export default {
+  standardConditions: [
+    {
+      text: 'Be of good behaviour and not behave in a way which undermines the purpose of the licence period.',
+    },
+    {
+      text: 'Not commit any offence.',
+    },
+    {
+      text: 'Keep in touch with the supervising officer in accordance with instructions given by the supervising officer.',
+    },
+    {
+      text: 'Receive visits from the supervising officer in accordance with instructions given by the supervising officer.',
+    },
+    {
+      text: 'Reside permanently at an address approved by the supervising officer and obtain the prior permission of the supervising officer for any stay of one or more nights at a different address.',
+    },
+    {
+      text: 'Not undertake work, or a particular type of work, unless it is approved by the supervising officer and notify the supervising officer in advance of any proposal to undertake work or a particular type of work.',
+    },
+    {
+      text: 'Not travel outside the United Kingdom, the Channel Islands or the Isle of Man except with the prior permission of your supervising officer or for the purposes of immigration deportation or removal.',
+    },
+  ],
+  additionalConditions: [
+    {
+      id: 'placeOfResidence',
+      groupName: 'Residence at a specific place',
+      text: 'You must reside within [INSERT REGION] while of no fixed abode, unless otherwise approved by your supervising officer.',
+      inputTemplate: 'placeOfResidence',
+    },
+    {
+      id: 'restrictionOfResidency',
+      groupName: 'Restriction of residency',
+      text: 'Not to reside (not even to stay for one night) in the same household as [ANY / ANY FEMALE / ANY MALE] child under the age of [INSERT AGE] without the prior approval of your supervising officer.',
+    },
+  ],
+}
--- a/server/config/conditions.ts
+++ b/server/config/conditions.ts
@@ -30,9 +30,9 @@ export default {
       inputTemplate: 'placeOfResidence',
     },
     {
-      id: 'restrictionOfResidency',
-      groupName: 'Restriction of residency',
-      text: 'Not to reside (not even to stay for one night) in the same household as [ANY / ANY FEMALE / ANY MALE] child under the age of [INSERT AGE] without the prior approval of your supervising officer.',
+      id: 'attendAppointments',
+      groupName: 'Making or maintaining contact with a person',
+      text: 'Attend all appointments arranged for you with a psychiatrist / psychologist / medical practitioner.',
     },
   ],
 }
--- a/server/middleware/validationMiddleware.ts
+++ b/server/middleware/validationMiddleware.ts
@@ -18,7 +18,7 @@ function validationMiddleware(type: new () => unknown): RequestHandler {
       return next()
     }
 
-    const addError = (
+    const buildError = (
       error: ValidationError,
       constraints: {
         [type: string]: string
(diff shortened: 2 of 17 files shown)


server/config/conditions.ts
```diff
@@ -0,0 +1,38 @@
+export default {
+  standardConditions: [
+    {
+      text: 'Be of good behaviour and not behave in a way which undermines the purpose of the licence period.',
+    },
```
timharrison-moj, line 5: Do standard conditions need a sequence? Or rely on the order they appear in the list?

server/config/conditions.ts
```diff
@@ -0,0 +1,38 @@
...
+      id: 'placeOfResidence',
+      groupName: 'Residence at a specific place',
+      text: 'You must reside within [INSERT REGION] while of no fixed abode, unless otherwise approved by your supervising officer.',
+      inputTemplate: 'placeOfResidence',
+    },
+    {
+      id: 'restrictionOfResidency',
+      groupName: 'Restriction of residency',
+      text: 'Not to reside (not even to stay for one night) in the same household as [ANY / ANY FEMALE / ANY MALE] child under the age of [INSERT AGE] without the prior approval of your supervising officer.',
+    },
```
timharrison-moj, line 36: Is this `restrictionOfResidency` incomplete? No input template or definition of data required?

timharrison-moj: Do we need a sequence number, within groupName, so can order these on the page when printing? Maybe later?

server/config/conditions.ts
smcveigh941: At the moment, conditions will appear in the order they are in config. Actually at the moment, standard conditons are not used anywhere just yet

server/config/conditions.ts
smcveigh941: At the moment, the conditions will appear on screen in the order they are defined in config.. If we need them in a specific order, we can just put them in config in the correct order? or have some kind of sequence number as you suggest.

On the incomplete condition, I did this as an example of a condition which does not have any nested inputs. Let me change it to a different text though to reflect that

server/config/conditions.ts
timharrison-moj: API will accept them on create call :-)

server/routes/creatingLicences/types/additionalConditions.ts
```diff
@@ -0,0 +1,16 @@
+import { Expose, Type } from 'class-transformer'
+import { ValidateIf, ValidateNested } from 'class-validator'
+import RegionOfResidence from './regionOfResidence'
+
+class AdditionalConditions {
+  @Expose()
+  additionalConditions: string[]
+
+  @Expose()
```
timharrison-moj, line 9: So this will grow with validation objects of different types, one per condition, and each with their own rules? Nice.

server/utils/conditionsProvider.ts
```diff
@@ -0,0 +1,14 @@
...
+export default function getGroupedAdditionalConditions(): Record<string, unknown>[] {
+  const map = new Map()
+  conditionsConfig.additionalConditions.forEach(condition => {
+    const collection = map.get(condition.groupName)
+    if (!collection) {
+      map.set(condition.groupName, [condition])
+    } else {
+      collection.push(condition)
+    }
+  })
```
timharrison-moj, line 12: Organising conditions into groups for display together?

server/views/formBuilder.njk
```diff
@@ -0,0 +1,5 @@
+{% macro formBuilder(params) %}
+    {% if params.inputTemplate == 'placeOfResidence' %}
+        {% include 'partials/additionalConditionsInputs/placeOfResidence.njk' %}
+    {% endif %}
+{% endmacro %}
```
timharrison-moj, line 5: Would we add an include for each group or condition? Not sure if the pages are being split by group or individual condition yet.

server/views/partials/additionalConditionsInputs/placeOfResidence.njk
```diff
@@ -0,0 +1,12 @@
...
+{{ govukInput({
+    id: 'regionOfResidence',
+    name: params.id + '[regionOfResidence]',
+    classes: "govuk-!-width-one-third",
+    label: {
+        text: "Enter probation region"
+    },
+    errorMessage: params.errorMessages | findError('regionOfResidence'),
+    value: params.formResponses[params.id].regionOfResidence
+}) }}
```
timharrison-moj, line 12: Its not that obvious what is contained within `params` here? I wonder if to keep things clear we could add a comment at the top of each additional condition template to describe what is expected to be provided to it? Might make this easier to maintain/ change in future?
e.g.
`{%
   This template expects a params object containing:
      id - the condition identifier
      y - xxxx
%}`

server/middleware/validationMiddleware.ts
```diff
@@ -22,19 +22,15 @@ function validationMiddleware(type: new () => unknown, useChildPath = false): Re
...
 
-    const mapErrors = (error: ValidationError) =>
-      error.children.length > 0
-        ? error.children.map((childError: ValidationError) =>
-            addError(error, childError.constraints, `${error.property}[${childError.property}]`)
-          )
-        : addError(error, error.constraints)
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const mapErrors: any = (error: ValidationError) =>
+      error.children.length > 0 ? error.children.flatMap(mapErrors) : addError(error, error.constraints)
```
timharrison-moj, line 34: Looks like this collects together all of the validation errors for dependent children into a single list? Maybe a comment could be useful to describe / make it clear?

server/views/pages/create/additionalConditions.test.ts
```diff
@@ -0,0 +1,145 @@
...
+              id: 'condition2',
+              text: 'conditionText2',
+            },
+          ],
+        },
+      ],
+    }
+
+    const $ = cheerio.load(compiledTemplate.render(viewContext))
+
```
timharrison-moj, line 141: Like it.

server/views/pages/create/additionalConditions.njk
```diff
@@ -20,98 +20,42 @@
...
+                {% for group in additionalConditions %}
+                    {% set items = [] %}
+                    {% for condition in group.conditions %}
+                        {% set conditionalHtml = {} %}
+                        {% if condition.inputTemplate %}
+                            {% set conditionalHtml = {
+                                html: formBuilder({
+                                    id: condition.id,
+                                    inputTemplate: condition.inputTemplate,
+                                    errorMessages: validationErrors,
```
timharrison-moj, line 32: This is quite neat. I can see why form-builder looks a reasonable choice here. Lets try run with this, and we could change if we see problems down the line.

timharrison-moj (dismissed):
I like it and looks reasonable concise and easy to extend. Couple of minor comments but mostly about adding some extra description to make it easier to understand and follow.

server/routes/creatingLicences/types/additionalConditions.ts
smcveigh941: Yes it will, but I anticipate that there might be some change around here regarding how these classes look depending on how the new design looks

server/utils/conditionsProvider.ts
smcveigh941: Yea, the template iterates through the groups and displays each condition within that group. This function collects the conditions into groups by looking at the `groupName` tags of each condition

server/views/formBuilder.njk
smcveigh941: The way the prototype currently looks, we would have an `include` here for each nested set of inputs. However again I expect this might change as we iterate

timharrison-moj (approved):
LGTM

timharrison-moj (approved):
LGTM

PR #34, opened 2021-09-17
CVSL-95: Take User roles into account to display the correct journey cards on the landing page

smcveigh941:
- Add user roles into the express.User object
- Util function to check for a specific role for the user
- Display cards on landing page depending on whether certain roles exist
- Removed the roles  'ROLE_CVL_OMU' and 'ROLE_GLOBAL_SEARCH' - replaced by relevant licence roles
- Started new cypress test for create a licence journey to be built on later


server/routes/home/handlers/home.test.ts
```diff
@@ -18,7 +21,11 @@ describe('Route Handlers - Home', () => {
   describe('GET', () => {
     it('', async () => {
       await handler.GET(req, res)
-      expect(res.render).toHaveBeenCalledWith('pages/index')
+      expect(res.render).toHaveBeenCalledWith('pages/index', {
+        shouldShowCreateLicenceCard: true,
+        shouldShowApproveLicenceCard: true,
+        shouldShowMyCaseloadCard: true,
+      })
```
timharrison-moj, line 28: This is nice.

integration_tests/integration/createLicence.spec.js
```diff
@@ -0,0 +1,15 @@
+const IndexPage = require('../pages/index')
+
+context('Create a Licence', () => {
```
timharrison-moj, line 3: Should there be an Axe test here to verify that the page is accessible? Or better to put these accessibility checks in a different test?

timharrison-moj (approved):
LGTM. One question about an Axe test, but could be a follow-on PR

integration_tests/integration/createLicence.spec.js
smcveigh941: Added Axe testing to the verifyPage function - this allows accessibility to be tested implicitly each time a page is loaded during the test

timharrison-moj (approved):
Sorted.

PR #36, opened 2021-09-20
CVSL-125: Placeholder for creating a licence and persisting to the database, returns an ID to be used as route param throughout journey.

smcveigh941:
- Generated types from licence API swagger docs.
- Integrated with the create licence API endpoint, pulling standard conditions from config.
- Licence uses hardcoded data, will need to pull real data from nomis and delius later.
- Licence API returns an id to be used as a route parameter throughout
- Added a create licence button to the caseload view to initiate a POST request for creating the licence. This page is to be revisited 
screencapture-localhost-3000-licence-create-caseload-2021-09-20-15_12_35


server/@types/licenceApiImport/index.d.ts
```diff
@@ -2,26 +2,278 @@
...
+        | 'IN_PROGRESS'
+        | 'SUBMITTED'
+        | 'ACTIVE'
+        | 'REJECTED'
+        | 'SUPERSEDED'
+        | 'IN_PROGRESS'
+        | 'SUBMITTED'
+        | 'REJECTED'
+        | 'ACTIVE'
+        | 'SUPERSEDED'
```
timharrison-moj, line 167: Are these in twice?

server/data/licenceApiClient.ts
```diff
@@ -1,6 +1,7 @@
 import RestClient from './restClient'
-import type { LicenceApiTestData } from './licenceApiClientTypes'
+import type { CreateLicenceResponse, LicenceApiTestData } from './licenceApiClientTypes'
 import config, { ApiConfig } from '../config'
+import { CreateLicenceRequest } from './licenceApiClientTypes'
```
timharrison-moj, line 5: Can this be combined into the other line above as a 3rd type to import?

server/services/licenceService.ts
```diff
@@ -11,6 +12,44 @@ export default class LicenceService {
...
   }
 
+  async createLicence(username: string): Promise<CreateLicenceResponse> {
+    const token = await this.hmppsAuthClient.getSystemClientToken(username)
+    // TODO construct with real licence data using prison and community APIs
+    const licence = {
+      typeCode: 'AP',
+      version: '1.0',
+      nomsId: Math.floor(Math.random() * 900000 + 100).toString(), // Generate random nomsId for now ... should be unique ID from nomis later
+      bookingNo: '12334',
```
timharrison-moj, line 22: We could use a real nomsId that we know is in DEV? e.g. This one is in both Nomis and Delius in the DEV env - `A5198DY`

timharrison-moj (dismissed):
LGTM. Minor questions / comments only.

server/@types/licenceApiImport/index.d.ts
smcveigh941: Yep, the script generated them twice for some reason... fixed now

server/data/licenceApiClient.ts
smcveigh941: done

server/services/licenceService.ts
smcveigh941: Reason for this randomness is the API rejects duplicate nomsIds, and since this is stubbed data we would have to delete it each time from the database when clicking through the flow. Hence this is randomised for now while we have the stubbed data

timharrison-moj (approved):
LGTM

PR #39, opened 2021-09-22
CVSL-100: Initial appointment person page + fetch licence middleware

smcveigh941:
* Added middleware to fetch the licence on each get request provided a licenceId is in the route
* Added the GET and PUT requests to the api client to update the licence with the content of the form
* Mapped form responses onto API requests
* Cypress tests to come in a following PR


server/middleware/fetchLicenceMiddleware.ts
```diff
@@ -0,0 +1,16 @@
...
+      const { username } = res.locals.user
+      const licence = await licenceService.getLicence(req.params.licenceId, username)
+
+      if (licence) {
+        res.locals.licence = licence
+      }
+    }
+    next()
+  }
+}
```
timharrison-moj, line 16: I like this.

server/routes/creatingLicences/handlers/checkAnswers.ts
```diff
@@ -5,7 +5,7 @@ export default class CheckAnswersRoutes {
   constructor(private readonly licenceService: LicenceService) {}
 
   GET = async (req: Request, res: Response): Promise<void> => {
-    const mockLicence = this.licenceService.getLicence()
+    const mockLicence = this.licenceService.getLicenceStub()
     res.render('pages/create/checkAnswers', { licence: mockLicence })
```
timharrison-moj, line 9: Just checking - `getLicenceStub` temporary until all forms have been visited for the real flow?

smcveigh941: yes!

server/routes/creatingLicences/handlers/initialMeetingName.ts
```diff
@@ -1,15 +1,18 @@
...
 
 export default class InitialMeetingNameRoutes {
+  constructor(private readonly licenceService: LicenceService) {}
+
   GET = async (req: Request, res: Response): Promise<void> => {
-    const offender = {
-      name: 'Adam Balasaravika',
-    }
-    res.render('pages/create/initialMeetingPerson', { offender })
+    res.render('pages/create/initialMeetingPerson')
```
timharrison-moj, line 8: This is the part where it may not be obvious that middleware is populating a licence based on pattern-matched URL and value.  Worth a comment in the places where we rely on the middleware doing its stuff?

timharrison-moj (dismissed):
LGTM. Should ease the flow for the rest of the tickets too.

server/routes/creatingLicences/handlers/initialMeetingName.ts
smcveigh941: Since we will have this pattern almost everywhere, I have added an explanatory comment into the `index.ts` file which initialises the routes.

timharrison-moj (approved):
LGTM

PR #40, opened 2021-09-22
CVSL-104 & CVSL-108: Completed yes/no question pages, pulling licence data from session

smcveigh941:
* Added enum for YES/NO
* Removed hardcoded offender data, replaced with licence data from session


timharrison-moj (approved):
Nice and neat. LGTM.

PR #42, opened 2021-09-25
CVSL-103 Date and time of initial meeting

timharrison-moj:
Includes:

* Updated API types - combined date and time into one field for initial meeting time to match API changes
* Date and time conversion utilities and tests to verify (SimpleDateTime <-> JSON date/time string)
* Unit tests for route, service and API calls for meeting time & contact person
* Prepopulating the form with current date/time + person to meet if either are present on the licence (from the API)
* Using details from the fetched current licence (surname, forename) for template display values


server/routes/creatingLicences/handlers/initialMeetingTime.ts
```diff
@@ -1,15 +1,20 @@
...
+
   GET = async (req: Request, res: Response): Promise<void> => {
-    const offender = {
-      name: 'Adam Balasaravika',
-    }
-    res.render('pages/create/initialMeetingTime', { offender })
+    const { appointmentTime } = res.locals.licence
+    const formDate = jsonToSimpleDateTime(appointmentTime)
+    res.render('pages/create/initialMeetingTime', { formDate })
   }
```
timharrison-moj, line 12: I wondered about this bit. An alternative way would be to use the `fetchLicenceMiddleware` to also do the appointment date and time conversion to a SimpleDateTime and add it into the licence object to simplify the route code and make all date/time conversions done in the service layer. Discuss.

server/@types/licenceApiImport/index.d.ts
```diff
@@ -4,6 +4,10 @@
  */
 
 export interface paths {
+  '/licence/id/{licenceId}/appointmentTime': {
+    /** Update the date and time for the initial appointmen. Requires ROLE_SYSTEM_USER or ROLE_CVL_ADMIN. */
```
smcveigh941, line 8: I'll fix this typo in the API

server/data/licenceApiClient.test.ts
```diff
@@ -47,10 +54,35 @@ describe('Licence API client tests', () => {
...
+
+  describe('Licence updates', () => {
+    const person = { contactName: 'Fred' } as PersonName
+    const appointmentPersonRequest = { appointmentPerson: person.contactName } as AppointmentPersonRequest
+    const appointmentTimeRequest = { appointmentTime: '12/12/2022 11:15' } as AppointmentTimeRequest
+    const simpleDate = new SimpleDate('12', '12', '2022')
+    const simpleTime = new SimpleTime('11', '15', AmPm.AM)
+    const simpleDateTime = SimpleDateTime.fromSimpleDateAndTime(simpleDate, simpleTime)
+
+    it('Update appointment person to meet', async () => {
```
smcveigh941, line 72: Thanks 🙏🏼

server/routes/creatingLicences/handlers/initialMeetingTime.ts
smcveigh941: I think what you have done is fine in general. I'm not a fan of converting it in the middleware, because it might be useful elsewhere to have it in a format that can be imported to moment for date calculations. The only use case for the SimpleDateTime format is to split it out into the 6 input fields we have on our forms

server/utils/utils.ts
```diff
@@ -18,4 +22,44 @@ const convertToTitleCase = (sentence: string): string =>
...
-export { convertToTitleCase, hasRole }
+/**
+ * Converts a SimpleDateTime display value to a JSON string format dd/mm/yyyy hh:mm
+ * @param dt: SimpleDateTime
+ */
+const simpleDateTimeToJson = (dt: SimpleDateTime): string | undefined => {
+  const { inductionDate, inductionTime } = dt
+  const momentDt = moment(
+    [inductionDate.year, inductionDate.month, inductionDate.day].join('-'),
+    ['YYYY-M-D', 'YY-M-D'],
```
smcveigh941, line 33: Why 2 formats?

server/utils/utils.ts
```diff
@@ -18,4 +22,44 @@ const convertToTitleCase = (sentence: string): string =>
...
+    true
+  )
+  if (!momentDt.isValid()) {
+    return undefined
+  }
+  let hour
+  if (inductionTime.ampm === 'pm') {
```
smcveigh941, line 40: Moment can probably do this for us? i.e. import into moment in the format `YYYY MM DD hh mm a`, and reformat to `YYYY-MM-DD HH:mm`? Something like this?:

```
const dateString = [inductionDate.year, inductionDate.month, inductionDate.day, inductionTime.hour, inductionTime.minute, inductionTime.ampm].join(' ')
return moment(dateString, 'YYYY MM DD hh mm a').format('YYYY-MM-DD HH:mm')
```

server/utils/utils.ts
```diff
@@ -18,4 +22,44 @@ const convertToTitleCase = (sentence: string): string =>
...
+  momentDt.add(+inductionTime.minute, 'minutes')
+  return momentDt.format('DD/MM/YYYY HH:mm')
+}
+
+/**
+ * Converts a JSON date time `dd/mm/yyyy hh:mm` to a SimpleDateTime for display
+ * @param dt: string
+ */
+const jsonToSimpleDateTime = (dt: string): SimpleDateTime | undefined => {
+  const momentDt = moment(dt, 'D/MM/YYYY HHmm')
```
smcveigh941, line 55: Is this the correct format as expected from the API? No colon?

smcveigh941 (commented):
Looks good in general, some questions about the util

server/@types/licenceApiImport/index.d.ts
timharrison-moj: Already done it in my next PR for contact number! Coming shortly.

server/utils/utils.ts
timharrison-moj: I copied this from the existing code for validating - assumed 2-digit years are also possible on input?

server/utils/utils.ts
timharrison-moj: I think the format ignores white-space and non-numeric characters, and the tests (a lot of edge cases tried) are all happy with this, so should be ok?

server/utils/utils.ts
timharrison-moj: I was thinking this is harder than it should be. Let me give this method a go on my branch to see if I can get it simpler.

timharrison-moj: Yep. That is better, and more simple. Thanks

server/utils/utils.ts
smcveigh941: Oh yes you are right, sorry. Might need added in again? 😢

smcveigh941 (approved):
LGTM, will merge after a green run of the pipeline

server/utils/utils.ts
timharrison-moj: Unit test added for a 2-digit year

PR #45, opened 2021-09-28
CVSL-101: Where will the initial appointment be

smcveigh941:
* Removed postcode regex
* Added new appointment address endpoint to api client and wiremock
* Set `req.body` equal to the class provided in validation. This is because of an interesting find where the csrf token was getting set as `addressLine1`, even though the type passed as a parameter did not include a csrf field
* Fixed bug where the form would get filled with the data from the licence instead of the formResponses when a validation error happens due to empty fields


server/utils/utils.ts
```diff
@@ -55,4 +56,37 @@ const jsonToSimpleDateTime = (dt: string): SimpleDateTime | undefined => {
...
+
+/**
+ * Converts a comma-separated string to an address object for display
+ * @param address
+ */
+const stringToAddressObject = (address: string): Address => {
+  if (!address) {
+    return undefined
+  }
+  const addressParts = address.split(', ')
```
mikehalmamoj, line 75: I guess string addresses are generated by your code so you can assume all the commas are there?

mikehalmamoj (approved):

timharrison-moj (approved):
LGTM too

PR #47, opened 2021-10-01
CVSL-132: Combine nomis and delius to build a caseload for the currently logged in user

smcveigh941:
* Shifted all the type declarations to the `@types` folder
* Implemented the prisoner search endpoint to search the nomisIds returned in caseload from delius.
* Combined data from nomis and delius to create a single type `ManagedCase` with all of the information about a case
* We should think about caching this caseload for use later


server/data/prisonerSearchApiClient.ts
```diff
@@ -10,9 +11,16 @@ export default class PrisonerSearchApiClient {
   }
 
   async searchPrisoners(prisonerSearchCriteria: PrisonerSearchCriteria): Promise<Prisoner[]> {
-    return this.restClient.post({
+    return (await this.restClient.post({
```
yasinmustafa, line 14: I dont think you need these awaits at this level - if you await at the calling function that should suffice

mikehalmamoj (dismissed):

yasinmustafa (dismissed):
LGTM - one minor comment

integration_tests/mockApis/prison.ts
```diff
@@ -0,0 +1,62 @@
+import { SuperAgentRequest } from 'superagent'
+import { stubFor } from '../wiremock'
+
```
timharrison-moj, line 3: As there are two APIs we'll use in prisons - maybe a name change to `prisonerSearchApi` for this one? Other one can be something like `prisonApi`? Just clarifying which is which?

server/routes/creatingLicences/handlers/caseload.test.ts
```diff
@@ -47,9 +44,9 @@ describe('Route Handlers - Create Licence - Caseload', () => {
       expect(res.render).toHaveBeenCalledWith('pages/create/caseload', {
         caseload: [
           {
-            name: 'Jones',
+            name: 'Joe Rogan',
             crnNumber: 'X381306',
-            conditionalReleaseDate: '03 August 2022',
+            conditionalReleaseDate: '09/09/2022',
           },
```
timharrison-moj, line 50: Might be some content challenge on date formats - but not sure we're had a steer on consistent date format yet!

server/services/caseloadService.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    return (await this.prisonerService.searchPrisonersByNomisIds(username, caseloadNomisIds)).map(offender => {
+      const matchingDeliusCase = managedOffenders.find(deliusCase => deliusCase.nomsNumber === offender.prisonerNumber)
+      return {
+        ...matchingDeliusCase,
+        ...offender,
+      } as ManagedCase
+
+      // TODO Cache this by username?
+    })
```
timharrison-moj, line 34: Can cache in redis later and wrap the get with a redis lookup first - good examples in other services to reuse.

timharrison-moj (approved):
LGTM - tiny comment about naming only.

timharrison-moj (approved):
LGTM.

PR #49, opened 2021-10-04
CVSL-133: Added prison register client, tests and integration mock API

timharrison-moj:
Adds prison register API to get prison description from code, and OMU email address from agencyId.
Unit and integration tests to verify.


server/data/prisonRegisterApiClient.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      fakeApi.get(`/prisons/id/${prisonCode}`).reply(200, prisonDto)
+      const data = await prisonRegisterService.getPrisonDescription(username, prisonCode)
+      expect(data).toEqual(prisonDto)
+      expect(nock.isDone()).toBe(true)
+      expect(hmppsAuthClient.getSystemClientToken).toBeCalled()
+    })
+
+    it('No client token', async () => {
+      hmppsAuthClient.getSystemClientToken.mockResolvedValue('')
+      fakeApi.get(`/prisons/id/${prisonCode}`).reply(401)
```
mikehalmamoj, line 45: I thought it was only the `/secure` endpoints that required a token?

server/data/prisonRegisterApiClient.ts
```diff
@@ -0,0 +1,23 @@
...
+    return this.restClient.get({
+      path: `/prisons/id/${agencyId}`,
+    }) as Promise<PrisonDto>
+  }
+
+  async getPrisonOmuContactEmail(agencyId: string): Promise<string> {
+    return this.restClient.get({
+      path: `/secure/prisons/id/${agencyId}/offender-management-unit/email-address`,
+    }) as Promise<string>
+  }
```
mikehalmamoj, line 22: Also need some tests around this endpoint?

mikehalmamoj (dismissed):

server/data/prisonRegisterApiClient.test.ts
timharrison-moj: Yes, you're right. First time using prison register for me. I'll sort that now.

server/data/prisonRegisterApiClient.ts
timharrison-moj: This one not currently used, it was more 'get the basics in now' whilst I had all the info to hand - but yes, I'll add some as I'm pretty sure we will use it later.

server/data/prisonRegisterApiClient.test.ts
timharrison-moj: Looking at this, and at other service implementations, some have amended the rest client to make the token nullable, and then conditionally made the call to superagent either with of without the authorization header if no token is provided. I'm just thinking that leaving this with the token always in place is probably less harmful and simpler? As an insecure endpoint it just won't check the auth header even when provided?

server/data/prisonRegisterApiClient.ts
timharrison-moj: I've put some tests around this now.

server/data/prisonRegisterApiClient.test.ts
mikehalmamoj: Hmmm I think you're right, though it feels a bit dirty.

mikehalmamoj (approved):

PR #52, opened 2021-10-06
CVSL-98: Deployment of a container to stub the Community API caseload.

timharrison-moj:
Part of a fiendish plan to avoid the data mismatch between Delius and Nomis.
This deploys a service in a container in the namespace which pretends to be the Community API for /staff and /managedOffenders, and returns a list of people who are known in Nomis via the prisoner offender search API. 
We know the Community API works well, but we cant setup the appropriate data there without breaking the Delius TEST environment for the Delius work, so this is the next best thing.

### DIFF
--- a/helm_deploy/create-and-vary-a-licence/values.yaml
+++ b/helm_deploy/create-and-vary-a-licence/values.yaml
@@ -88,5 +88,45 @@ gotenberg:
     runAsUser: 1001
     runAsNonRoot: true
 
+delius-wiremock:
+  nameOverride: delius-wiremock
+  replicaCount: 2
+
+  image:
+    repository: smcveigh941/delius-wiremock
+    tag: latest
+    port: 5000
+
+  namespace_secrets:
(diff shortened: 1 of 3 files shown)


helm_deploy/create-and-vary-a-licence/values.yaml
```diff
@@ -88,5 +88,45 @@ gotenberg:
...
+
+  image:
+    repository: smcveigh941/delius-wiremock
+    tag: latest
+    port: 5000
+
+  namespace_secrets:
+    create-and-vary-a-licence:
+      SYSTEM_CLIENT_ID: "SYSTEM_CLIENT_ID"
+      SYSTEM_CLIENT_SECRET: "SYSTEM_CLIENT_SECRET"
```
smcveigh941, line 103: I assume these string literals point to where the secret it stored, and shouldnt be something like `${SYSTEM_CLIENT_ID}`

PR #56, opened 2021-10-07
CVSL-136: Filter offenders from caseload screen if they already have a licence

smcveigh941:
* Filter offenders from the caseload if they already have a licence
* Currently checking this by staffId. Maybe we should be checking by nomisId?
* Also filtering the offender out if they are INACTIVE, if they have a release date which in the past, or if they are HDC eligible

### DIFF
--- a/integration_tests/integration/createLicence.spec.ts
+++ b/integration_tests/integration/createLicence.spec.ts
@@ -18,6 +18,7 @@ context('Create a licence', () => {
     cy.task('stubGetManagedOffenders', 2000)
     cy.task('searchPrisonersByNomisIds')
     cy.task('stubGetPrisonDescription', 'LEI')
+    cy.task('stubGetLicencesByStaffIdAndStatus', 2000)
     cy.signIn()
   })
 
--- a/integration_tests/mockApis/licence.ts
+++ b/integration_tests/mockApis/licence.ts
@@ -142,4 +142,23 @@ export default {
       },
     })
   },
+
+  stubGetLicencesByStaffIdAndStatus: (staffId: number): SuperAgentRequest => {
+    return stubFor({
+      request: {
+        method: 'GET',
+        urlPathPattern: `/licence/staffId/${staffId}`,
+        queryParameters: {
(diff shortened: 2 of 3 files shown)


server/services/caseloadService.ts
```diff
@@ -15,6 +22,16 @@ export default class CaseloadService {
       .filter(offender => offender.nomsNumber)
       .map(offender => offender.nomsNumber)
 
+    /*
+     TODO: Maybe this should be checking for existing licences by nomisId rather than staffId? What if the offender
+      changes their managing officer after a licence has been created for them?
+     */
+    const existingLicences = await this.licenceService.getLicencesByStaffIdAndStatus(staffIdentifier, username, [
```
timharrison-moj, line 29: I agree it should use nomisId - or a list of nomisId - to get all licences in officer's caseload? 
We need to think about what happens if the staff details on the licence is now different to the staff user who has this offender in their caseload. Its a suggestion that the offender has been reassigned to a different probation officer - so should we also update the details on the licence to match the current user?  OK for now though - this needs to be discussed as a different ticket.

server/services/caseloadService.ts
```diff
@@ -32,5 +49,12 @@ export default class CaseloadService {
       .filter(offender => !offender.indeterminateSentence)
       .filter(offender => !offender.paroleEligibilityDate)
       .filter(offender => offender.legalStatus !== 'DEAD')
+      .filter(offender => offender.status && offender.status.startsWith('ACTIVE'))
+      .filter(offender => !offender.releaseDate || moment().isBefore(moment(offender.releaseDate, 'yyyy-MM-DD')))
+      .filter(offender => !offender.homeDetentionCurfewEndDate)
+      .filter(offender => {
+        const existingLicence = existingLicences.find(licence => licence.nomisId === offender.nomsNumber)
+        return !existingLicence
+      })
```
timharrison-moj, line 58: Is this filter right? Any chance of two licences for the same nomisId, in different statuses? Probably right.

PR #65, opened 2021-10-13
CVSL-112: Check your answers, change links, and final validation before submission

smcveigh941:
- Since the getLicence stub needs to return a completed licence at the review page, and an incomplete licence elsewhere, it prompted me to stub each endpoint as and when they are needed in the page objects rather than at the start
- Change links add `fromReview=true` query param to the urls which drives the return to review page action
- Used `class-validator` to validate the whole licence before submission
- Change links and submit button are hidden if licence status is not `IN_PROGRESS`

### DIFF
--- a/integration_tests/pages/bespokeConditions.ts
+++ b/integration_tests/pages/bespokeConditions.ts
@@ -12,7 +12,7 @@ export default class BespokeConditionsPage extends Page {
 
   enterBespokeCondition = (conditionId: number, condition: string): BespokeConditionsPage => {
     const fieldName = `conditions[${conditionId}]`
-    cy.get(`textarea[name="${fieldName}"]`).clear().type(condition)
+    cy.get(`textarea[name="${fieldName}"]`).type(condition)
     return this
   }
 


server/routes/creatingLicences/handlers/initialMeetingPlace.ts
```diff
@@ -15,6 +15,11 @@ export default class InitialMeetingPlaceRoutes {
     const { licenceId } = req.params
     const { username } = res.locals.user
     await this.licenceService.updateAppointmentAddress(licenceId, req.body, username)
-    res.redirect(`/licence/create/id/${licenceId}/initial-meeting-contact`)
+
+    if (req.query?.fromReview) {
+      res.redirect(`/licence/create/id/${licenceId}/check-your-answers`)
+    } else {
+      res.redirect(`/licence/create/id/${licenceId}/initial-meeting-contact`)
+    }
```
timharrison-moj, line 23: As its a repeating logic, I wonder if there is a middleware solution to redirecting to the `check-your-answers` page, but happy to get this in and then look at improving if or when we can?

server/utils/utils.test.ts
```diff
@@ -171,13 +171,13 @@ test.each`
...
-  ${'22/10/2024 14:23'} | ${'02 23 pm'}
-  ${'22/10/24 14:23'}   | ${'02 23 pm'}
+  ${'12/12/2021 23:15'} | ${'11:15 pm'}
+  ${'31/01/2022 12:01'} | ${'12:01 pm'}
+  ${'31/12/2022 12:00'} | ${'12:00 pm'}
+  ${'31/12/2022 00:00'} | ${'12:00 am'}
+  ${'01/01/2022 01:01'} | ${'01:01 am'}
+  ${'22/10/2024 14:23'} | ${'02:23 pm'}
+  ${'22/10/24 14:23'}   | ${'02:23 pm'}
   ${'null'}             | ${'null'}
```
timharrison-moj, line 181: Thanks! I was going to add that colon!

integration_tests/mockApis/prison.ts
```diff
@@ -2,11 +2,11 @@ import { SuperAgentRequest } from 'superagent'
...
 
 export default {
-  stubGetPrisonerDetail: (nomisId: string): SuperAgentRequest => {
+  stubGetPrisonerDetail: (): SuperAgentRequest => {
     return stubFor({
       request: {
         method: 'GET',
-        urlPattern: `/api/offenders/${nomisId}`,
+        urlPattern: `/api/offenders/G9786GC`,
       },
```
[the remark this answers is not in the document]
smcveigh941: If we did have a different prisoner, would it return different data? If it did, then it would require a different stub anyway? Not sure yet..

PR #77, opened 2021-10-21
CVSL-185: Update standard conditions with unique identifiers and generate sequence numbers

smcveigh941:


server/services/licenceService.ts
```diff
@@ -1,20 +1,20 @@
 import type HmppsAuthClient from '../data/hmppsAuthClient'
 import {
+  AdditionalConditionsRequest,
```
smcveigh941, line 3: This was just a format of the file which re-ordered imports to alphabetical order... no real changes

PR #78, opened 2021-10-21
Cvsl 172 populate user details in session

timharrison-moj:
This PR:
* Requires that the latest delius-wiremock server is deployed alongside it.
* Uses the `populateCurrentUser` middeware function to get and store user information in their session
* Merges this user information from Redis into `res.locals.user` for each request (instead of looking it up again)
* For prison users it stores - name, caseload, displayName, activeCaseload (+ email address from hmpps-auth)
* For probation users it stores - name, staffId, staffCode, displayName, telephoneNumber and emailAddress
* For other users, it retrieves and stores just name, displayName and emailAddress from hmpps-auth
* Alters the two current usages of this data in approval cases and probation caseload.
* Updates all unit and integration tests in line with these changes.

This is quite a big PR and fiddly so didn't want to make it bigger ... hence a couple of follow-ons:

* There were no unit tests around populateCurrentUser before and none now - on my list to put some in
* The user service has no unit tests around `getProbationUser`  - the structure of the CommunityApiClient needs to alter to make it more suitable for mocking in unit tests - on my list.

### DIFF
--- a/server/services/index.ts
+++ b/server/services/index.ts
@@ -8,7 +8,7 @@ import CaseloadService from './caseloadService'
 import PrisonApiClient from '../data/prisonApiClient'
 
 const hmppsAuthClient = new HmppsAuthClient(new TokenStore())
-const prisonApiClient = new PrisonApiClient('no admin token required for user service')
+const prisonApiClient = new PrisonApiClient('admin token not used in user service')
 const userService = new UserService(hmppsAuthClient, prisonApiClient)
 const prisonerService = new PrisonerService(hmppsAuthClient)
 const communityService = new CommunityService(hmppsAuthClient)
--- a/server/utils/utils.ts
+++ b/server/utils/utils.ts
@@ -101,8 +101,7 @@ const jsonDtTo12HourTime = (dt: string): string => {
 }
 
 const removeDuplicates = (list: string[]): string[] => {
-  const deDuplicatedList = [...new Set(list.map(s => s))].map(s => s)
-  return deDuplicatedList
+  return [...new Set(list)]
 }
 
 export {
--- a/server/services/userService.test.ts
+++ b/server/services/userService.test.ts
@@ -30,7 +30,7 @@ describe('User service', () => {
 
     it('Propagates any errors', async () => {
       hmppsAuthClient.getUser.mockRejectedValue(new Error('some error'))
-      await expect(userService.getAuthUser(token)).rejects.toEqual(new Error('some error'))
+      await expect(() => userService.getAuthUser(token)).rejects.toThrow('some error')
       expect(hmppsAuthClient.getUser).toBeCalled()
     })
   })
@@ -49,7 +49,7 @@ describe('User service', () => {
 
     it('Propagates any errors', async () => {
       hmppsAuthClient.getUserEmail.mockRejectedValue(new Error('some error'))
-      await expect(userService.getAuthUserEmail(token)).rejects.toEqual(new Error('some error'))
+      await expect(() => userService.getAuthUserEmail(token)).rejects.toThrow('some error')
       expect(hmppsAuthClient.getUserEmail).toBeCalled()
     })
   })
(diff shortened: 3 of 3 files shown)


integration_tests/pages/index.ts
```diff
@@ -12,7 +12,6 @@ export default class IndexPage extends Page {
   }
 
   clickCreateALicence = (): CaseloadPage => {
-    cy.task('stubGetStaffDetails')
     cy.task('stubGetManagedOffenders')
```
timharrison-moj, line 15: This is no longer retrieved in the caseload page - already present in the session and passed into the service now.

server/@types/CvlUserDetails.ts
```diff
@@ -0,0 +1,39 @@
...
+  // probation area code
+  probationArea?: string
+
+  // Probation area description
+  probationAreaDescription?: string
+
+  // Email address - from Nomis, Delius or Auth depending on source
+  emailAddress?: string
+}
+
```
timharrison-moj, line 38: The object stored in the session, representing a superset of Nomis, Delius and Auth user details.

server/middleware/authorisationMiddleware.ts
```diff
@@ -9,21 +9,25 @@ const isAuthorisedRole = (role: string): boolean =>
...
 export default function authorisationMiddleware(req: Request, res: Response, next: NextFunction): void {
-  if (res.locals && res.locals.user && res.locals.user.token) {
-    const { authorities: roles = [] } = jwtDecode(res.locals.user.token) as { authorities?: string[] }
+  if (res.locals?.user?.token) {
+    let roles = res.locals.user?.userRoles
+    if (!roles) {
+      const { authorities: tokenRoles = [] } = jwtDecode(res.locals.user.token) as { authorities?: string[] }
+      roles = tokenRoles
+    }
+
```
timharrison-moj, line 18: Belt and braces here - the useRoles should always be populated.

server/utils/utils.ts
```diff
@@ -100,6 +100,11 @@ const jsonDtTo12HourTime = (dt: string): string => {
   return momentTime.isValid() ? momentTime.format('hh:mm a') : null
 }
 
+const removeDuplicates = (list: string[]): string[] => {
+  const deDuplicatedList = [...new Set(list.map(s => s))].map(s => s)
```
smcveigh941, line 104: What does `.map()` do here? Isn't this enough?

```
return [...new Set(list)]
```

server/services/index.ts
```diff
@@ -5,9 +5,11 @@ import CommunityService from './communityService'
 import HmppsAuthClient from '../data/hmppsAuthClient'
 import TokenStore from '../data/tokenStore'
 import CaseloadService from './caseloadService'
+import PrisonApiClient from '../data/prisonApiClient'
 
 const hmppsAuthClient = new HmppsAuthClient(new TokenStore())
-const userService = new UserService(hmppsAuthClient)
+const prisonApiClient = new PrisonApiClient('no admin token required for user service')
```
smcveigh941, line 11: Putting this string as a parameter will actually result in the requests to prison API having this header?

`Authorization: Bearer no admin token required for user service`

server/data/hmppsAuthClient.ts
```diff
@@ -32,46 +46,36 @@ function getSystemClientTokenFromHmppsAuth(username?: string): Promise<superagen
...
-
-export interface UserRole {
-  roleCode: string
-}
-
 export default class HmppsAuthClient {
   constructor(private readonly tokenStore: TokenStore) {}
 
-  private restClient(token: string): RestClient {
+  public restClient(token: string): RestClient {
```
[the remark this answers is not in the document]
timharrison-moj: Yeah I've been confused by the different implementation of these, and it really needs sorting out so they are all consistent, can be mocked easily, and can operate with either an admin client token or the user's own token when necessary. I've been mulling this over, but didn't want to extend this work any further. 3 days and counting...

server/utils/utils.ts
timharrison-moj: I think you're right. I've got some tests around this so can try a few alternatives in the morning.

PR #90, opened 2021-11-03
CVSL-184: Add more additional condition child pages and form types

smcveigh941:
- Added `jest.setup.ts` to import `reflect-metadata` instead of needing to import it in every test
- In my next PR I intend to abstract the nunjucks filters into their own unit functions that can be more easily unit tested and mocked

### DIFF
--- a/integration_tests/pages/additionalConditionInput.ts
+++ b/integration_tests/pages/additionalConditionInput.ts
@@ -1,3 +1,4 @@
+import moment from 'moment'
 import Page from './page'
 import BespokeConditionsQuestionPage from './bespokeConditionsQuestion'
 import { Context } from '../support/context'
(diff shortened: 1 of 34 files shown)


integration_tests/pages/additionalConditionInput.ts
```diff
@@ -26,6 +27,33 @@ export default class AdditionalConditionsInputPage extends Page {
     return this
   }
 
+  enterTime = (): AdditionalConditionsInputPage => {
+    cy.get("input[name*='hour']").type('11')
+    cy.get("input[name*='minute']").type('30')
```
timharrison-moj, line 32: What does the star do on name* ?

timharrison-moj: Just a wildcard match on name as it appears?

server/config/conditions.ts
```diff
@@ -449,41 +533,106 @@ export default {
...
+          label: 'Enter the timeframe',
+          name: 'timeframe',
+        },
+        {
+          type: InputTypes.DATE,
+          name: 'endDate',
+        },
+      ],
+      type: ElectronicMonitoringPeriod,
+    },
```
timharrison-moj, line 605: This condition has only one input field, but two values? Is this an either-or scenario? If a time-frame, the sentence won't make sense!

smcveigh941: Oh! good spot, I hadn't realised the condition text had changed content. The inputs are correct, the text is not

integration_tests/pages/additionalConditionInput.ts
smcveigh941: This jQuery selector matches any input where the name attribute contains some text. For example we have inputs called `appointmentTime[hour]`  `inductionDate[hour]` `appointmentTime[minute]`  `inductionDate[minute]` etc. This selector matches any input with hour/minute as a field

smcveigh941: https://api.jquery.com/attribute-contains-selector/

server/services/licenceService.ts
```diff
@@ -159,12 +160,23 @@ export default class LicenceService {
...
+        if (Array.isArray(formData[key])) {
+          return formData[key].map((v: string, i: number) => {
+            return build(v, i)
+          })
+        }
+        if (formData[key] instanceof Stringable) {
+          return build(formData[key].stringify())
         }
+        return build(formData[key])
       }),
```
timharrison-moj, line 180: Perhaps worth a few comments around this part? e.g. Why some formData elements are arrays, others Stringable, others neither?

timharrison-moj (dismissed):
LGTM. A bit hard to follow in one place, hence the comment suggestion there, but otherwise looks good.

server/services/licenceService.ts
smcveigh941: Added a comment now and rejigged the logic a bit!

timharrison-moj (approved):
LGTM

PR #91, opened 2021-11-04
CVSL-184: Added conditional radio button

smcveigh941:
- Added conditional reveal inputs nested beneath radio buttons
- Recursive formBuilder nunjucks macro which passes in the conditional reveal inputs
- Added a field decorator for validating that a time entered is after another time entered
- Conditional reveal field inputs have their parent's value stored in DB and displayed on the check your answers page, as below. Cannot easily overwrite the value of the radio button with the value of the conditional reveal because then it's hard to reconstruct the form when arriving from the change links.

screencapture-localhost-3000-licence-create-id-1-additional-conditions-condition-87-2021-11-04-21_14_48
Screenshot 2021-11-04 at 21 12 52
screencapture-localhost-3000-licence-create-id-1-additional-conditions-condition-87-2021-11-04-21_17_05

### DIFF
--- a/integration_tests/integration/createLicence.spec.ts
+++ b/integration_tests/integration/createLicence.spec.ts
@@ -40,6 +40,7 @@ context('Create a licence', () => {
       .selectCondition('fce34fb2-02f4-4eb0-9b8d-d091e11451fa')
       .selectCondition('a7c57e4e-30fe-4797-9fe7-70a35dbd7b65')
       .selectCondition('89e656ec-77e8-4832-acc4-6ec05d3e9a98')
+      .selectCondition('0a370862-5426-49c1-b6d4-3d074d78a81a')
       .clickContinue()
 
     const bespokeConditionsQuestionPage = additionalConditionsInputPage
@@ -53,6 +54,11 @@ context('Create a licence', () => {
       .enterAddress()
       .nextInput()
       .checkBoxes()
+      .nextInput(false) // aria-expanded attribute causes issues with Axe
+      .enterTime('10', '30', 'curfewStart')
+      .enterTime('11', '30', 'curfewEnd')
+      .selectRadios(3)
+      .enterText('Annually', 'alternativeReviewPeriod')
       .clickContinue()
 
     const bespokeConditionsPage = bespokeConditionsQuestionPage.selectYes().clickContinue()
--- a/server/services/licenceService.ts
+++ b/server/services/licenceService.ts
@@ -160,27 +160,29 @@ export default class LicenceService {
     const token = await this.hmppsAuthClient.getSystemClientToken(username)
 
     const requestBody = {
-      data: Object.keys(formData).flatMap((key, index) => {
-        // The POST request to the API will only accept an array of objects where value is a string.
-        // Therefore, if the type of data entered from the form is an array or an object, we need to convert that to a string type.
-        // For arrays, we can just split it into multiple objects each with a value from each member of that array.
-        // For objects, the objects needs to be converted into a string representation. Types which need to be represented as a string
-        // should extend Stringable and implement the stringify() method
+      data: Object.keys(formData)
+        .filter(key => formData[key])
+        .flatMap((key, index) => {
+          // The POST request to the API will only accept an array of objects where value is a string.
+          // Therefore, if the type of data entered from the form is an array or an object, we need to convert that to a string type.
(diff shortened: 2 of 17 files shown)


server/services/licenceService.ts
```diff
@@ -160,27 +160,29 @@ export default class LicenceService {
...
 
     const requestBody = {
-      data: Object.keys(formData).flatMap((key, index) => {
-        // The POST request to the API will only accept an array of objects where value is a string.
-        // Therefore, if the type of data entered from the form is an array or an object, we need to convert that to a string type.
-        // For arrays, we can just split it into multiple objects each with a value from each member of that array.
-        // For objects, the objects needs to be converted into a string representation. Types which need to be represented as a string
-        // should extend Stringable and implement the stringify() method
+      data: Object.keys(formData)
+        .filter(key => formData[key])
```
smcveigh941, line 164: Filtering out any undefined fields (i.e. where fields are hidden under a conditional radio button which has not been selected)

server/validators/isSimpleTimeAfter.ts
```diff
@@ -0,0 +1,25 @@
+import { registerDecorator, ValidationOptions, ValidationArguments } from 'class-validator'
+import SimpleTime from '../routes/creatingLicences/types/time'
+
+export default function IsSimpleTimeAfter(property: string, validationOptions?: ValidationOptions) {
```
smcveigh941, line 4: Field decorator which takes in another SimpleTime object and compares it. Will pass if the field being validated is a time after the field its being compared against

```
earlierTime: SimpleTime

@IsSimpleTimeAfter('earlierTime', { message: 'nowTime should be after the earlierTime' })
nowTime: SimpleTime 
```

server/views/formBuilder.njk
```diff
@@ -5,162 +5,175 @@
 {% from "partials/datePicker.njk" import datePicker %}
 {% from "partials/timePicker.njk" import timePicker %}
 
-{% for input in config.inputs %}
```
smcveigh941, line 8: Made this into a macro and indented, thats why alot of lines "changed" here

server/services/licenceService.ts
```diff
@@ -160,27 +160,29 @@ export default class LicenceService {
...
+        .flatMap((key, index) => {
+          // The POST request to the API will only accept an array of objects where value is a string.
+          // Therefore, if the type of data entered from the form is an array or an object, we need to convert that to a string type.
+          // For arrays, we can just split it into multiple objects each with a value from each member of that array.
+          // For objects, the objects needs to be converted into a string representation. Types which need to be represented as a string
+          // should extend Stringable and implement the stringify() method
 
-        const build = (value: unknown, i?: number) => {
```
smcveigh941, line 170: The rest of this is just indented

server/validators/decorators.ts
```diff
@@ -0,0 +1,15 @@
+import { registerDecorator, ValidationOptions } from 'class-validator'
```
smcveigh941, line 1: Field decorator which takes in another SimpleTime object and compares it. Will pass if the field being validated is a time after the field its being compared against

```
earlierTime: SimpleTime

@IsSimpleTimeAfter('earlierTime', { message: 'nowTime should be after the earlierTime' })
nowTime: SimpleTime 
```

smcveigh941: *facepalm* spent time doing this then realised we probably NEED to be able to put in for example 7pm in the evening curfewStart and 7am in the morning curfew end _**the next day**_

Oh well, I'm going to leave the decorator in for now in case it comes in useful or so that we have a reference on how to do custom decorators later on

integration_tests/integration/createLicence.spec.ts
```diff
@@ -53,6 +54,11 @@ context('Create a licence', () => {
       .enterAddress()
       .nextInput()
       .checkBoxes()
+      .nextInput(false) // aria-expanded attribute causes issues with Axe
+      .enterTime('10', '30', 'curfewStart')
```
timharrison-moj, line 58: This might be a legitimate accessibility issue. But one for the design guys to worry about.

server/validators/decorators.ts
timharrison-moj: Agree - leave it in as the reference example.

timharrison-moj (approved):
LGTM. Its getting quite complex this stuff!

PR #93, opened 2021-11-07
CVSL-199 expand additional conditions data into placeholders

timharrison-moj:
This PR does a few things:
* Adds templates into the configuration for additional conditions which require input.
* Expands placeholders with their values prior to printing a licence
* Caters for data item case (lower, upper, capitalise) in expanded values
* Caters for lists of data items, formatted with commas and/or 'and' between values.
* Tests for various types of additional conditions and their value replacement.


server/utils/conditionsProvider.ts
```diff
@@ -35,3 +37,156 @@ export function getGroupedAdditionalConditions(): Record<string, unknown>[] {
...
+  const expandedConditions: string[] = []
+
+  conditions?.forEach(condition => {
+    const configCondition = getAdditionalConditionByCode(condition.code)
+    if (configCondition?.requiresInput) {
+      const placeholders = getPlaceholderNames(configCondition.tpl as string)
+      const inputConfig = configCondition.inputs as unknown as Record<string, unknown>[]
+      let thisConditionText = configCondition.tpl as string
+
+      for (let i = 0; i < placeholders.length; i += 1) {
```
smcveigh941, line 56: `placeholders.forEach()` ?

server/utils/conditionsProvider.ts
```diff
@@ -35,3 +37,156 @@ export function getGroupedAdditionalConditions(): Record<string, unknown>[] {
...
+          // Single matching value
+          const rules = inputConfig.find(item => item.name === ph)
+          const { value } = matchingDataItems[0]
+          thisConditionText = replacePlaceholderWithValue(
+            ph,
+            thisConditionText,
+            adjustCase(rules.case as string, value)
+          )
+        } else {
+          // List of values for this placeholder
```
smcveigh941, line 76: I might be inclined to group fields by the same name into one single string before doing this processing. Doing that preprocessing would mean you don't need your else block here. I have this function in my PR to group fields into one string. But not to worry

```
    const map = new Map()
    data.forEach(item => {
      const collection = map.get(item.field)
      if (!collection) {
        map.set(item.field, [item])
      } else {
        collection.push(item)
      }
    })
    return Array.from(map, ([fieldName, items]) => ({ fieldName, items })).map(groupedDataItems => {
      return groupedDataItems.items.map((dataItem: AdditionalConditionData) => dataItem.value).join(', ')
    })
```

server/utils/conditionsProvider.ts
```diff
@@ -35,3 +37,156 @@ export function getGroupedAdditionalConditions(): Record<string, unknown>[] {
...
+}
+
+/**
+ * Formats a list of data items according to rules (add commas, and)
+ * @param matchingDataItems
+ */
+export const produceValueAsFormattedList = (matchingDataItems: AdditionalConditionData[]): string => {
+  let value = ''
+  let valueCounter = 1
+  for (let i = 0; i < matchingDataItems.length; i += 1) {
```
smcveigh941, line 105: matchingDataItems.forEach(item => {})

server/utils/conditionsProviderExpand.test.ts
```diff
@@ -0,0 +1,151 @@
...
+          code: 'ed607a91-fe3a-4816-8eb9-b447c945935c',
+          category: 'Possession, ownership, control or inspection of specified items or documents',
+          sequence: 1,
+          text: 'Not to own or use a camera without the prior approval of your supervising officer.',
+          data: [],
+        },
+      ]
+      const listOfConditions = expandAdditionalConditions(conditions)
+      expect(listOfConditions).toHaveLength(1)
+      expect(listOfConditions[0]).toEqual(conditions[0].text)
```
smcveigh941, line 25: Always inclined to use string literals over variables in assertions

server/utils/conditionsProviderExpand.test.ts
```diff
@@ -0,0 +1,151 @@
...
+        },
+      ]
+      const listOfConditions = expandAdditionalConditions(conditions)
+      expect(listOfConditions).toHaveLength(1)
+      expect(listOfConditions[0]).toEqual(
+        'To comply with any requirements specified by your supervising officer for the purpose of ensuring that you address your alcohol, drug, sexual, violent, gambling and anger problems.'
+      )
+    })
+  })
+})
```
smcveigh941, line 151: Good tests!

smcveigh941 (approved):
Looks sensible to me, some suggestions

server/utils/conditionsProvider.ts
timharrison-moj: I did this first time around, however it never goes into the loop! Something odd going on with Array.forEach in these circumstances, which I was thinking I would come back to once the code and tests were in and working.

server/utils/conditionsProvider.ts
timharrison-moj: Its the same thing here - I had it like this originally, but it never went into the forEach loop, despite it having values.

server/utils/conditionsProviderExpand.test.ts
timharrison-moj: In this case, its taking the .text verbatim - so should be identical. Agree where there are changes being made.

server/utils/conditionsProviderExpand.test.ts
timharrison-moj: Thanks!

server/utils/conditionsProvider.ts
timharrison-moj: Any ideas appreciated why the .forEach might not work here?

server/utils/conditionsProvider.ts
timharrison-moj: I'll take a another look at this now.

server/utils/conditionsProvider.ts
timharrison-moj: Done now - next PR.

PR #94, opened 2021-11-07
CVSL-184: Form builder option for "Add another" button

smcveigh941:
- Adding remaining conditions
- Transformer function to remove empty strings from array inputs (i.e. empty 'Add another' fields)
- Re-jigged how form values are populated on formBuilder

### DIFF
--- a/integration_tests/mockApis/licence.ts
+++ b/integration_tests/mockApis/licence.ts
@@ -163,6 +163,27 @@ export default {
                 },
               ],
             },
+            {
+              id: 5,
+              code: '3932e5c9-4d21-4251-a747-ce6dc52dc9c0',
+              category: 'Possession, ownership, control or inspection of specified items or documents',
+              sequence: 4,
+              text: 'Not to own or possess a [SPECIFIED ITEM] without the prior approval of your supervising officer.',
+              data: [
+                {
+                  id: 10,
+                  sequence: 0,
+                  field: 'item',
+                  value: 'Knives',
+                },
+                {
+                  id: 11,
+                  sequence: 1,
+                  field: 'item',
+                  value: 'Needles',
+                },
+              ],
+            },
           ],
           bespokeConditions: [
             {
--- a/server/config/conditions.ts
+++ b/server/config/conditions.ts
@@ -18,6 +18,10 @@ import AlcoholMonitoringPeriod from '../routes/creatingLicences/types/additional
 import CurfewTerms from '../routes/creatingLicences/types/additionalConditionInputs/curfewTerms'
 import CurfewAddress from '../routes/creatingLicences/types/additionalConditionInputs/curfewAddress'
 import NoContactWithVictim from '../routes/creatingLicences/types/additionalConditionInputs/noContactWithVictim'
+import ReportToStaff from '../routes/creatingLicences/types/additionalConditionInputs/reportToStaff'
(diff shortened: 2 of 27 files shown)


integration_tests/mockApis/licence.ts
```diff
@@ -163,6 +163,27 @@ export default {
...
+                  value: 'Knives',
+                },
+                {
+                  id: 11,
+                  sequence: 1,
+                  field: 'item',
+                  value: 'Needles',
+                },
+              ],
+            },
```
timharrison-moj, line 186: Is this going to contain `knives OR needles` rather than 'knives AND needles`? Might need another input rule on lists that are 'or' type lists?

server/config/conditions.ts
```diff
@@ -546,13 +622,51 @@ export default {
...
+                  },
+                ],
+              },
+            },
+          ],
+        },
+      ],
+      type: ReportToStaff,
+    },
     {
```
timharrison-moj, line 670: I'll visit these to add the required .tpl templates after this goes in.

server/utils/nunjucksSetup.ts
```diff
@@ -183,5 +144,30 @@ export function registerNunjucks(app?: express.Express): Environment {
...
+  njkEnv.addGlobal('getFormResponses', (formResponses: unknown, inputName: string) => {
+    return formResponses ? [formResponses[inputName]].flat() : undefined
+  })
+
+  njkEnv.addGlobal('getValueFromAdditionalCondition', (additionalCondition: AdditionalCondition, inputName: string) => {
+    return additionalCondition
+      ? additionalCondition.data.filter(item => item.field === inputName).map(item => item.value)
+      : undefined
+  })
+
```
timharrison-moj, line 171: Are there tests for the newly added filters?

timharrison-moj: OK as a follow-on piece if you want to get this in.

smcveigh941: In my next PR I want to separate these functions away from nunjucks, and have them as reusable utils. This means they can be used elsewhere in our code (not just nunjucks) and also makes them more easily unit tested. When I do that I will increase coverage on them

server/views/formBuilder.njk
```diff
@@ -2,178 +2,219 @@
...
-            }) %}
-                {{ govukInput({
-                    id: 'addressLine1',
-                    label: {
-                        html: 'Building and street <span class="govuk-visually-hidden">line 1 of 2</span>'
+        {% set validationError = validationErrors | findError(input.name) %}
+        {% set formValues = getFormResponses(formResponses, input.name) or getValueFromAdditionalCondition(additionalCondition, input.name) %}
+
+        <div data-module="moj-add-another">
+            {% if input.type == 'text' %}
```
timharrison-moj, line 18: Will this always create a div for 'moj-add-another' ? For ALL input types?  (No surrounding {% if  %})

timharrison-moj (approved):
LGTM. Couple of questions only.

server/views/formBuilder.njk
smcveigh941: The `data-module` attribute will enable the add another client side javascript functionality for all data types, but it will not show the button unless addAnother is provided in the field config. 

So yes even if addAnother is not provided in the config, it will be wrapped in a div with the module add another, but it will not have any effect. In the interest of keeping this already complicated form builder as simple as possible, I went with wrapping all inputs in this

PR #95, opened 2021-11-08
CVSL-209 missing image placeholder in licence documents

timharrison-moj:
When no offender image exists, this sends the placeholder image instead.


smcveigh941 (approved):
LGTM

PR #96, opened 2021-11-08
CVSL-206 - use the approvers full name in signatures

timharrison-moj:
* Use stored approver's name in signatures
* Supply approvers full name in status change for approval.
* Add missing .tpl templates for newly entered conditions.


server/config/conditions.ts
```diff
@@ -194,12 +194,14 @@ export default {
       code: '4858cd8b-bca6-4f11-b6ee-439e27216d7d',
       category: 'Making or maintaining contact with a person',
       text: 'Not to seek to approach or communicate with [INSERT NAME OF VICTIM AND / OR FAMILY MEMBERS] without the prior approval of your supervising officer and / or [INSERT NAME OF APPROPRIATE SOCIAL SERVICES DEPARTMENT].',
+      tpl: 'Not to seek to approach or communicate with {name} without the prior approval of your supervising officer and / or {socialServicesDepartment}.',
```
smcveigh941, line 197: So here, social services department is optional... I guess thats in the pipeline

smcveigh941 (approved):
LGTM

server/config/conditions.ts
timharrison-moj: Yes, this one and the one where we have an Other review period are not done yet.

PR #102, opened 2021-11-09
CVSL-208: Integration test for view and print licence journey.

timharrison-moj:
This PR:
* Creates page objects for view cases, view licence and HTML printed licence
* Implements a click-through journey to view and print a licence
* Verifies accessibility of pages
* Verifies that the printed licence contains key structure and data items.
* It overrides the PDF print link with the HTML version (same stylesheet, structure and content, just easier to verify :-) )


smcveigh941 (approved):
LGTM

PR #103, opened 2021-11-10
CVSL-220: Remove "Add another" from social services department fields in additional conditions

smcveigh941:
- Removed "Add another" options from social services department as agreed
- Do not persist empty date/time/address in the case that they are optional fields
- Configured `docker-compose.yml` to run `create-and-vary-a-licence-api` as a container locally for convenience & local dev (still uses auth in dev)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,7 +29,7 @@ services:
     image: quay.io/hmpps/create-and-vary-a-licence-api:latest
     networks:
       - hmpps
-    container_name: creare-and-vary-a-licence-api
+    container_name: create-and-vary-a-licence-api
     restart: always
     ports:
       - "8089:8080"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,19 +11,6 @@ services:
     ports:
       - '6379:6379'
 
-  hmpps-auth:
-    image: quay.io/hmpps/hmpps-auth:latest
-    networks:
-      - hmpps
-    container_name: hmpps-auth 
-    ports:
-      - "9090:8080"
-    healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8080/auth/health"]
-    environment:
-      - SPRING_PROFILES_ACTIVE=dev
-      - DELIUS_ENABLED=false
-
   # For the API to seed & interact with - there is no direct connection from the UI service
   licences-db:
     image: postgres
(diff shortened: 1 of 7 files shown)


server/routes/creatingLicences/types/address.ts
```diff
@@ -24,7 +24,8 @@ class Address extends Stringable {
   addressPostcode: string
 
   stringify(): string {
-    return Object.values(this).join(', ')
+    if (Object.values(this).join('').length > 0) return Object.values(this).join(', ')
```
smcveigh941, line 27: Added this so that address is not persisted to the database at all if its empty. Previously it was persisted as ", , , , "

server/routes/creatingLicences/types/time.ts
```diff
@@ -25,10 +25,7 @@ class SimpleTime extends Stringable {
   ampm: AmPm
 
   stringify(): string {
-    if (this.hour && this.minute && this.ampm) {
```
smcveigh941, line 28: No longer needed because the licence service checks for an empty object before persisting to database, so no need to convert an empty object to an empty string

mikehalmamoj (approved):

docker-compose.yml
```diff
@@ -38,17 +25,19 @@ services:
...
 
-  prison-api:
-    image: quay.io/hmpps/prison-api:latest
+  create-and-vary-a-licence-api:
+    image: quay.io/hmpps/create-and-vary-a-licence-api:latest
     networks:
       - hmpps
-    container_name: prison-api
+    container_name: creare-and-vary-a-licence-api
+    restart: always
```
timharrison-moj, line 33: Spelling?

docker-compose.yml
```diff
@@ -11,19 +11,6 @@ services:
...
-      - hmpps
-    container_name: hmpps-auth 
-    ports:
-      - "9090:8080"
-    healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8080/auth/health"]
-    environment:
-      - SPRING_PROFILES_ACTIVE=dev
-      - DELIUS_ENABLED=false
-
```
timharrison-moj, line 26: Do we want to remove auth from docker-compose.yml? I'm not sure anyone now runs the entire thing locally, as requires too many containers. However, maybe we still could? Or create two -dev and -dev-api.yaml files for different flavours of running locally?

timharrison-moj (approved):
Couple of questions / observations around the docker-compose usage & one typo.

docker-compose.yml
smcveigh941: We dont have the required roles in auth I thought? So if we want to keep it for local we need to add the roles in

docker-compose.yml
smcveigh941: Fixed now thanks

PR #104, opened 2021-11-10
CVSL-224: Add condition 9B

smcveigh941:
- Added condition 9B
- Added addAnother button to conditions 9A and 9B, would like to query these with service design as it doesn't seem to make sense to me but I have build as the ticket suggests, I can revert if design changes after discussion

### DIFF
--- a/server/config/conditions.ts
+++ b/server/config/conditions.ts
@@ -18,10 +18,11 @@ import AlcoholMonitoringPeriod from '../routes/creatingLicences/types/additional
 import CurfewTerms from '../routes/creatingLicences/types/additionalConditionInputs/curfewTerms'
 import CurfewAddress from '../routes/creatingLicences/types/additionalConditionInputs/curfewAddress'
 import NoContactWithVictim from '../routes/creatingLicences/types/additionalConditionInputs/noContactWithVictim'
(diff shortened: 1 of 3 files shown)


server/config/conditions.ts
```diff
@@ -677,6 +678,62 @@ export default {
...
+      tpl: 'Report to staff at {policeStation} at {reportingTime}, unless otherwise authorised by your supervising officer. This condition will be reviewed by your supervising officer on a {reviewPeriod} {alternativeReviewPeriod} basis and may be amended or removed if it is felt that the level of risk you present has reduced appropriately.',
+      requiresInput: true,
+      inputs: [
+        {
+          type: InputTypes.TEXT,
+          label: 'Enter name of police station',
+          name: 'policeStation',
+          case: 'capitalised',
+          listType: 'AND',
+          addAnother: {
```
timharrison-moj, line 734: Is policeStation a list or single data item? If single, could lose the listType: AND.

timharrison-moj: It does look like a list in config/validation, so ignore this one ^

timharrison-moj (approved):
LGTM

PR #106, opened 2021-11-11
CVSL-224: Remove add another from conditions 9A and 9B

smcveigh941:
- No longer need @RemoveEmptyArrayItems annotation, checked in service layer instead
- In form builder, checked that the input was "addAnother" enabled before building multiple fields, needed to do this in the case of a field being made a single field in config when it has already been "addAnother" in production


server/views/formBuilder.njk
```diff
@@ -216,32 +216,34 @@
...
-            classes: 'moj-add-another__title govuk-visually-hidden',
-            isPageHeading: false
-        }
-    }) %}
-        {{ govukInput({
-            id: input.name if not input.addAnother else input.name + '[' + index + ']',
-            name: input.name if not input.addAnother else input.name + '[' + index + ']',
-            classes: "govuk-!-width-one-half",
-            label: {
+    {% if index == 0 or (index > 0 and input.addAnother) %}
```
smcveigh941, line 219: Most of this file is just indented, only wrapped in this IF

mikehalmamoj (approved):

PR #107, opened 2021-11-11
CVSL-215: Expanding the trixy conditions that alter the sentence text.

timharrison-moj:
5 of the additional conditions include additional text to make sentences make sense when optional values are either present or omitted. This PR deals with those by using `.includeBefore` fixed text which is prepended to the value when it is present, or omitted when the value is not present.

### DIFF
--- a/server/config/conditions.ts
+++ b/server/config/conditions.ts
@@ -74,7 +74,7 @@ export default {
     {
       code: '5db26ab3-9b6f-4bee-b2aa-53aa3f3be7dd',
       category: 'Residence at a specific place',
-      text: 'You must reside within [INSERT REGION] while of no fixed abode, unless otherwise approved by your supervising officer.',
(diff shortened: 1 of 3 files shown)


server/config/conditions.ts
```diff
@@ -75,7 +75,7 @@ export default {
       code: '5db26ab3-9b6f-4bee-b2aa-53aa3f3be7dd',
       category: 'Residence at a specific place',
       text: 'You must reside within [INSERT REGION] while of no fixed abode, unless otherwise approved by your supervising officer.',
```
smcveigh941, line 77: The text which shows on the screens should match what is on the resultant licence maybe?

timharrison-moj: yes, it should. It altered the mean when selecting from drop down, so I'll update this to include `the [value] probation region`

timharrison-moj: Pushed now.

smcveigh941 (approved):
LGTM

PR #111, opened 2021-11-16
CVSL-55 HDC eligibility checks and filters

timharrison-moj:
This alters alters the HDC eligibility checks.

* The presence of a HDC end date is no longer the sole check.
* If the HDC end date is not present - its valid for our service.
* If the HDC end date is present but there is no HDC status in Nomis - its valid for our service.
* If the HDC status is present and checks have failed - its valid for our service.
* If the HDC status is present, and checks have passed but approvalStatus is REJECTED - its valid for our service.
* Any other result from HDC checks will mean the case is filtered from our service.

NOTE: The HDC endpoint is a single bookingId and we need to do this for a list of cases - so it is hit multiple times - raised as team issue - and in discussion. Would be better to do this after selecting a case from the list! i.e. land on a "profile" or "checking" page before starting to create a licence.


server/@types/HdcStatus.ts
```diff
@@ -0,0 +1,17 @@
+class HdcStatus {
+  public constructor(
+    bookingId: string,
+    private readonly homeDetentionCurfewEndDate: string = null,
+    private readonly checksPassed: boolean = false,
```
smcveigh941, line 5: Dont think you need to specify the type as it is inferred by the default value

server/services/caseloadService.ts
```diff
@@ -49,7 +52,10 @@ export default class CaseloadService {
       .filter(offender => offender.status && offender.status.startsWith('ACTIVE'))
       .filter(offender => !offender.indeterminateSentence && offender.conditionalReleaseDate)
       .filter(offender => !offender.releaseDate || moment().isBefore(moment(offender.releaseDate, 'YYYY-MM-DD')))
-      .filter(offender => !offender.homeDetentionCurfewEndDate)
+      .filter(offender => {
+        const hdcStatus = hdcStatuses.find(hdc => hdc.bookingId === offender.bookingId)
+        return hdcStatus?.eligibleForHdc ? !hdcStatus.eligibleForHdc : true
```
smcveigh941, line 57: Is this more convoluted than it needs to be? is this enough?

```
return !hdcStatus?.eligibleForHdc
```

server/services/prisonerService.ts
```diff
@@ -56,4 +57,23 @@ export default class PrisonerService {
...
   }
+
+  async getHdcStatuses(username: string, offenders: Prisoner[]): Promise<HdcStatus[]> {
+    const listOfHdcStatus: HdcStatus[] = []
+    const token = await this.hmppsAuthClient.getSystemClientToken(username)
+    const client = new PrisonApiClient(token)
+    await Promise.all(
+      offenders.map(async o => {
+        try {
+          const hdc = await client.getLatestHdcStatus(o.bookingId)
```
smcveigh941, line 68: Just wonder if there's a limit to how many async threads node can handle at once, I dont know the answer

smcveigh941 (approved):
Looks ok to me, few thoughts

smcveigh941:
oooo, a "checking" page before starting the journey would be awesome. Would solve that "button disguised as a link" issue on the caseload page

server/services/prisonerService.ts
timharrison-moj: Seems ok at 20-30. I don't like this either, and suggested to David and Jon that the caseload page is probably the wrong one, or to adjust the prison-API endpoint to a POST accepting a list of booking Ids. Its short-term, but working for now.

server/services/caseloadService.ts
timharrison-moj: Yes, probably! I'll alter and verify with the tests :-)

server/@types/HdcStatus.ts
timharrison-moj: No - the default value was added last minute during testing, so no it doesn't need the type. I will change

timharrison-moj: Actually, it complains it could be type any without specifying it. Worth a check though.

server/services/prisonerService.ts
timharrison-moj: The best solution would be an intermediate page after selection where the offenders/ licences are narrowed down to one.

smcveigh941 (approved):

PR #112, opened 2021-11-16
CVSL-216: Confirmation page for PSS licences

smcveigh941:
- disabled lint rule for a default case for switch statement with predetermined values
- Logic in confirmation page handler to determine the message


server/routes/creatingLicences/handlers/confirmation.ts
```diff
@@ -1,7 +1,28 @@
...
+        break
+      case LicenceType.AP:
+        titleText = `Licence conditions for ${licence.forename} ${licence.surname} sent`
+        confirmationMessage = `We have sent the licence to ${licence.prisonDescription} for approval.`
+        break
+      case LicenceType.PSS:
+        titleText = `Post sentence supervision order for ${licence.forename} ${licence.surname} sent`
+        confirmationMessage = `We have sent the post sentence supervision order to ${licence.prisonDescription} for approval.`
+        break
+    }
```
timharrison-moj, line 24: We're always sure its one of these from the enum, so ok without default.

timharrison-moj: Could disable lint for default on this switch, rather then service wide? Ok for now though.

timharrison-moj (approved):
LGTM.

PR #118, opened 2021-11-18
CVSL-27: Use nonce value for scripts started by google tag manager

timharrison-moj:
Last try for today..


server/middleware/setUpWebSecurity.ts
```diff
@@ -24,10 +24,17 @@ export default function setUpWebSecurity(): Router {
             "'self'",
             (req, res: Response) => `'nonce-${res.locals.cspNonce}'`,
             'code.jquery.com',
+            'www.googletagmanager.com',
+            'www.google-analytics.com',
```
smcveigh941, line 28: need `https://` on these ones?

timharrison-moj: No i don't think so - had it blocking them before this change locally, and allowed them after.

smcveigh941 (approved):
Give it a try :)

PR #120, opened 2021-11-19
CVSL-237 security fixes

timharrison-moj:
* Updates typescript and node version.
* Suppresses issue in class-validator that has no fix yet.
* Hides the my caseload tile - in preparation for private beta.


smcveigh941 (approved):
Lgtm

PR #122, opened 2021-11-22
CVSL-245: Content changes for private beta + pdf watermark in dev

smcveigh941:
Example of the generated PDF with watermark

[WATERMARK.pdf](https://github.com/ministryofjustice/create-and-vary-a-licence/files/7584482/WATERMARK.pdf)

### DIFF
--- a/helm_deploy/create-and-vary-a-licence/values.yaml
+++ b/helm_deploy/create-and-vary-a-licence/values.yaml
@@ -41,6 +41,7 @@ generic-service:
       SYSTEM_CLIENT_SECRET: "SYSTEM_CLIENT_SECRET"
       SESSION_SECRET: "SESSION_SECRET"
       TAG_MANAGER_CONTAINER_ID: "TAG_MANAGER_CONTAINER_ID"
+      LICENCE_WATERMARK: false
     elasticache-redis:
       REDIS_HOST: "primary_endpoint_address"
       REDIS_AUTH_TOKEN: "auth_token"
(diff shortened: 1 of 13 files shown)


helm_deploy/create-and-vary-a-licence/values.yaml
```diff
@@ -41,6 +41,7 @@ generic-service:
       SYSTEM_CLIENT_SECRET: "SYSTEM_CLIENT_SECRET"
       SESSION_SECRET: "SESSION_SECRET"
       TAG_MANAGER_CONTAINER_ID: "TAG_MANAGER_CONTAINER_ID"
+      LICENCE_WATERMARK: false
     elasticache-redis:
```
timharrison-moj, line 45: Is this in the wrong place? These are secrets to read from the K8s secrets, so it will look for a secret called "false". I think it can just go into the `env:` section directly?

helm_deploy/values-dev.yaml
```diff
@@ -30,6 +30,8 @@ generic-service:
     # A reference to the CVL service - for links in HTML to get images/stylesheets (internal within namespace)
     LICENCES_URL: "http://create-and-vary-a-licence"
 
+    LICENCE_WATERMARK: true
+
```
timharrison-moj, line 34: This is in the right place.

timharrison-moj (commented):
LGTM - though comment added about the watermark setting - in wrong place in file?

helm_deploy/create-and-vary-a-licence/values.yaml
smcveigh941: Thanks, moved it now

timharrison-moj (approved):
LGTM.

PR #127, opened 2021-11-26
CVSL-198: Display status badge on caseload view for case admin

smcveigh941:


server/views/pages/view/cases.njk
```diff
@@ -26,6 +27,12 @@
         {
             attributes: { id: 'conditional-release-date-' + loop.index },
             text: case.conditionalReleaseDate | datetimeToDate
+        },
+        {
+            attributes: {
+                id: 'licence-status-' + loop.index
+            },
+            html: statusBadge({ status: case.licenceStatus, size: 'small' }, statusConfig)
         }
```
timharrison-moj, line 36: I wonder if, for consistency, the licence type should also be on the case list for CAs? Not a tech question, more of an inconsistent design question.

timharrison-moj (approved):
LGTM

PR #130, opened 2021-11-30
CVSL-174: Add job to activate approved licences

smcveigh941:
- k8s cronjob to update the status to ACTIVE, any licences which are APPROVED and within the licence period
- also disabled the codecov comment

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,9 +8,4 @@ coverage:
     patch:
       default:
         informational: true
-comment:
-  require_head: no
-  require_base: no
-  require_changes: false
-  layout: "reach,diff,files"
-  behavior: default
+comment: false
--- a/helm_deploy/create-and-vary-a-licence/templates/activate-licences-cronjob.yaml
+++ b/helm_deploy/create-and-vary-a-licence/templates/activate-licences-cronjob.yaml
@@ -1,0 +1,23 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: activate-licences
+spec:
+  schedule: "0 19 * * *"
+  concurrencyPolicy: Replace
+  failedJobsHistoryLimit: 5
+  startingDeadlineSeconds: 43200
+  successfulJobsHistoryLimit: 5
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: create-and-vary-a-licence
+            image: {{ index .Values "generic-service" "image.repository" }}
+            args:
+            - node
+            - dist/jobs/activateLicences
+            env: {{ index .Values "generic-service" "env" }}
+          restartPolicy: Never
+          activeDeadlineSeconds: 3600
--- a/helm_deploy/create-and-vary-a-licence/templates/activate-licences-cronjob.yaml
+++ b/helm_deploy/create-and-vary-a-licence/templates/activate-licences-cronjob.yaml
@@ -3,7 +3,7 @@ kind: CronJob
 metadata:
   name: activate-licences
 spec:
-  schedule: "0 19 * * *"
+  schedule: "*/10 * * * *"
   concurrencyPolicy: Replace
   failedJobsHistoryLimit: 5
   startingDeadlineSeconds: 43200
--- a/jobs/activateLicences.ts
+++ b/jobs/activateLicences.ts
@@ -1,16 +1,14 @@
-import superagent from 'superagent'
-import querystring from 'querystring'
 import moment from 'moment'
 import { initialiseAppInsights, buildAppInsightsClient, flush } from '../server/utils/azureAppInsights'
 import logger from '../logger'
 
-import generateOauthClientToken from '../server/authentication/clientCredentials'
-import config from '../server/config'
 import { LicenceSummary } from '../server/@types/licenceApiClientTypes'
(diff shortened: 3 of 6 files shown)


.codecov.yml
```diff
@@ -8,9 +8,4 @@ coverage:
     patch:
       default:
         informational: true
-comment:
-  require_head: no
-  require_base: no
-  require_changes: false
-  layout: "reach,diff,files"
-  behavior: default
```
timharrison-moj, line 16: You finally had enough of it then?!

helm_deploy/create-and-vary-a-licence/templates/activate-licences-cronjob.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: activate-licences
+spec:
+  schedule: "0 19 * * *"
+  concurrencyPolicy: Replace
```
timharrison-moj, line 7: I would suggest an early morning run time for this, or maybe even twice per day - early morning + after lunchtime?

.codecov.yml
smcveigh941: The comment's inconsistency annoys me.. cant figure out why it sometimes updates and sometimes doesnt 😆

helm_deploy/create-and-vary-a-licence/templates/activate-licences-cronjob.yaml
smcveigh941: Can do, this time now is temporary, I will have a follow-up PR to set the time to early morning once we can see it working

smcveigh941: 7pm this evening has passed.. so I'll set it again when ready to merge

timharrison-moj: Actually, looking at the code - it only checks against the date, not a date and time so a once-per-day run is enough.

smcveigh941: Unless there are licences which are only approved on the day they are supposed to be released!

jobs/activateLicences.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  logger.info(`HMPPS Auth request '${authRequest}' for client id '${config.apis.hmppsAuth.systemClientId}'`)
+
+  return superagent
+    .post(`${config.apis.hmppsAuth.url}/oauth/token`)
+    .set('Authorization', clientToken)
+    .set('content-type', 'application/x-www-form-urlencoded')
+    .send(authRequest)
+    .timeout(config.apis.hmppsAuth.timeout)
+}
```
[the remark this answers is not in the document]
smcveigh941: The existing one uses a Redis container, will it work for this? That Redis container would likely be available to this job?

jobs/activateLicences.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+  return new PrisonerSearchApiClient(authResponse.body.access_token).searchPrisonersByNomsIds(prisonerSearchCriteria)
+}
+
+const batchActivateLicences = async (licenceIds: number[]): Promise<void> => {
+  if (licenceIds.length > 0) {
+    const authResponse = await getSystemClientTokenFromHmppsAuth()
+    await new LicenceApiClient(authResponse.body.access_token).batchActivateLicences(licenceIds)
+  }
+}
```
timharrison-moj, line 74: What happens on a ValidationException? Caught and a 400 returned here?  I think it unlikely to happen between getting the list and calling the API, so maybe take out the ValidationException check in the API? Sorry, I know it was my suggestion but it complicates things and probably would never happen - particularly if the job runs early each morning.

helm_deploy/create-and-vary-a-licence/templates/activate-licences-cronjob.yaml
timharrison-moj: Aha. And there will be - so good shout - perhaps early evening to get today's releases.

jobs/activateLicences.ts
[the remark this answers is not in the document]
timharrison-moj: Yes its available, but if it complicates it, don't worry about it. Doesn't need session info - its a one-shot thing.

jobs/activateLicences.ts
timharrison-moj: I've since read the whole thing, and its being caught. But still think its probably something that wouldn't happen, and if it did, just skipping that one licence would be better than failing the job?

timharrison-moj (dismissed):
LGTM. Couple of comments only, but I think you've got those for the API change.

jobs/activateLicences.ts
smcveigh941: Yep it will bomb out.. I will remove the exception from the API and just log it instead

jobs/activateLicences.ts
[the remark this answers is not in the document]
smcveigh941: Now reusing the auth client, works fine locally

timharrison-moj (approved):
LGTM

timharrison-moj (approved):
LGTM

PR #132, opened 2021-12-01
CVSL-174: Helm deploy fix for cronjob

smcveigh941:
- Upgrade `generic-service` and `generic-prometheus-alerts`


timharrison-moj (approved):
LGTM

PR #136, opened 2021-12-03
CVSL-264: Adds a link to conditions with no child pages.

timharrison-moj:


server/views/partials/additionalLicenceConditionsSummary.njk
```diff
@@ -39,26 +27,45 @@
...
             actions: {
                 items: [
                     {
-                        href: "/licence/create/id/" + licence.id + "/additional-licence-conditions/condition/" + condition.id + "?fromReview=true",
+                        href: actionLink + "?fromReview=true",
                         text: "Change",
                         visuallyHiddenText: "Change condition"
                     }
                 ]
-            } if condition | checkConditionRequiresInput and isEditable
```
smcveigh941, line 53: Can you remove that filter if its no longer needed?

smcveigh941 (approved):
LGTM, maybe remove the redundant nunjucks filter?

server/views/partials/additionalLicenceConditionsSummary.njk
smcveigh941: Nevermind, its still used.

PR #137, opened 2021-12-03
CVSL-246: Add delete condition button to condition child pages

smcveigh941:


integration_tests/mockApis/licence.ts
```diff
@@ -24,7 +24,7 @@ const licencePlaceholder = {
   licenceStartDate: '01/04/2021',
   licenceExpiryDate: '26/04/2060',
   topupSupervisionStartDate: '26/04/2022',
-  topupSupervisionExpiryDate: '26/06/2022',
+  topupSupervisionExpiryDate: '26/06/2060',
```
smcveigh941, line 27: To avoid validation error during integration test

server/routes/creatingLicences/handlers/additionalPssConditionInput.ts
```diff
@@ -30,4 +31,27 @@ export default class AdditionalPssConditionInputRoutes {
...
+    const { username } = req.user
+
+    const additionalConditionCodes = licence.additionalPssConditions
+      .filter((condition: AdditionalCondition) => condition.id !== parseInt(conditionId, 10))
+      .map((condition: AdditionalCondition) => condition.code)
+
+    await this.licenceService.updateAdditionalConditions(
+      licence.id,
+      LicenceType.PSS,
+      { additionalConditions: additionalConditionCodes },
```
timharrison-moj, line 47: Wondered if we should have a specific enum for `ConditionType.Additional` and `ConditionType.PSS` in place of overloading the use of LicenceType.AP | .PSS ? This did catch me out for little while on a previous change until I realised.

timharrison-moj (approved):
LGTM. Small suggestion only, but could be another change.

server/routes/creatingLicences/handlers/additionalPssConditionInput.ts
smcveigh941: Yes we should, agree

PR #141, opened 2021-12-07
CVSL-272: Refactor REST client and API adapters + isolated test coverage

smcveigh941:
- Deleted code around the `/test/data` endpoint as its no longer needed
- Refactored API clients to use inheritance and dependency injection
- Increased test coverage of all clients
- Increase in coverage of service layers to come in a following PR this week
- Handlers now pass in the User object through the service layer and to the client. The clients then decide whether to use user provided token or system token


server/data/hmppsAuthClient.ts
```diff
@@ -24,58 +16,16 @@ export type AuthUserRole = {
...
-function getSystemClientTokenFromHmppsAuth(username?: string): Promise<superagent.Response> {
-  const clientToken = generateOauthClientToken(
-    config.apis.hmppsAuth.systemClientId,
-    config.apis.hmppsAuth.systemClientSecret
-  )
-
-  const authRequest = username
-    ? querystring.stringify({ grant_type: 'client_credentials', username })
-    : querystring.stringify({ grant_type: 'client_credentials' })
-
```
timharrison-moj, line 36: This is an important bit when calling the community-api - no username in system token. Hopefully its catered for in the changes below.. I'll check.

server/data/hmppsRestClient.ts
```diff
@@ -137,18 +137,20 @@ export default class RestClient {
     }
   }
 
-  async stream({ path = null, headers = {} }: StreamRequest = {}): Promise<unknown> {
+  async stream({ path = null, headers = {} }: StreamRequest, username?: string): Promise<unknown> {
+    const token = await this.tokenStore.getAuthToken(username)
+
```
timharrison-moj, line 142: How does this know whether to use/supply a) The user's own token, and where is this provided?, or b) A system token with a username, or c) A system token without a username (as required by the CommunityApi) ?

server/data/licenceApiClient.test.ts
```diff
@@ -1,412 +1,230 @@
...
 
-    describe('Bespoke conditions', () => {
-      it('No bespoke conditions entered', async () => {
-        const formConditions = makeFormBespokeConditions([])
-        const apiRequestConditions = makeApiRequestBespokeConditions([])
-        hmppsAuthClient.getSystemClientToken.mockResolvedValue('a token')
-        fakeApi.put('/licence/id/1/bespoke-conditions', apiRequestConditions).reply(200)
-        await licenceService.updateBespokeConditions('1', formConditions, username)
-        expect(nock.isDone()).toBe(true)
-        expect(hmppsAuthClient.getSystemClientToken).toBeCalled()
```
timharrison-moj, line 148: Are all these removed tests here being migrated into the LicenceService tests?

server/data/prisonApiClient.test.ts
```diff
@@ -1,247 +1,109 @@
...
-      firstName: 'Robert',
-      lastName: 'Charles',
-      lockedFlag: false,
-      staffId: 123,
-      thumbnailId: 345,
-      username: 'RCHARLES',
-    } as PrisonApiUserDetail
-
-    it('Get /api/users/me', async () => {
-      fakeApi.get('/api/users/me').reply(200, stubbedPrisonUserData)
```
timharrison-moj, line 137: This one should pass through the logged-in users own token. Can't see the equivalent (yet)?

server/data/prisonApiClient.test.ts
```diff
@@ -1,247 +1,109 @@
...
-        caseLoadId: 'LEI',
-        caseloadFunction: 'GENERAL',
-        currentlyActive: true,
-        description: 'Leeds (HMP)',
-        type: 'INST',
-      },
-    ]
-
-    it('Get /api/users/me/caseLoads', async () => {
-      fakeApi.get('/api/users/me/caseLoads').reply(200, stubbedPrisonCaseloadData)
```
timharrison-moj, line 164: And this one takes the logged-in user's own token too.

server/data/prisonApiClient.ts
```diff
@@ -9,63 +9,75 @@ import type {
...
+    )) as Promise<HomeDetentionCurfew[]>
   }
 
-  // Called with the user's own token
-  async getUser(userToken: string): Promise<PrisonApiUserDetail> {
-    return this.restClient.getWithUserToken({
-      userToken,
-      path: '/api/users/me',
-    }) as Promise<PrisonApiUserDetail>
+  async getUser(username: string): Promise<PrisonApiUserDetail> {
```
timharrison-moj, line 66: Should pass through the logged-in user's token?

server/data/prisonApiClient.ts
```diff
@@ -9,63 +9,75 @@ import type {
...
 
-  // Called with the user's own token
-  async getUserCaseloads(userToken: string): Promise<PrisonApiCaseload[]> {
-    return this.restClient.getWithUserToken({
-      userToken,
-      path: '/api/users/me/caseLoads',
-    }) as Promise<PrisonApiCaseload[]>
+  async getUserCaseloads(username: string): Promise<PrisonApiCaseload[]> {
+    return (await this.get(
+      {
```
timharrison-moj, line 77: Should pass through the logged-in user's own token?

server/data/tokenStore.test.ts
```diff
@@ -1,34 +1,66 @@
...
+
+    const token = await tokenStore.getAuthToken('joebloggs')
+
+    expect(token).toEqual('saved token')
   })
 
-  it('Can retrieve token', async () => {
-    redisClient.get.mockImplementation((key, callback) => callback(null, 'token-1'))
+  it('Auth is called to get and save a token for user if one is not saved', async () => {
+    nock(config.apis.hmppsAuth.url, {
```
timharrison-moj, line 36: I see the issue - normal users do not have client-credential accounts - and we don't call out to get them another token. Only system-api clients have these accounts. This is why some client methods (not many - about 3 or 4) accept the user's own token, held res.locals.user, and passed through to the API client. All the rest will get a system token, from redis, or if absent there, from auth and store it.

server/data/tokenStore.ts
```diff
@@ -28,11 +29,43 @@ export default class TokenStore {
...
+    const key = username || '%ANONYMOUS%'
+    const token = await this.getToken(key)
+    if (token) {
+      return token
+    }
+
+    const clientToken = generateOauthClientToken(
+      config.apis.hmppsAuth.systemClientId,
+      config.apis.hmppsAuth.systemClientSecret
+    )
```
timharrison-moj, line 42: This is always getting a system token.

server/middleware/populateCurrentUser.ts
```diff
@@ -26,16 +26,16 @@ export default function populateCurrentUser(userService: UserService): RequestHa
...
           const cvlUser = new CvlUserDetails()
 
           if (authSource === 'nomis') {
             // Assemble user information from Nomis via prison API
             const [prisonUser, prisonUserCaseload] = await Promise.all([
-              userService.getPrisonUser(token),
-              userService.getPrisonUserCaseloads(token),
+              userService.getPrisonUser(username),
+              userService.getPrisonUserCaseloads(username),
             ])
```
timharrison-moj, line 39: This was passing through the logged-in user's own token.

server/data/hmppsAuthClient.test.ts
```diff
@@ -1,108 +1,40 @@
...
-
-      fakeHmppsAuthApi
-        .get('/api/user/me')
-        .matchHeader('authorization', `Bearer ${token.access_token}`)
-        .reply(200, response)
-
-      const output = await hmppsAuthClient.getUser(token.access_token)
-      expect(output).toEqual(response)
-    })
   })
```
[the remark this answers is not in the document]
smcveigh941: Checking the construction of the HTTP request is tested in `hmppsRestClient.test.ts`, these clients now only delegate the request to the rest client by passing the required headers, url, and data. Therefore nock is only used in the rest client test where the actual HTTP request is happening

server/data/hmppsAuthClient.ts
smcveigh941: Now moved to `tokenStore.test.ts` where token is generated based on whether username is supplied

server/services/communityService.ts
```diff
@@ -1,34 +1,28 @@
...
+  constructor(
+    private readonly communityApiClient: CommunityApiClient,
+    private readonly probationSearchApiClient: ProbationSearchApiClient
+  ) {}
 
   async getStaffDetail(deliusUsername: string): Promise<CommunityApiStaffDetails> {
-    // Important: No username is passed into the getSystemClientToken for the community API
-    const token = await this.hmppsAuthClient.getSystemClientToken()
-    return new CommunityApiClient(token).getStaffDetailByUsername(deliusUsername)
+    return this.communityApiClient.getStaffDetailByUsername(deliusUsername)
```
timharrison-moj, line 13: This is the bit which skipped the username parameter when getting a system token for the community api.

server/data/hmppsRestClient.ts
```diff
@@ -38,36 +40,41 @@ interface StreamRequest {
...
+  constructor(private readonly name: string, private readonly apiConfig: ApiConfig) {
+    this.agent = apiConfig.url.startsWith('https') ? new HttpsAgent(apiConfig.agent) : new Agent(apiConfig.agent)
+    this.tokenStore = new TokenStore()
   }
 
-  async get({ path = null, query = '', headers = {}, responseType = '', raw = false }: GetRequest): Promise<unknown> {
-    logger.info(`Get using admin client credentials: calling ${this.name}: ${path} ${query}`)
+  async get(
+    { path = null, query = {}, headers = {}, responseType = '', raw = false }: GetRequest,
+    signedWithMethod?: SignedWithMethod
```
smcveigh941, line 60: Now introduced this option to the rest client. Can either provide:

1. User token from their session `{ token: 'abcde' }`
2. Username from which to generate a system token associated to the user `{ username: 'joebloggs' }`
3. Supply nothing, in which case a system token will be generated (i.e. community API)

server/data/hmppsRestClient.ts
```diff
@@ -38,36 +40,41 @@ interface StreamRequest {
...
+  ): Promise<unknown> {
+    const signedWith = signedWithMethod?.token || (await this.tokenStore.getSystemToken(signedWithMethod?.username))
```
smcveigh941, line 62: Supplied token takes precedence over supplied username

server/data/hmppsRestClient.ts
smcveigh941: Should be addressed now!

server/data/licenceApiClient.ts
```diff
@@ -15,79 +14,109 @@ import type {
...
-      path: `/licence/create`,
-      data: licence,
-    })) as Promise<LicenceSummary>
+  async createLicence(licence: CreateLicenceRequest, user: User): Promise<LicenceSummary> {
+    return (await this.post(
+      {
+        path: `/licence/create`,
+        data: licence,
+      },
+      { username: user.username }
```
smcveigh941, line 30: Example here of an API client choosing to supply the username from the User object rather than the token

server/data/prisonApiClient.test.ts
smcveigh941: Should be addressed now!

server/data/prisonApiClient.ts
```diff
@@ -1,71 +1,90 @@
...
-    return this.restClient.getWithUserToken({
-      userToken,
-      path: '/api/users/me',
-    }) as Promise<PrisonApiUserDetail>
+  async getUser(user: User): Promise<PrisonApiUserDetail> {
+    return (await this.get(
+      {
+        path: '/api/users/me',
+      },
+      { token: user.token }
```
smcveigh941, line 78: And here is an example of an API client choosing to sign the request with the user's own token

server/data/tokenStore.ts
smcveigh941: Yes this method will now always get system tokens. The request will not come in here if a user's token is supplied

server/middleware/populateCurrentUser.ts
smcveigh941: The decision of whether or not to pass user token or system token is delegated to the API layer where it makes sense in my opinion

server/data/hmppsRestClient.ts
```diff
@@ -38,36 +40,41 @@ interface StreamRequest {
   errorLogger?: (e: UnsanitisedError) => void
 }
 
-export default class RestClient {
-  agent: Agent
+interface SignedWithMethod {
+  token?: string
+  username?: string
+}
```
timharrison-moj, line 46: I like this, and how it switches all calls between either the user token, if present, or generates a system token with or without the username.

server/@types/CvlUserDetails.ts
```diff
@@ -36,4 +36,4 @@ class CvlUserDetails {
   emailAddress?: string
 }
 
-export = CvlUserDetails
+export type User = Express.User & CvlUserDetails
```
timharrison-moj, line 39: I wonder about this bit? The `CvlUserDetails` was created to hold the values we pull from Nomis, Auth and Delius based on the user auth source, values useful to hold in their session (e.g. staff ids, teams, ldu, caseload etc). Do we need to pass this structure, enhanced with the express User details to each of the API clients? Would it be clearer to have a new type which only contains the `username` and `user token` for the clear purpose for API clients to decide themselves which value they will use for each call?

timharrison-moj: Sorry to be a pain - I know you've worked hard on this (69 files and counting on this PR!), and its all great stuff which will help - in fact, I think we could share this work for other services to follow as a template. Its really core to the service and the structure of how routes, services, API client and rest clients are wired together, so really important to make this as clear and simple as possible.

timharrison-moj: By the way, please argue the case if you think this IS the right place for it!

server/data/communityApiClient.ts
```diff
@@ -1,26 +1,22 @@
...
+  constructor() {
+    super('Community API', config.apis.communityApi as ApiConfig)
   }
 
-  async getStaffDetailByUsername(deliusUsername: string): Promise<CommunityApiStaffDetails> {
-    logger.info(`communityClient: getStaffDetailByUsername(${deliusUsername}`)
-    return this.restClient.get({
-      path: `/secure/staff/username/${deliusUsername}`,
-    }) as Promise<CommunityApiStaffDetails>
+  async getStaffDetailByUsername(user: User): Promise<CommunityApiStaffDetails> {
```
timharrison-moj, line 11: Might be good to retain a comment here? `// Uses the user's own token`

server/data/hmppsAuthClient.ts
```diff
@@ -24,58 +17,16 @@ export type AuthUserRole = {
...
+  constructor() {
+    super('HMPPS Auth Client', config.apis.hmppsAuth as ApiConfig)
   }
 
-  async getUserRoles(token: string): Promise<string[]> {
-    return this.restClient(token)
-      .get({ path: '/api/user/me/roles' })
-      .then(roles => (<AuthUserRole[]>roles).map(role => role.roleCode)) as Promise<string[]>
+  async getUser(user: User): Promise<AuthUserDetails> {
+    return (await this.get({ path: '/api/user/me' }, { token: user.token })) as Promise<AuthUserDetails>
```
timharrison-moj, line 26: Maybe retain the comment about using user's own token?

server/data/hmppsAuthClient.ts
```diff
@@ -24,58 +17,16 @@ export type AuthUserRole = {
...
-    const token = await this.tokenStore.getToken(key)
-    if (token) {
-      return token
-    }
-    const newToken = await getSystemClientTokenFromHmppsAuth(username)
-    // Set the TTL slightly less than expiry of token
-    await this.tokenStore.setToken(key, newToken.body.access_token, newToken.body.expires_in - 60)
-    return newToken.body.access_token
+  async getUserEmail(user: User): Promise<AuthUserEmail> {
+    return (await this.get({ path: '/api/me/email' }, { token: user.token })) as Promise<AuthUserEmail>
```
timharrison-moj, line 30: As above?

server/data/licenceApiClient.ts
timharrison-moj: Like it. Though, as described above, not sure about overloading the CvlUserDetails class with Express User - could be a separate, much more simple object? Argue it though, as I'm not 100% - there might be other reasons for doing so!

server/data/prisonApiClient.ts
timharrison-moj: Very good. Maybe retain the comment? Not sure - its clearly passing the token instead of the username, so should be ok!

server/data/prisonerSearchApiClient.ts
```diff
@@ -1,26 +1,31 @@
...
-      data: nomisIdsToSearch,
-    })) as Promise<Prisoner[]>
+  async searchPrisonersByNomsIds(nomisIdsToSearch: PrisonerSearchByNomisIds, user?: User): Promise<Prisoner[]> {
+    return (await this.post(
+      {
+        path: '/prisoner-search/prisoner-numbers',
+        data: nomisIdsToSearch,
+      },
+      { username: user?.username }
+    )) as Promise<Prisoner[]>
```
timharrison-moj, line 29: To get here can the the `user?.username` ever be nullable?

server/data/tokenStore.test.ts
```diff
@@ -1,34 +1,66 @@
...
+
+afterEach(done => {
+  redisClient.flushall(done)
+  jest.clearAllMocks()
+  nock.cleanAll()
+})
+
+describe('Token Store', () => {
+  it('Token for user is returned from redis is exists', async () => {
+    redisClient.set('joebloggs', 'saved token')
```
timharrison-moj, line 28: This test is checking a scenario which won't happen? Only the system api client token is cached in Redis here.

server/data/licenceApiClient.ts
timharrison-moj: Lets leave it like it is for now. I can see that the User object is a convenient vehicle to pass around, containing everything that services and clients need.

server/@types/CvlUserDetails.ts
timharrison-moj: Change of mind - lets leave it! Its convenient, available and contains everything services need. A minimal object could be introduced in future, just containing `{ username: xxx, token: xxxx, displayName: xxxx }` which is required by services and API clients.

timharrison-moj (approved):
Lots of very good structural changes here, and all improvements. Nice PR! 
Small comments only, happy with these - but I'm going to check out and run through all the flows locally against DEV to verify.

server/@types/CvlUserDetails.ts
smcveigh941: Yes probably should have a specific type, would include `prisonCaseload` i think too though

server/data/prisonerSearchApiClient.ts
smcveigh941: The Cron job uses this method, which will not pass a username

server/data/tokenStore.test.ts
smcveigh941: This scenario will happen, its testing that the system token is returned from cache instead of generated again

PR #143, opened 2021-12-10
CVSL-67 Access controls in 3 layers

timharrison-moj:
Implements layered access control methods for CVL journeys :-

3 levels of checking...

1. Checks the user roles within the journeys (RO for create routes, DM for approval routes, RO or CA for view/print routes) - access-denied if no appropriate role is present. This is in `roleCheckMiddleware`.
2. Checks the user's auth source - if prison their caseload must includes the licensee's prison code. If probation their LDUs must include the licensee's managing LDU. This is part of `fetchLicenceMiddleware`
3. Checks the URL path requested against the fetched licence status. Only a small list of RegExp URL matches are required to check that the access is allowed for a licence in this status. This is also part of `fetchLicenceMiddleware`.

Unit tests for all changes. 
Slight alteration to integration tests to include the probation LDUs against teams in the Delius staff details mocked endpoint (similar to the Delius wire mock change).


server/middleware/fetchLicenceMiddleware.test.ts
```diff
@@ -5,39 +5,109 @@ import { Licence } from '../@types/licenceApiClientTypes'
...
 const next = jest.fn()
+
+const prisonRes = {
+  locals: {
+    user: {
+      username: 'XX',
+      nomisStaffId: 123,
+      prisonCaseload: ['LEI', 'MDI'],
+    },
+    licence: undefined,
```
smcveigh941, line 21: Dont need it here if its undefined?

server/middleware/fetchLicenceMiddleware.test.ts
```diff
@@ -5,39 +5,109 @@ import { Licence } from '../@types/licenceApiClientTypes'
...
+    await middleware(req, res, next)
+    expect(res.locals.licence).toBeUndefined()
+    expect(licenceService.getLicence).toBeCalledTimes(1)
+    expect(res.redirect).toBeCalledWith('/access-denied')
+    expect(next).not.toBeCalled()
+  })
+
+  it('should allow access for a probation user based on LDU', async () => {
+    res = probationRes
+    res.locals.licence = undefined
```
smcveigh941, line 96: It would already be undefined here right?

server/middleware/fetchLicenceMiddleware.ts
```diff
@@ -1,16 +1,43 @@
...
+import logger from '../../logger'
 
 export default function fetchLicence(licenceService: LicenceService): RequestHandler {
   return async (req, res, next) => {
     if (req.params.licenceId) {
       const { user } = res.locals
       const licence = await licenceService.getLicence(req.params.licenceId, user)
 
+      // Does this prison user have a caseload which allows access this licence?
+      if (licence && user?.nomisStaffId) {
```
smcveigh941, line 13: Can user be undefined? That would suggest they are not logged in?

server/middleware/fetchLicenceMiddleware.ts
```diff
@@ -1,16 +1,43 @@
...
+        if (!user.prisonCaseload.includes(licence.prisonCode)) {
+          logger.info(
+            `Prison caseload ${user?.prisonCasload} denied access to licence ID ${licence.id} ${licence.prisonCode}`
+          )
+          return res.redirect('/access-denied')
```
smcveigh941, line 18: Do we want to redirect or just render a different view? Not sure, redirect is probably ok

smcveigh941: Yea, same pattern as `/authError`, fine

server/routes/approvingLicences/index.ts
```diff
@@ -15,10 +16,21 @@ export default function Index({ licenceService }: Services): Router {
   const routePrefix = (path: string) => `/licence/approve${path}`
 
   const get = (path: string, handler: RequestHandler) =>
-    router.get(routePrefix(path), fetchLicence(licenceService), asyncMiddleware(handler))
+    router.get(
+      routePrefix(path),
+      roleCheckMiddleware(['ROLE_LICENCE_DM']),
```
smcveigh941, line 21: What I had in mind too 👍

server/utils/urlAccessByStatus.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const allowedPaths = [
+  {
+    status: 'IN_PROGRESS',
+    allowed: ['/licence/create/.*', '/licence/view/.*'],
+  },
+  {
+    status: 'SUBMITTED',
+    allowed: [
+      '/licence/create/id/(\\d)*/check-your-answers.*',
```
smcveigh941, line 11: As the service design progresses, I wonder if this could catch us out, may be hard to think about every status when adding pages and journeys. Not sure what other ways there are of doing it though..unless just a check on every handler but this feels nice and abstract

server/views/pages/accessDenied.njk
```diff
@@ -0,0 +1,33 @@
...
+
+{% block content %}
+    <div class="govuk-grid-row govuk-body">
+        <div class="govuk-grid-column-two-thirds">
+            {{ govukWarningText({
+                text: "Access is denied",
+                iconFallbackText: "Warning"
+            }) }}
+
+            <h2 class="govuk-heading-s">Why am I seeing this page?</h2>
```
smcveigh941, line 17: I think Jo has been thinking about content for error pages so this might have discussion but fine for now

server/views/pages/accessDenied.njk
```diff
@@ -0,0 +1,33 @@
...
+            <p>
+                You have attempted to access an area of the service which your permissions do not allow.
+                This may be because the person or licence accessed is no longer within your prison or probation case view.
+            </p>
+
+            <p class="govuk-body">
+                {{ govukButton({
+                    text: "Home page",
+                    classes: "govuk-button--primary govuk-!-margin-bottom-0 govuk-body",
```
smcveigh941, line 26: Dont think `govuk-body` belongs on a button

server/middleware/fetchLicenceMiddleware.test.ts
timharrison-moj: The tests interfere, as res gets changed, so needed here when using probationRes.

server/views/pages/accessDenied.njk
```diff
@@ -0,0 +1,33 @@
...
+                iconFallbackText: "Warning"
+            }) }}
+
+            <h2 class="govuk-heading-s">Why am I seeing this page?</h2>
+            <p>
+                You have attempted to access an area of the service which your permissions do not allow.
+                This may be because the person or licence accessed is no longer within your prison or probation case view.
+            </p>
+
+            <p class="govuk-body">
```
smcveigh941, line 23: Probably dont need the button wrapped in a paragraph tag

server/middleware/fetchLicenceMiddleware.ts
timharrison-moj: Could be a probation user, in which case would be undefined.

server/views/pages/accessDenied.njk
```diff
@@ -0,0 +1,33 @@
+{% extends "layout.njk" %}
+
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% set pageTitle = applicationName + " - access denied" %}
+{% set pageId = "access-denied-page" %}
+
+{% block content %}
+    <div class="govuk-grid-row govuk-body">
```
smcveigh941, line 10: The `govuk-body` class should be used on each `<p>` tag, not to worry about these html comments though, they don't make any difference

server/utils/urlAccessByStatus.ts
timharrison-moj: Its generally just checking the base path of each journey, except check-your-answers and edit, both of which need to be accessed when the licence is in other statuses. I think its pretty clean and follows a pattern. See how it goes?

smcveigh941 (approved):

server/views/pages/accessDenied.njk
timharrison-moj: yes I agree. Its a placeholder, until she sees it.

server/views/pages/accessDenied.njk
timharrison-moj: Wonder how that got there? I'll remove.

server/middleware/fetchLicenceMiddleware.test.ts
timharrison-moj: I think its needed, but can check again.

timharrison-moj: Tests alter the object by populating a licence in it.

server/views/pages/accessDenied.njk
timharrison-moj: I'll make a note of these things to check and sort as a follow on PR along with access denied page content.

PR #147, opened 2021-12-13
CVSL-266: Service layer test coverage

smcveigh941:
Details of coverage changes here https://app.codecov.io/gh/ministryofjustice/create-and-vary-a-licence/compare/main...cvsl-266-service-layer-test-coverage/changes

* 3 services now with 100% coverage
* Coverage of licenceService still to come in another PR

### DIFF
--- a/server/services/prisonerService.ts
+++ b/server/services/prisonerService.ts
@@ -27,8 +27,8 @@ export default class PrisonerService {
       base64String = image.toString('base64')
     } catch (error) {
       logger.info(`No image data found for ${nomsId} - sending placeholder image`)
-      const content = await fs.readFileSync('assets/images/image-missing.png')
-      base64String = Buffer.from(content).toString('base64')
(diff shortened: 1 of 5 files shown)


server/services/prisonerService.ts
```diff
@@ -27,8 +27,8 @@ export default class PrisonerService {
       base64String = image.toString('base64')
     } catch (error) {
       logger.info(`No image data found for ${nomsId} - sending placeholder image`)
-      const content = await fs.readFileSync('assets/images/image-missing.png')
-      base64String = Buffer.from(content).toString('base64')
+      const content = fs.readFileSync('assets/images/image-missing.png')
+      base64String = content.toString('base64')
```
smcveigh941, line 31: Type Buffer is already returned from `fs.readFileSync`. And since the method is synchronous it does not need to be awaited

PR #149, opened 2021-12-13
CVSL-281: Fixed missing URLs in allowed list.

timharrison-moj:
Approval flow sets the status to APPROVED and then shows the confirmation - needed confirmation URL in allow list for APPROVED.
Approval flow sets the status to REJECTED and then shows a rejected confirmation - needed confirmation URL in allow list for REJECTED.

### DIFF
--- a/server/utils/urlAccessByStatus.ts
+++ b/server/utils/urlAccessByStatus.ts
@@ -20,6 +20,7 @@ const allowedPaths = [
     allowed: [
       '/licence/create/id/(\\d)*/check-your-answers.*',
       '/licence/create/id/(\\d)*/edit.*',
+      '/licence/approve/id/(\\d)*/confirm-approved.*',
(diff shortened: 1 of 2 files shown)


server/utils/urlAccessByStatus.ts
```diff
@@ -28,6 +29,7 @@ const allowedPaths = [
     allowed: [
       '/licence/create/id/(\\d)*/check-your-answers.*',
       '/licence/create/id/(\\d)*/edit.*',
+      '/licence/approve/id/(\\d)*/confirm-rejected.*',
```
smcveigh941, line 32: Integration tests will not have caught this if wiremock has not been told to return REJECTED status on the confirmation page ... sometimes I wonder if we need regression tests which run on the full stack ... an overnight selenium job or something ?

smcveigh941 (approved):

PR #160, opened 2022-01-02
CVSL-255 Allow upload of the MapMaker PDF file for the exclusion zone condition

timharrison-moj:
Quite a big PR to enable the following:
* Adds a file upload input type to form builder for the exclusion zone condition, if selected.
* Processes a `multipart/form-data` POST to upload a file
* Uses the Node `multer` middleware to accept uploaded files
* Validation to check for a file and a description being input.
* Sends the file to the API as a `multipart/form-data` POST if validation passed.
* Adds a new structure - `uploadSummary` - to additional conditions retrieved by the fetch licence operation. This contains the details, including a thumbnail image, of any uploads relating to this condition.
* Only 1 condition has an upload, and only 1 upload is allowed for that condition. The upload can be removed, and re-uploaded.
* Shows the thumbnail image on all view licence forms and the add/change condition forms.
* Shows the full image and description as an Appendix when printing the licence.


integration_tests/pages/appointmentTime.ts
```diff
@@ -23,7 +23,7 @@ export default class AppointmentTimePage extends Page {
 
   enterDate = (moment: Moment): AppointmentTimePage => {
     cy.get(this.inductionDateDayId).type(moment.date().toString())
-    cy.get(this.inductionDateMonthId).type(moment.month().toString())
+    cy.get(this.inductionDateMonthId).type((moment.month() + 1).toString())
     cy.get(this.inductionDateYearId).type(moment.year().toString())
```
timharrison-moj, line 27: For January, this was entering 0 for the month and failing validation. Months in moment are 0-11, so added 1.

server/routes/viewingLicences/handlers/printLicence.ts
```diff
@@ -35,22 +46,30 @@ export default class PrintLicenceRoutes {
...
+    const { licence, user } = res.locals
     const { qrCodesEnabled } = res.locals
     const { licencesUrl, pdfOptions, watermark } = config.apis.gotenberg
-
     const qrCode = qrCodesEnabled ? await this.qrCodeService.getQrCode(licence) : null
     const additionalLicenceConditions = expandAdditionalConditions(licence.additionalLicenceConditions)
     const additionalPssConditions = expandAdditionalConditions(licence.additionalPssConditions)
-    const imageData = await this.prisonerService.getPrisonerImageData(username, licence.nomsId)
+    const imageData = await this.prisonerService.getPrisonerImageData(licence.nomsId, user)
     const filename = licence.nomsId ? `${licence.nomsId}.pdf` : `${licence.lastName}.pdf`
```
timharrison-moj, line 64: This was a bug left over from previous PRs - just correcting it here so it finds the prisoner image.

server/routes/viewingLicences/handlers/printLicence.ts
```diff
@@ -86,4 +107,24 @@ export default class PrintLicenceRoutes {
...
+  getExclusionZoneDescription = (additionalConditions: AdditionalCondition[]): string => {
+    let description = ''
+    additionalConditions.forEach(condition => {
+      if (condition?.uploadSummary?.length > 0) {
+        description = condition.uploadSummary[0]?.description
+      }
+    })
+    return description
+  }
 }
```
timharrison-moj, line 130: These both make assumptions that there is only one upload per condition, and at present, one upload for all conditions. Might need a rethink if we identify other uploads.

server/config.ts
```diff
@@ -70,8 +70,8 @@ export default {
     licenceApi: {
       url: get('LICENCE_API_URL', 'http://localhost:8089', requiredInProduction),
       timeout: {
-        response: get('LICENCE_API_TIMEOUT_RESPONSE', 10000),
-        deadline: get('LICENCE_API_TIMEOUT_DEADLINE', 10000),
+        response: get('LICENCE_API_TIMEOUT_RESPONSE', 30000),
```
smcveigh941, line 73: Is this because the image endpoint can be slow?

timharrison-moj: Yes - on the first call to pdfDoc it needs to load fonts and with a 10 sec timeout this did time-out once when running locally. However, subsequent calls are very quick. I only saw this once so not sure that 30s is the right value.

server/data/hmppsRestClient.test.ts
```diff
@@ -331,4 +332,31 @@ describe('Hmpps Rest Client tests', () => {
       expect(nock.isDone()).toBe(true)
     })
   })
+
+  describe('POST MULTIPART', () => {
+    it('Should accept a file upload', async () => {
+      fs.writeFileSync('./test-file.txt', 'a test file')
```
smcveigh941, line 338: This writes a file to the disk? Maybe add it to `.gitignore`?

server/middleware/validationMiddleware.ts
```diff
@@ -14,6 +14,11 @@ function validationMiddleware(type?: new () => object): RequestHandler {
 
     const classType = (getAdditionalConditionByCode(req.body.code)?.type as ClassConstructor<object>) || type
 
+    // Must add the uploaded filename to the req.body for file upload types to validate that a file has been selected
+    if (req.file && classType?.name === 'OutOfBoundsRegion') {
+      req.body = { ...req.body, outOfBoundFilename: req.file.originalname }
+    }
+
     const validationScope = plainToClass(classType, { ...req.body, licence }, { excludeExtraneousValues: false })
```
smcveigh941, line 22: Instead of explicitly harcoding `OutOfBoundsRegion`, you could add `req.file` to the validation scope here? and add a validator annotation to `outOfBoundsFilename`?

server/routes/creatingLicences/handlers/additionalLicenceConditionInput.ts
```diff
@@ -34,6 +36,20 @@ export default class AdditionalLicenceConditionInputRoutes {
 
     await this.licenceService.updateAdditionalConditionData(licenceId, conditionId, req.body, user)
 
+    if (req.file) {
+      // Check if there was a file upload in this POST request - only one allowed - the exclusion zone(outOfBoundFilename)
+      if (req.file.fieldname === 'outOfBoundFilename') {
```
smcveigh941, line 41: If the `req.file` is passed through the class-transformer correctly in the validation middleware, I dont think this check is needed, since the class-transformer will throw away anything which isn't defined in the class

server/routes/creatingLicences/types/additionalConditionInputs/outOfBoundsRegion.ts
```diff
@@ -3,8 +3,12 @@ import { IsNotEmpty } from 'class-validator'
 
 class OutOfBoundsRegion {
   @Expose()
-  @IsNotEmpty({ message: 'Enter the name of the area on the map' })
+  @IsNotEmpty({ message: 'Enter the name of the area shown in the attached map' })
   outOfBoundArea: string
+
+  @Expose()
+  @IsNotEmpty({ message: 'You must attach an exclusion zone MapMaker PDF file' })
+  outOfBoundFilename: string
```
smcveigh941, line 11: Could annotate this with a custom validator which checks that `req.file` contains the correct things which match this field

server/routes/creatingLicences/handlers/additionalLicenceConditionInput.ts
timharrison-moj: I don't like this bit either and open to suggestions. Separate validator sounds good, so will take a look.

server/views/pages/create/additionalLicenceConditionInput.njk
```diff
@@ -15,9 +16,24 @@
...
+       {% if input.type === 'fileUpload' %}
+           {% set fileUpload = true %}
+       {% endif %}
+    {% endfor %}
+
+    {# Alter the form encoding to multi-part formdata and supply csrf token in query params for file upload conditions #}
+    <form method="POST"
+        {% if fileUpload %}
+            encType="multipart/form-data"
+            action="?_csrf={{ csrfToken }}{{ fromReviewAppend }}"
```
smcveigh941, line 31: Can be the following? Then you dont need the extra `fromReviewAppend`

`action="{{fromReview}}&_csrf={{csrfToken}}"`

timharrison-moj: Thanks. Will take a look now.

server/routes/creatingLicences/types/additionalConditionInputs/outOfBoundsRegion.ts
timharrison-moj: Good suggestion - looking now.

server/services/licenceService.ts
```diff
@@ -198,6 +201,39 @@ export default class LicenceService {
     return this.licenceApiClient.updateAdditionalConditionData(licenceId, additionalConditionId, requestBody, user)
   }
 
+  async uploadExclusionZoneFile(
+    licenceId: string,
+    additionalConditionId: string,
+    fileToUpload: Express.Multer.File,
+    user: User,
+    testMode = false
```
smcveigh941, line 209: Whats this for?

server/services/licenceService.ts
```diff
@@ -198,6 +201,39 @@ export default class LicenceService {
...
+  ): Promise<void> {
+    logger.info(`Called uploadExclusionZoneFile - name ${fileToUpload.originalname} path ${fileToUpload.path}`)
+    await this.licenceApiClient.uploadExclusionZoneFile(licenceId, additionalConditionId, user, fileToUpload)
+    if (!testMode) {
+      logger.info(`Removing file ${fileToUpload.path} after uploading content to API`)
+      fs.unlinkSync(fileToUpload.path)
```
smcveigh941, line 215: Does the file uploaded by the user get stored on the node file system server side?

server/services/licenceService.ts
```diff
@@ -198,6 +201,39 @@ export default class LicenceService {
...
+    }
+  }
+
+  async removeExclusionZoneFile(licenceId: string, conditionId: string, user: User): Promise<void> {
+    logger.info(`removeExclusionZoneFile - licenceId ${licenceId}, conditionId  ${conditionId}`)
+    return this.licenceApiClient.removeExclusionZoneFile(licenceId, conditionId, user)
+  }
+
+  // Get the streamed image data for rendering in HTML templates
+  async getExclusionZoneImage(licenceId: string, conditionId: string, user: User): Promise<Readable> {
```
smcveigh941, line 225: Unit test?

server/data/hmppsRestClient.ts
```diff
@@ -168,4 +175,32 @@ export default class HmppsRestClient {
...
+      })
+      .auth(signedWith, { type: 'bearer' })
+      .set(headers)
+      .responseType(responseType)
+      .timeout(this.apiConfig.timeout)
+      .then(response => {
+        return response.body
+      })
+      .catch(error => {
+        const sanitisedError = sanitiseError(error)
```
smcveigh941, line 201: Unit test for the error handling?

server/services/licenceService.ts
timharrison-moj: In some service tests, it doesn't actually create the file as the `licenceApiClient` is mocked. The test mode prevents it trying to remove this file.

server/services/licenceService.ts
timharrison-moj: It is temporarily stored on the node file system and removed after uploading to the API. I would prefer not to do this, but struggled for alternatives.

server/services/licenceService.ts
timharrison-moj: it('Get the exclusion zone map image as base64 JPEG', async () ? This one? Ah, you mean the other implementation which streams the data? Its not actually used anywhere as yet. I can add a test for it.

server/data/hmppsRestClient.ts
timharrison-moj: I'll take a look.

timharrison-moj: I've added two new tests here. One for when no file is supplied and one for not found on the endpoint.

server/services/licenceService.ts
timharrison-moj: Added a test for this too, even though unused currently.

server/views/pages/create/additionalLicenceConditionInput.njk
timharrison-moj: No, it needs the `?` prior to the query parameters. The `_csrf` param is always needed in the form action for file uploads, but the fromReview is optional.

server/routes/creatingLicences/handlers/additionalLicenceConditionInput.ts
timharrison-moj: A separate validator is now used, which checks for a file upload and validates the field name, file name, mime type and size > 0 bytes.

server/routes/creatingLicences/types/additionalConditionInputs/outOfBoundsRegion.ts
timharrison-moj: Custom validator is now used.

server/validators/isValidExclusionZoneFile.ts
```diff
@@ -0,0 +1,30 @@
+import { registerDecorator, ValidationArguments, ValidationOptions } from 'class-validator'
```
smcveigh941, line 1: For the other validators we have a separate test file which unit tests them in isolation

smcveigh941 (approved):
LGTM

server/validators/isValidExclusionZoneFile.ts
timharrison-moj: Thanks, I saw that late on, and I've added this requirement to the next ticket, CVSL-304.

PR #161, opened 2022-01-05
CVSL-276: Added event listeners which consume prison and probation events from AWS SQS

smcveigh941:


server/config.ts
```diff
@@ -6,7 +6,7 @@ function get<T>(name: string, fallback: T, options = { requireInProduction: fals
   if (process.env[name]) {
     return process.env[name]
   }
-  if (fallback !== undefined && (!production || !options.requireInProduction)) {
+  if (fallback !== undefined && !(production && options.requireInProduction)) {
```
smcveigh941, line 9: DeMorgan's law makes it easier to understand, at least for me

integration_tests/pages/appointmentTime.ts
```diff
@@ -23,7 +23,7 @@ export default class AppointmentTimePage extends Page {
 
   enterDate = (moment: Moment): AppointmentTimePage => {
     cy.get(this.inductionDateDayId).type(moment.date().toString())
-    cy.get(this.inductionDateMonthId).type(moment.month().toString())
+    cy.get(this.inductionDateMonthId).type((moment.month() + 1).toString())
```
smcveigh941, line 26: Broken because of the new year, January is entered as 0

server/config.ts
timharrison-moj: oh my. Don't like either much but yours is definitely better.

docker-compose-dev.yml
```diff
@@ -40,5 +40,26 @@ services:
     healthcheck:
       test: [ 'CMD', 'curl', '-f', 'http://localhost:3000/ping' ]
 
+  localstack:
+    image: localstack/localstack:0.11.2
+    networks:
```
timharrison-moj, line 45: I think there are later images now?

helm_deploy/create-and-vary-a-licence/values.yaml
```diff
@@ -45,6 +45,14 @@ generic-service:
...
       REDIS_HOST: "primary_endpoint_address"
       REDIS_AUTH_TOKEN: "auth_token"
+    create-and-vary-a-licence-prison-events-sqs-instance-output:
+      SQS_PRISON_EVENTS_ACCESS_KEY_ID: "access_key_id"
+      SQS_PRISON_EVENTS_SECRET_ACCESS_KEY: "secret_access_key"
+      SQS_PRISON_EVENTS_QUEUE_URL: "sqs_id"
+    create-and-vary-a-licence-probation-events-sqs-instance-output:
+      SQS_PROBATION_EVENTS_ACCESS_KEY_ID: "access_key_id"
+      SQS_PROBATION_EVENTS_SECRET_ACCESS_KEY: "secret_access_key"
+      SQS_PROBATION_EVENTS_QUEUE_URL: "sqs_id"
```
timharrison-moj, line 55: Think we also need the DLQ creds too?

integration_tests/pages/appointmentTime.ts
timharrison-moj: Ah, I also fixed in my branch too.

server/config.ts
```diff
@@ -44,6 +44,26 @@ export default {
     secret: get('SESSION_SECRET', 'app-insecure-default-session', requiredInProduction),
     expiryMinutes: Number(get('WEB_SESSION_TIMEOUT_IN_MINUTES', 120)),
   },
+  sqs: {
+    prisonEvents: {
+      accessKeyId: get('SQS_PRISON_EVENTS_ACCESS_KEY_ID', '', requiredInProduction),
+      secretAccessKey: get('SQS_PRISON_EVENTS_SECRET_ACCESS_KEY', '', requiredInProduction),
+      queueUrl: get(
```
timharrison-moj, line 51: We could default these to the same values used for localstack - just any value, like foo and bar? Or set in .env.

helm_deploy/create-and-vary-a-licence/values.yaml
smcveigh941: Well we don't actually use the dead letter queue. If we have a Cron job which clears it then yea we'll need them

server/config.ts
smcveigh941: Access key and secret are not needed for localstack, but can stick foo and bar in. No need for anything though

timharrison-moj (approved):
LGTM. Couple of minor questions, but nothing to hold it up.

helm_deploy/create-and-vary-a-licence/values.yaml
timharrison-moj: That's fine then, can put them in as and when needed.

PR #162, opened 2022-01-06
CVSL-304: Create temp upload directory and set permissions during image build

timharrison-moj:
This is to fix the failing deployment, which complains it cannot create the /app/uploads directory.


smcveigh941 (approved):
Lgtm

PR #163, opened 2022-01-07
CVSL-285: Link button focus. IE SCSS updates.

timharrison-moj:
Three things here:
* Set the focus property on the link button (names on case list view)
* Updated the SCCS modules included in the IE version of application.css - missing modules.
* Experimental - add an empty main div with a tabindex -1 - attempt to get IE11 to tab into the content for keyboard-only navigation.

### DIFF
--- a/assets/sass/application-ie8.sass
+++ b/assets/sass/application-ie8.sass
@@ -4,8 +4,15 @@ $path: "/assets/images/"
 @import 'govuk/all-ie8'
 @import 'moj/all'
 
-@import './components/header-bar'
-@import './components/card'
+@import "./constants/colours"
+
 @import './components/add-another'
+@import './components/card'
+@import './components/caseload'
 @import './components/check-answers-header'
+@import './components/file_upload'
+@import './components/header-bar'
+@import './components/link-button'
+@import './components/status-tags'
+@import './components/utils'
 @import './local'
--- a/server/views/layout.njk
+++ b/server/views/layout.njk
@@ -75,6 +75,9 @@
   {% include 'partials/licenceDetailsBanner.njk' %}
   {% include 'partials/formError.njk' %}
   <span class="govuk-visually-hidden" id="{{ pageId }}"></span>
+
+  {# IE issue where keyboard navigation does not tab into main content #}
+  <div id="main" tabindex="-1"></div>
 {% endblock %}
 
 {% block bodyEnd %}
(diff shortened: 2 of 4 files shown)


assets/sass/application-ie8.sass
```diff
@@ -4,8 +4,15 @@ $path: "/assets/images/"
 @import 'govuk/all-ie8'
 @import 'moj/all'
 
-@import './components/header-bar'
-@import './components/card'
+@import "./constants/colours"
+
 @import './components/add-another'
+@import './components/card'
```
smcveigh941, line 10: I doubt that this will fix the IE backlink issue. `js-backlink` class is does not have any CSS, its just used as an identifier in JS

smcveigh941: I would maybe suggest trying this in `govukFrontendInit.js`? looks like `Jquery.click()` is deprecated

```
// Initiate the back links
$('[class$=js-backlink]').on('click', e => {
  e.preventDefault()
  if ($('ul.govuk-error-summary__list').length > 0) {
    window.history.go(-2)
  } else {
    window.history.go(-1)
  }
})
```

server/views/layout.njk
```diff
@@ -75,6 +75,9 @@
   {% include 'partials/licenceDetailsBanner.njk' %}
   {% include 'partials/formError.njk' %}
   <span class="govuk-visually-hidden" id="{{ pageId }}"></span>
+
+  {# IE issue where keyboard navigation does not tab into main content #}
+  <div id="main" tabindex="-1"></div>
```
smcveigh941, line 80: Not familiar with this issue

assets/sass/application-ie8.sass
timharrison-moj: Yeah I did look at this JS as well, though this particular change was just to update the ie8 module to include all the same SCCS modules as we include in application.sccs. I'll alter the JS above to see if that deprecated click() method is the issue for IE.

server/views/layout.njk
timharrison-moj: Its a funny one where IE does not focus the main window without a mouse-click into it.  This is one suggested solution, and the other is an empty link `<a id="main" name="main" tabindex="-1">` - neither actually do anything, but apparently make some versions of IE allow tab into the main content. Experimental, as marked! Trouble is I can't test this locally :-(

smcveigh941 (approved):

PR #165, opened 2022-01-07
CVSL-304 exclusion zone tests and styling tweaks

timharrison-moj:
Contains:
* Updates for view a licences /check your answers to use full-width column for displaying licence details.
* Updates for word-wrap on exclusion zone text
* New tests for the validators on the exclusion zone 'OutOfBoundRegion' class.


smcveigh941 (approved):
LGTM

PR #166, opened 2022-01-10
CVSL-305: Add basic user details into the dataLayer on page loads when user is present in res.locals.

timharrison-moj:
This is to expose some details about the user to Google Analytics.
Kam can configure GTM to look for these values and provide some analytics about who is using the service.


smcveigh941 (approved):
LGTM

PR #168, opened 2022-01-12
CVSL-277: Handle released prison event from domain event subscription

smcveigh941:
- Refactored event listeners to comsume service layer as dependency
- Added a new domain event listener
- Handle the `prison-offender-events.prisoner.released` event to set APPROVED licences to ACTIVE
- Integration test to come in a separate PR


server/services/licenceService.ts
```diff
@@ -240,11 +240,11 @@ export default class LicenceService {
...
   }
 
-  async updateStatus(id: string, newStatus: LicenceStatus, user: User): Promise<void> {
+  async updateStatus(id: string, newStatus: LicenceStatus, user?: User): Promise<void> {
     const requestBody = {
       status: newStatus,
-      username: user.username,
-      fullName: user.displayName,
+      username: user?.username || 'SYSTEM',
+      fullName: user?.displayName || 'SYSTEM',
```
smcveigh941, line 247: These are used for auditing, who updated the status? In this instance, it's the system who updates the status.

server/data/licenceApiClient.ts
```diff
@@ -109,8 +109,8 @@ export default class LicenceApiClient extends RestClient {
     )
   }
 
-  async updateLicenceStatus(licenceId: string, statusRequest: StatusUpdateRequest, user: User): Promise<void> {
-    await this.put({ path: `/licence/id/${licenceId}/status`, data: statusRequest }, { username: user.username })
+  async updateLicenceStatus(licenceId: string, statusRequest: StatusUpdateRequest, user?: User): Promise<void> {
+    await this.put({ path: `/licence/id/${licenceId}/status`, data: statusRequest }, { username: user?.username })
   }
```
timharrison-moj, line 114: Is this so it can be called from a user-less session i.e. the queue listener?

smcveigh941: Yes, the actions taken by the system will not have a user associated, so the system token is used

server/listeners/eventHandlers/probationEvents/index.ts
```diff
@@ -1,12 +1,13 @@
...
 import logger from '../../../../logger'
 
-export default async function handleProbationEvents(messages: SQSMessage[]) {
-  messages.forEach(message => {
-    const deliusEvent = message.Body
-    logger.info(`Delius Event : ${deliusEvent}`)
+export default function buildEventHandler() {
+  return async (messages: SQSMessage[]) => {
+    messages.forEach(message => {
+      logger.info(`Delius Event : ${message.Body}`)
```
timharrison-moj, line 8: Changes in preparation to accept required services?

smcveigh941: Yep, refactored it to be the same as the domain events one, so it can accept services as dependency

timharrison-moj (approved):
LGTM. Nice prep for using services in all eventhandlers.

PR #169, opened 2022-01-12
CVSL-305: Fix for pushing user details to GTM data layer

timharrison-moj:
Undefined reference to user object. 
My fault - values were not wrapped in nunjucks curly braces.


yasinmustafa (approved):
LGTM

PR #170, opened 2022-01-12
CVSL-305 second attempt at populating a data layer object for users

timharrison-moj:
It was slightly more involved to populate the user object in the dataLayer, as needed to convert lists to string representations and quote the strings correctly.


server/utils/nunjucksSetup.ts
```diff
@@ -145,6 +145,18 @@ export function registerNunjucks(app?: express.Express): Environment {
     return address ? address.split(', ').filter(line => line.trim().length > 0) : undefined
   })
 
+  njkEnv.addFilter('formatListAsString', (list?: string[]) => {
+    if (list && list.length > 0) {
+      let val = '['
+      list.forEach(item => {
+        val = val.concat(`'${item}',`)
```
smcveigh941, line 152: return `'[${list.join(,)}]'`

I think this would do all of this

timharrison-moj: I'll give it go, thanks. I did think it was a bit convoluted!

timharrison-moj: Close, but not quite! Not quoting the items within the list i.e. [ 'ROLE_XXX', 'ROLE_YYY' ]. Hence why I think it needs to be a bit more convoluted.

smcveigh941 (approved):
LGTM

PR #171, opened 2022-01-12
CVSL-317: Add integration tests for AWS event consumer

smcveigh941:
- Created custom CircleCi executor including localstack (pathfinder and manage intelligence have done this, should add it to the orb for future and other projects?)
- Created queues in localstack for integration test
- Added new integration test helper to push events onto the stack
- Added new helper to check wiremock `count` endpoint to assert the correct endpoint is called after consuming an event
- Handled rejected promise from the event handler

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -136,6 +136,8 @@ jobs:
           name: Create localstack queues
           command: |
             export AWS_DEFAULT_REGION=eu-west-2
+            export AWS_ACCESS_KEY_ID=foo
+            export AWS_SECRET_ACCESS_KEY=bar
 
             aws --endpoint-url=http://localhost:4576 sqs create-queue --queue-name create_and_vary_a_licence_prison_events_queue
             aws --endpoint-url=http://localhost:4576 sqs create-queue --queue-name create_and_vary_a_licence_prison_events_queue_dl
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,19 +133,19 @@ jobs:
           command: sudo apt-get install libxss1
       - hmpps/wait_till_ready # Wait for localstack to start before creating resources
       - run:
-        name: Create localstack queues
-        command: |
-          aws --endpoint-url=http://localhost:4576 sqs create-queue --queue-name create_and_vary_a_licence_prison_events_queue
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -18,6 +18,30 @@ parameters:
...
+      resource_class:
+        default: medium
+        type: string
+      node_tag:
+        default: "14.16-browsers"
+        type: string
+      redis_tag:
+        default: "buster"
+        type: string
+
```
timharrison-moj, line 44: Does seem like an orb candidate, as prisoner-offender-search, probation-offender-search, pathfinder, IMS and hmpps-communication-monitoring will all do this. I would have raised to dps-tech-team, but they've all moved into other teams now, so the ORB is "group-maintained". Ok for now, I'll raise in appropriate channel.

integration_tests/integration/eventHandlers.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+      cy.task('stubUpdateLicenceStatus')
+
+      cy.task(
+        'sendDomainEvent',
+        `{
+            "Message" :  "{\\"eventType\\":\\"prison-offender-events.prisoner.released\\",\\"additionalInformation\\":{\\"nomsNumber\\":\\"A7774DY\\",\\"reason\\":\\"RELEASED\\",\\"details\\":\\"Movement reason code CR\\",\\"currentLocation\\":\\"OUTSIDE_PRISON\\",\\"prisonId\\":\\"MDI\\",\\"currentPrisonStatus\\":\\"NOT_UNDER_PRISON_CARE\\"},\\"version\\":1,\\"occurredAt\\":\\"2022-01-12T14:56:51.662128Z\\",\\"publishedAt\\":\\"2022-01-12T14:58:25.021008001Z\\",\\"description\\":\\"A prisoner has been released from prison\\"}"
+         }`
+      )
+
+      cy.task('verifyEndpointCalled', { verb: 'PUT', path: '/licence/id/1/status', times: 1 }).then(success => {
```
timharrison-moj, line 20: Any value in verifying that the GET endpoint is also called?

timharrison-moj (approved):
LGTM. Couple of questions only.

integration_tests/integration/eventHandlers.spec.ts
smcveigh941: I dont think so, this verifies that the event is consumed. The actual handler is unit tested

PR #175, opened 2022-01-15
Wait for 10 seconds for localstack to be ready before creating resources

smcveigh941:
- Pathfinder uses the `hmpps/wait_till_ready` command, but this will wait for ports to start accepting requests. This is not necessarily the same time that localstack is ready to start creating resources.
- Manage intelligence project simply waits 10 seconds.
- If this still experiences problems, we can do a more sophisticated check by logging out the localstack logs and grepping for "READY"

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,7 +131,9 @@ jobs:
       - run:
           name: Install missing OS dependency
           command: sudo apt-get install libxss1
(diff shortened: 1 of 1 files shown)


timharrison-moj (approved):
LGTM.  I also think there's a way to add queue creation into a file that is automatically run after localstack starts, but lets see how we get on with this one.

PR #179, opened 2022-01-21
CVSL-308: Team caseload view

smcveigh941:
- Refactored `caseloadService` to reuse the caseload filtering rules
- Cacheing can be introduced later if we feel we need it
- Increased test coverage of `caseloadService`
- Added new tab to caseload page
- Integrated with new `teams/managedCases` endpoint
- Removed the `currentOm` filter and replaced with the `?current=true` query parameter

### DIFF
--- a/server/services/caseloadService.ts
+++ b/server/services/caseloadService.ts
@@ -2,7 +2,7 @@ import moment from 'moment'
 import CommunityService from './communityService'
 import PrisonerService from './prisonerService'
 import LicenceService from './licenceService'
-import { CaseTypeAndStatus, ManagedCase } from '../@types/managedCase'
+import { CaseTypeAndStatus } from '../@types/managedCase'
 import LicenceStatus from '../enumeration/licenceStatus'
 import LicenceType from '../enumeration/licenceType'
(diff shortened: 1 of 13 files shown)


server/services/caseloadService.ts
```diff
@@ -81,41 +82,66 @@ export default class CaseloadService {
...
-    const managedOffenders = await this.communityService.getManagedOffenders(deliusStaffIdentifier)
-    const caseloadNomisIds = managedOffenders
-      .filter(offender => offender.currentOm)
-      .filter(offender => offender.nomsNumber)
-      .map(offender => offender.nomsNumber)
-
-    return this.licenceService.getLicencesByNomisIdsAndStatus(caseloadNomisIds, [LicenceStatus.ACTIVE], user)
-  }
-
   private getLicenceType = (tused: string, led: string, sed: string): LicenceType => {
```
timharrison-moj, line 145: Good reorganisation.

timharrison-moj (approved):
LGTM. Like the reorganisation of the caseload service. Caching potentially to add, but we can try this and see if / where issues are.

PR #183, opened 2022-01-25
CVSL-341: Refactor OMU case view and COM details page to pull probation practitioner details from delius + add probation practitioner column to COM caseload

smcveigh941:
- Moved OMU caseload generation to `caseloadService`
- Probation practitioner details are now pulled from Delius - by a list of comUsernames in the case of the OMU caseload
- In the case of the COM caseload, COM details are pulled from the `managedOffenders` endpoint
- OMU staff will only be able to see details of COMs who are associated with a licence they are allowed to see
- COM staff will only be able to see details of COMs in their team
- Side note - I don't think any COM details should be stored on the licence at this stage, apart from `comUsername` or `staffId`. It introduces an issue with COM details changing in delius. If we want to store COM details our side, we should put them in a separate table, and reference them to licences with an ID.


server/views/pages/view/cases.njk
```diff
@@ -48,28 +45,24 @@
         </div>
     </div>
 
-    <form method="POST">
```
smcveigh941, line 51: Not sure why this was ever a POST? No data is posted and the handler was just redirecting... can be a GET

server/data/communityApiClient.ts
```diff
@@ -5,19 +5,25 @@ import {
...
 
-  async getStaffDetail(user: User): Promise<CommunityApiStaffDetails> {
+  async getStaffDetailByUsername(username: string): Promise<CommunityApiStaffDetails> {
     return (await this.get({
-      path: `/secure/staff/username/${user.username}`,
+      path: `/secure/staff/username/${username}`,
     })) as Promise<CommunityApiStaffDetails>
   }
 
+  async getStaffDetailsByUsernameList(usernames: string[]): Promise<CommunityApiStaffDetails[]> {
```
smcveigh941, line 20: Endpoint to get staff details by a list of comUsernames. Used to load COM column on OMU caseload

server/routes/viewingLicences/index.ts
```diff
@@ -22,26 +27,16 @@ export default function Index({ licenceService, prisonerService, qrCodeService }
...
-  const comDetailsHandler = new ComDetailsRoutes()
+  const comDetailsHandler = new ComDetailsRoutes(communityService)
 
   get('/cases', viewCasesHandler.GET)
-  post('/cases', viewCasesHandler.POST)
   get('/id/:licenceId/show', viewLicenceHandler.GET)
   get('/id/:licenceId/html-print', printHandler.preview)
   get('/id/:licenceId/pdf-print', printHandler.renderPdf)
-  get('/id/:licenceId/com-details', comDetailsHandler.GET)
+  get('/id/:licenceId/probation-practitioner', comDetailsHandler.GET)
```
smcveigh941, line 39: COMs will only have access to here if they are allowed to view the licence, access control is handled by `fetchLicenceMiddleware`

server/services/caseloadService.ts
```diff
@@ -142,6 +147,21 @@ export default class CaseloadService {
       })
   }
 
+  private async mapLicencesAndResponsibleComs(licences: LicenceSummary[]): Promise<LicenceAndResponsibleCom[]> {
+    const comUsernames = licences.map(licence => licence.comUsername)
+    const coms = await this.communityService.getStaffDetailsByUsernameList(comUsernames)
```
smcveigh941, line 152: For each licence, get the responsible COMs in bulk, and match them to the licences based on username

server/services/licenceService.ts
```diff
@@ -274,37 +274,15 @@ export default class LicenceService {
...
-        statuses,
-        filteredPrisons,
-        [],
-        [],
-        'conditionalReleaseDate',
-        null,
-        user
-      )
-    }
-    if (user.authSource === 'delius') {
```
smcveigh941, line 296: Took this out as it doesn't make sense for a probation user to see the OMU caseload. Perhaps if we need COMs to see active licences later we will have a different service call

server/views/pages/comDetails.njk
```diff
@@ -25,32 +25,32 @@
...
                     }
                 ]
             }) }}
 
             {{ govukButton({
                 text: 'Return to case list',
-                href: '/licence/view/cases',
-                classes: 'govuk-!-margin-top-6',
+                href: returnLink,
+                classes: 'govuk-!-margin-top-6 js-backlink',
```
smcveigh941, line 53: Using the `js-backlink` class, so that the correct tab will be selected when returning to caseload

server/views/pages/view/cases.njk
timharrison-moj: It might have been a POST redirected to a GET to avoid the back button asking for resubmission of form values? There is a pattern used to avoid these.

timharrison-moj: https://stackoverflow.com/questions/10827242/understanding-the-post-redirect-get-pattern

server/services/caseloadService.ts
timharrison-moj: Could it be quite a big caseload? Lots of matching to do? Any impact on performance?

server/services/licenceService.ts
timharrison-moj: I think COMs do need to see ACTIVE cases and its on the Vary caseload?

timharrison-moj (approved):
LGTM. Quite a big reorganisation considering the original issue was to prevent the viewing of COM details by people who are not in their team, or for prison users, where their caseload includes the prison where the offender resides. If you're happy it meets that and generally looks to improve caseload processing, them I'm happy too. Couple of minor questions added.

server/services/caseloadService.ts
smcveigh941: I don't think so, the COMs are fetched in bulk, and javascript core functions used to map and find, so should be efficiencies built-in

smcveigh941: Can test it out though

server/services/licenceService.ts
smcveigh941: This is handled by the `getVaryCaseload` function in this case

server/services/caseloadService.ts
timharrison-moj: Lets do that in DEV or preprod?

PR #184, opened 2022-01-26
CVSL-347: Added CSRF token to the remove upload link.

timharrison-moj:
Small change to pass through formBuilder, add link and update view context for tests.

### DIFF
--- a/server/views/formBuilder.njk
+++ b/server/views/formBuilder.njk
@@ -11,7 +11,7 @@
 {# TODO: Currently, this form builder only supports the "add another" functionality for the text input type. #}
 {# It can be extended in future if the other field types require Add Another functionality #}
 
-{%  macro formBuilder(licenceId, config, additionalCondition, validationErrors, formResponses) %}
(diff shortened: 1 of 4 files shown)


server/views/formBuilder.njk
```diff
@@ -11,7 +11,7 @@
 {# TODO: Currently, this form builder only supports the "add another" functionality for the text input type. #}
 {# It can be extended in future if the other field types require Add Another functionality #}
 
-{%  macro formBuilder(licenceId, config, additionalCondition, validationErrors, formResponses) %}
+{%  macro formBuilder(licenceId, config, additionalCondition, validationErrors, formResponses, csrfToken) %}
```
smcveigh941, line 14: I took a few moments to research how the CSRF token could be made a global variable within the nunjucks environment which would enable it to be accessed from within macros. Not nice that it needs to be passed as a parameter but also not easy to add variables globally to the nunjucks environment on each request as the environment is initialised at startup. The same could be said for `validationErrors` and `formResponses`, perhaps a future task to make the form builder see the nunjucks context which would remove the need for parameters.

Fine for now - the issue is not exclusive to this change!

smcveigh941 (approved):

PR #186, opened 2022-01-26
CVSL-346: New decorator and validator to avoid markup in address fields.

timharrison-moj:
* Altered `html: xxxx` to `text: xxxx` on address fields - they were different to others.
* The `noMarkup` validator has been left in for now + tests - but can remove later if we don't have a use for it, or similar.
* Added missing error message for addressLine2 - did not show error message for this optional field.

### DIFF
--- a/server/middleware/validationMiddleware.test.ts
+++ b/server/middleware/validationMiddleware.test.ts
@@ -1,6 +1,6 @@
 // eslint-disable-next-line max-classes-per-file
 import { Request, Response } from 'express'
-import { IsAlphanumeric, IsIn, IsNotEmpty, ValidateNested } from 'class-validator'
+import { IsIn, IsNotEmpty, ValidateNested } from 'class-validator'
 import { Expose, Type } from 'class-transformer'
 import validationMiddleware from './validationMiddleware'
 
--- a/server/middleware/validationMiddleware.test.ts
+++ b/server/middleware/validationMiddleware.test.ts
@@ -6,7 +6,7 @@ import validationMiddleware from './validationMiddleware'
 
 import * as conditionsProvider from '../utils/conditionsProvider'
 import IsValidExclusionZoneFile from '../validators/isValidExclusionZoneFile'
-import DoesNotContainMarkup from '../validators/doesNotContainMarkup'
+import NoMarkup from '../validators/noMarkup'
 
 const conditionsProviderSpy = jest.spyOn(conditionsProvider, 'getAdditionalConditionByCode')
 
@@ -48,7 +48,7 @@ describe('validationMiddleware', () => {
 
     class DummyAddress {
       @Expose()
-      @DoesNotContainMarkup({ message: 'Markup disallowed' })
+      @NoMarkup({ message: 'Markup disallowed' })
       addressLine: string
     }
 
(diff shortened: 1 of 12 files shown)


server/middleware/validationMiddleware.test.ts
```diff
@@ -45,6 +46,12 @@ describe('validationMiddleware', () => {
       outOfBoundFilename: string
     }
 
+    class DummyAddress {
```
smcveigh941, line 49: Does this cover anything which `server/validators/doesNotContainMarkup.test.ts` doesnt?

smcveigh941:
Were you able to find out why it only happens in the address field and not any other field?

server/validators/doesNotContainMarkup.ts
```diff
@@ -0,0 +1,18 @@
+import { registerDecorator, ValidationArguments, ValidationOptions } from 'class-validator'
+
+export default function DoesNotContainMarkup(validationOptions?: ValidationOptions) {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const doesNotContainMarkup = (value: string, { object }: ValidationArguments) => {
+    return !(value.includes('<') || value.includes('>'))
```
smcveigh941, line 6: Not sure that disallowing characters is the way to go. We should just escape the content that the user has entered..

If we could find out why the markup is escaped in other fields but not this field, it might go a ways towards finding the real issue we have

smcveigh941: The way we concatenate the address with commas isn't good either, if the user enters "text1, test2" as a single line, it will get parsed as 2 lines because of the comma

smcveigh941:
I wondered why this issue didn't exist for the address fields in the additional conditions... After a bit of digging, I found that nunjucks already escapes html for us, see the following config

```
const njkEnv = nunjucks.configure(
    {
      autoescape: true,
      express: app,
    }
  )

```

The only issue is that in `inductionMeetingDetailsSummary.njk` we have this which is explicitly telling nunjucks to render the address value as html

```
value: { html: licence.appointmentAddress | formatAddress },
```

The only thing we need to do is change it to this, telling nunjucks to render it as text, not markup

```
value: { text: licence.appointmentAddress | formatAddress },
```

server/middleware/validationMiddleware.test.ts
timharrison-moj: Well, it's a final check on a wider-scoped function - this checks via the `validationMiddleware` whilst the other tests are limited to the `validate()` function.

server/validators/doesNotContainMarkup.ts
timharrison-moj: Agree with the address fields concatenated into the DB - all a bit awkward. I did initially escape the values, but found that this caused further issues in saving/reading values to/from the DB via the API - it depends how invalid characters are escaped. Also, tried limiting characters in a nunjucks `picture` - but clearer for the user when validated together. The flip-side is that there should never be < or > characters in addresses - and avoiding these getting into the system, with a clear validation message, does tell the user what is wrong with their input and why they aren't allowed.

smcveigh941 (approved):
LGTM

PR #188, opened 2022-01-28
CVS-295: Add event handler to update in-flight licences with the details of a COM newly allocated to an offender

smcveigh941:


server/services/communityService.ts
```diff
@@ -33,14 +34,18 @@ export default class CommunityService {
...
+  }
+
   async searchProbationers(searchCriteria: SearchDto): Promise<OffenderDetail[]> {
     return this.probationSearchApiClient.searchProbationer(searchCriteria)
   }
 
-  async getProbationer(nomisId: string): Promise<OffenderDetail> {
-    const deliusRecords = await this.probationSearchApiClient.searchProbationer({ nomsNumber: nomisId })
+  async getProbationer(searchDto: SearchDto): Promise<OffenderDetail> {
+    const deliusRecords = await this.probationSearchApiClient.searchProbationer(searchDto)
```
timharrison-moj, line 46: Not sure why the change of parameter type? Do we search here by different values?

timharrison-moj (approved):
LGTM.

server/services/communityService.ts
smcveigh941: No reason, I was going to use the endpoint to search for offenders by CRN but didn't need to in the end

PR #191, opened 2022-01-28
CVSL-239 changes from design review

timharrison-moj:
Contains:
* Small wording changes from design on conditions.
* Altering the prefix for data values provided that do/don't start with a vowel (a / an)


server/utils/conditionsProvider.ts
```diff
@@ -119,6 +125,15 @@ export const produceValueAsFormattedList = (listType = 'AND', matchingDataItems:
   return value
 }
 
+/**
+ * Checks whether the first letter of a string value starts with a vowel.
+ * @param value
+ */
+const startsWithAVowel = (value: string): boolean => {
+  const regex = new RegExp('^[aeiou].*', 'i')
```
smcveigh941, line 133: Does this cover capital letters?

server/config/conditions.ts
```diff
@@ -624,6 +626,8 @@ export default {
             label: 'Select a review period',
             name: 'reviewPeriod',
             case: 'lower',
+            beforeVowel: 'an ',
```
smcveigh941, line 629: The value of `beforeVowel` and `beforeConsonant` will always be `a` and `and` respectively right? Instead of writing a value in each place its used, maybe just a flag like this?

`handleIndefiniteArticle: true`

server/utils/conditionsProvider.ts
```diff
@@ -75,13 +75,19 @@ export function expandAdditionalConditions(conditions: AdditionalCondition[]): A
...
+            value = startsWithAVowel(value) ? `${rules.beforeVowel}${value}` : `${rules.beforeConsonant}${value}`
+          }
           conditionText = replacePlaceholderWithValue(ph, conditionText, value)
         } else {
           // List of values for this placeholder (lists of values can have listType 'AND' or 'OR')
           const rules = inputConfig.find(item => item.name === fieldName)
           let value = produceValueAsFormattedList(rules?.listType as string, matchingDataItems)
           value = adjustCase(rules?.case as string, value)
           value = rules?.includeBefore ? `${rules.includeBefore}${value}` : `${value}`
+          if (rules?.beforeVowel) {
```
smcveigh941, line 88:
```suggestion
          if (rules?.handleIndefiniteArticle) {
              value = startsWithAVowel(value) ? `an ${value}` : `a ${value}`
          }
```

smcveigh941 (approved):
Looks good but have a suggestion to avoid repetitive definition of the indefinite article

server/utils/conditionsProvider.ts
timharrison-moj: Yes - the "i" option ignores case.

server/utils/conditionsProvider.ts
timharrison-moj: Yes, seems better so will look to add over the weekend. Thanks.

timharrison-moj: A quick read suggests the handling of indefinite articles should be a bit more complex as there are a number of edge cases, but this will catch the majority. Things like `hourly` or `honourable` should be an `an` word, but don't start with a vowel. However, in this context, we are unlikely to hit many of these exceptions.

smcveigh941 (approved):
LGTM

PR #195, opened 2022-02-04
CVSL-336 record auditable events

timharrison-moj:
Adds calls to record audit events at key points in the journeys:
* Licence created
* Licence edited
* Licence submitted for approval
* Licence approved
* Licence rejected
* Licence viewed
* Licence printed

It felt a bit like littering the code with extra calls in key places :-( 

Some things to consider in future:
* Switch to use the `hmpps-audit service` - similar to ours, and offers pageable queries and a front-end service, but does not allow calls by a time range (currently). It also requires ROLE_AUDIT, whereas we might want to allow people within our service access to audit trails, so with ROLE_LICENCE_XXX. Pros and cons with this option.

* Move more of this audit stuff to the API, and record based on changing status or other values/operations e.g. create a licence. The issue here is that we don't currently know whether a `fetchLicence` is for view, print, or an approve, or just for moving to the next form in the creation process. If we could add a `fetch-type` in some way, that might offer more leeway to move auditing into the API. For some key operations, the only interaction with the API is a fetch (and without a username lookup, its hard to get the full user details to record).

* Other auditable events to add in future 
** Licence set to active
** Licence recalled
** Licence superseded
** Licence varied
** Licence variation edited/submitted/approved/rejected
** Licence reassigned COM
** Licence updated - transfer, detail updated, dates changed

* Make auditable events have individual types eg. AUDIT_TYPE_CREATE_LICENCE, and set these up with template wording that only requires values passing in.

image


server/data/licenceApiClient.ts
```diff
@@ -202,4 +204,18 @@ export default class LicenceApiClient extends RestClient {
       { username: user?.username }
     )
   }
+
+  async recordAuditEvent(auditEvent: AuditEvent, user?: User): Promise<void> {
+    await this.put({ path: `/audit/save`, data: auditEvent }, { username: user?.username })
```
smcveigh941, line 209: Should we be passing the username to record audit events? Feels like this is a case for using the system token

server/services/licenceService.ts
```diff
@@ -290,6 +297,43 @@ export default class LicenceService {
...
+    user: User
+  ): Promise<AuditEvent[]> {
+    const requestBody = {
+      username: forUsername || null,
+      licenceId: forLicenceId || null,
+      startTime: moment(startTime).format('DD/MM/YYYY hh:mm:ss'),
+      endTime: moment(endTime).format('DD/MM/YYYY hh:mm:ss'),
+    } as AuditRequest
+
+    return this.licenceApiClient.getAuditEvents(requestBody, user)
```
yasinmustafa, line 334: Dont you need an await here?!? probably not if it's tested OK!

smcveigh941 (commented):
Does feel littered.. why do we need to audit that someone has viewed the licence?? A DB insert for every refresh of a page? Seems a bit much to me. Would we not just need to audit when data changes? In which case DB triggers and history tables would seem more suitable. At the least I would have hoped that auditing would be isolated to just the API layer (i.e. VIEW event is recorded from within the `getLicence` endpoint).

smcveigh941 (approved):
Approved on sound code, although I think there may be problems with the approach and question the need for inserting rows into the DB on a GET request

PR #203, opened 2022-02-17
CVSL-393 refactor probation locations in the UI service

timharrison-moj:
This PR needs to go in after the API PR of the same ticket number, and the `delius-wiremock` changes, and refactors the probation location structure in the same way:

     `Probation Area ->  PDU (or borough) -> LAU (or district) ->  Team  ->  Staff`

Changes:

* Probation area, borough (PDU), district (LAU) and team are now stored on the licence at the time of creation
* Access checks were formerly done via the staff membership of an LDU - this is now changed to team membership.


server/@types/licenceApiImport/index.d.ts
```diff
@@ -88,41 +88,25 @@ export interface paths {
...
+      /** The unique identifier of the COM, retrieved from Delius */
       staffIdentifier: number
-      /**
-       * @description The Delius username for the COM
-       * @example jbloggs
-       */
+      /** The Delius username for the COM */
       staffUsername: string
-      /**
-       * @description The email address of the COM
```
smcveigh941, line 105: We must have different configurations or something? Mine generates these `description` and `example` tags

PR #212, opened 2022-02-28
CVSL-334: Use the new community API endpoints for getting staff and team caseloads from delius

smcveigh941:


server/services/caseloadService.test.ts
```diff
@@ -1137,35 +1355,14 @@ describe('Caseload Service', () => {
       )
     })
 
-    it('should filter managed offenders returned from community API without a nomis ID', async () => {
```
timharrison-moj, line 1140: Is this one not needed?

timharrison-moj (approved):
LGTM. I like the split of data into `nomisRecord` and `deliusRecord` for caseload building.

PR #216, opened 2022-03-01
CVSL-417: Varied conditions comparison

smcveigh941:
- Implemented the following algorithm to compare two lists
- Added a custom CSS component, which would ideally be added to the MoJ design library. 

```
sort left-list and right-list
adds = {}
deletes = {}
get first right-item from right-list
get first left-item from left-list
while (either list has items)
  if left-item < right-item or right-list is empty
    add left-item to deletes
    get new left-item from left-list
  else if left-item > right-item or left-list is empty
    add right-item to adds
    get new right-item from right-list
  else
    get new right-item from right-list
    get new left-item from left-list
```

screencapture-localhost-3000-licence-vary-id-4-reason-for-variation-2022-03-01-19_02_44

### DIFF
--- a/server/utils/licenceComparator.test.ts
+++ b/server/utils/licenceComparator.test.ts
@@ -1,0 +1,281 @@
+import _ from 'lodash'
+import compareLicenceConditions from './licenceComparator'
+import * as conditionsProvider from './conditionsProvider'
+import { Licence } from '../@types/licenceApiClientTypes'
+
+jest.spyOn(conditionsProvider, 'expandAdditionalConditions').mockImplementation(conditions => conditions)
+
+const licenceTemplate = {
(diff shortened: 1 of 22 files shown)


server/utils/licenceComparator.test.ts
```diff
@@ -0,0 +1,281 @@
...
+
+    expect(compareLicenceConditions(originalLicence, _.cloneDeep(originalLicence))).toEqual({
+      licenceConditionsAdded: [],
+      licenceConditionsAmended: [],
+      licenceConditionsRemoved: [],
+      pssConditionsAdded: [],
+      pssConditionsAmended: [],
+      pssConditionsRemoved: [],
+    })
+  })
```
timharrison-moj, line 280: Good tests 👍

PR #217, opened 2022-03-02
CVSL-425: Updating dependencies (resolve overnight check-outdated fails)

timharrison-moj:


package.json
```diff
@@ -105,7 +105,6 @@
     "connect-redis": "^6.0.0",
     "cookie-session": "^1.4.0",
     "csurf": "^1.11.0",
-    "cypress-axe": "^0.13.0",
```
smcveigh941, line 108: `axe-core` is also listed as a dependency, can it be moved to devDependencies?

timharrison-moj: Yes, missed that one. I'll move it.

smcveigh941 (approved):
LGTM

PR #218, opened 2022-03-02
CVSL-256: If LED and SED are different, print them separately on the licence

smcveigh941:


server/listeners/eventHandlers/prisonEvents/sentenceDatesChangedEventHandler.ts
```diff
@@ -20,24 +20,22 @@ export default class SentenceDatesChangedEventHandler {
     if (licence) {
       const prisoner = await this.prisonerService.getPrisonerDetail(nomisId)
 
-      await Promise.all([
```
smcveigh941, line 23: Turns out updating the same licence via two different endpoints concurrently causes some unexpected behaviour (i.e. status wasnt getting updated because the other endpoint was updating the same licence at the exact same time). Now calling these endpoints in sequence

server/views/pages/licence/partials/headshot.njk
```diff
@@ -1,5 +1,5 @@
 {% if htmlPrint %}
-    <img id="offender-image" src="/prisoner/{{ licence.nomsId or 'placeholder' }}/image" width="90px" height="120px" alt="offender headshot">
```
smcveigh941, line 2: Inline attribute moved to CSS

server/views/pages/licence/AP.njk
```diff
@@ -24,9 +24,18 @@
             <p>
                 Date of birth: <span class="bold">{{ licence.dateOfBirth | datetimeToDate }}</span>
             </p>
-            <p>
-                Your licence and sentence expire on <span class="bold">{{ licence.licenceExpiryDate | datetimeToDate }}</span>
-            </p>
+            {% if licence.licenceExpiryDate == licence.sentenceEndDate %}
+                <p>
```
timharrison-moj, line 28: Should it be a triple === ?

timharrison-moj (approved):
LGTM. One question on equality operator on dates.

server/views/pages/licence/AP.njk
smcveigh941: I don't think it makes a difference. They cant be anything except strings

PR #219, opened 2022-03-02
CVSL-350: Add missing apostrophe on licence document

smcveigh941:


timharrison-moj (approved):
Not sure I agree with the placing of the apostrophe, but recognise it's one way of describing it. Correct or not, LGTM.

PR #220, opened 2022-03-03
Change SQS event handlers to take eventType from MessageAttributes instead of the body

smcveigh941:
### Background

Domain and prison events work fine in production. These events have the `eventType` inside the body of the SQS message. Probation events **do not**. This means our probation event handler is wrong, and is not properly processing the events raised in production.

See how offender-events-ui gets its event type for all events [here](https://github.com/ministryofjustice/offender-events-ui/blob/7c28797020e5fdd4f500c21aa44829b081f07562/src/main/kotlin/uk/gov/justice/hmpps/offenderevents/service/OffenderEventListener.kt#L32)

---
 
### Fix

- All 3 events have the `eventType` in the `MessageAttributes` of the SQS message, so we should get the message type from there instead.
- Integration tests to verify change


server/listeners/eventHandlers/probationEvents/offenderManagerChangedEventHandler.ts
```diff
@@ -1,15 +1,11 @@
-import { ProbationEvent } from '../../../@types/events'
+import { ProbationEventMessage } from '../../../@types/events'
 import CommunityService from '../../../services/communityService'
 import LicenceService from '../../../services/licenceService'
 
 export default class OffenderManagerChangedEventHandler {
   constructor(private readonly communityService: CommunityService, private readonly licenceService: LicenceService) {}
 
-  // TODO: I suspect this is an incorrect interpretation of the meaning of this event?
```
smcveigh941, line 8: Think we have agreed that this listener is doing as it should, and another listener would be added to handle the OFFICER_DETAIL_CHANGED event

integration_tests/integration/eventHandlers.spec.ts
```diff
@@ -50,11 +62,18 @@ context('Event handlers', () => {
     it('should listen to the offender manager changed event and call endpoint to update responsible COM', () => {
       cy.task('stubGetAnOffendersManagers')
       cy.task('stubGetStaffDetailsByStaffId')
+      cy.task('stubUpdateResponsibleCom')
```
smcveigh941, line 65: This was not already stubbed. The test did not fail for some reason.. but it threw a warning

server/listeners/eventHandlers/probationEvents/offenderManagerChangedEventHandler.ts
timharrison-moj: Yes, agreed.

integration_tests/integration/eventHandlers.spec.ts
```diff
@@ -13,7 +13,13 @@ context('Event handlers', () => {
       cy.task(
         'sendDomainEvent',
         `{
-            "Message" :  "{\\"eventType\\":\\"prison-offender-events.prisoner.released\\",\\"additionalInformation\\":{\\"nomsNumber\\":\\"A7774DY\\",\\"reason\\":\\"RELEASED\\",\\"details\\":\\"Movement reason code CR\\",\\"currentLocation\\":\\"OUTSIDE_PRISON\\",\\"prisonId\\":\\"MDI\\",\\"currentPrisonStatus\\":\\"NOT_UNDER_PRISON_CARE\\"},\\"version\\":1,\\"occurredAt\\":\\"2022-01-12T14:56:51.662128Z\\",\\"publishedAt\\":\\"2022-01-12T14:58:25.021008001Z\\",\\"description\\":\\"A prisoner has been released from prison\\"}"
+            "Message": "{\\"additionalInformation\\":{\\"nomsNumber\\":\\"A7774DY\\",\\"reason\\":\\"RELEASED\\",\\"details\\":\\"Movement reason code CR\\",\\"currentLocation\\":\\"OUTSIDE_PRISON\\",\\"prisonId\\":\\"MDI\\",\\"currentPrisonStatus\\":\\"NOT_UNDER_PRISON_CARE\\"},\\"version\\":1,\\"occurredAt\\":\\"2022-01-12T14:56:51.662128Z\\",\\"publishedAt\\":\\"2022-01-12T14:58:25.021008001Z\\",\\"description\\":\\"A prisoner has been released from prison\\"}",
```
timharrison-moj, line 16: I thought domain events were ok and it was the probation events that were the issue?

timharrison-moj (approved):
LGTM

PR #225, opened 2022-03-10
CVSL-433: Restrict access to Wales probation area and prisons in production

timharrison-moj:
Switches on RESTRICT_ROLLOUT env variable in production.
Configures the set of prisons and probation areas which are allowed to access.


joelstobart-moj (approved):
LGTM

PR #227, opened 2022-03-12
CVSL-363: Prompt licence creation

smcveigh941:


server/services/caseloadService.test.ts
```diff
@@ -1485,6 +1485,23 @@ describe('Caseload Service', () => {
         },
       ] as OffenderDetail[])
 
+      prisonerService.searchPrisonersByNomisIds.mockResolvedValue([
```
smcveigh941, line 1488: Unit test file for `caseloadService` has become quite messy, need to come back to tidy it up

helm_deploy/create-and-vary-a-licence/templates/prompt-licence-create-cronjob.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: activate-licences
+spec:
```
timharrison-moj, line 5: Wrong name? Or same name as another job?

jobs/promptLicenceCreation.ts
```diff
@@ -0,0 +1,152 @@
...
+      .searchPrisonersByPrison(prisonCode)
+      .then(prisoners => filterPrisonersReleasedWithinNoticePeriod(prisoners))
+      .then(prisoners => caseloadService.mapOffendersToLicences(prisoners))
+      .then(prisoners =>
+        prisoners
+          .filter(
+            prisoner =>
+              prisoner.licenceStatus === LicenceStatus.NOT_STARTED ||
+              prisoner.licenceStatus === LicenceStatus.IN_PROGRESS
+          )
```
timharrison-moj, line 37: Are these status values right? Filter to include licences IN_PROGRESS? I would expect only for NOT_STARTED licences to find their way into the email notifications?

server/data/prisonerSearchApiClient.ts
```diff
@@ -28,4 +28,14 @@ export default class PrisonerSearchApiClient extends RestClient {
...
   }
+
+  async searchPrisonersByPrison(prisonCode: string, user?: User): Promise<PagePrisoner> {
+    return (await this.get(
+      {
+        path: `/prisoner-search/prison/${prisonCode}`,
+        query: { size: 2000 },
+      },
+      { username: user?.username }
+    )) as Promise<PagePrisoner>
```
timharrison-moj, line 39: If we have any prisons with > 2000 offender in them we probably won't raise alerts? Should we set a smaller page size and iterate over  many pages as there are, or do you think this will suffice?

timharrison-moj (approved):
LGTM. Small number of questions to consider.

timharrison-moj (approved):
LGTM

PR #228, opened 2022-03-14
CVSL-441: Upgrade docker containers for ARM support

smcveigh941:
- Apple have stopped manufacturing Macbooks with intel x86_64 chips
- New Macbooks have ARM architecture with the new M1 chip
- Some of the docker containers we use locally have not been packaged for ARM architecture, which means that running them locally either fails or is very slow, as the x86_64 architecture needs to be emulated
- Newer versions of these docker images do support ARM, so upgrading to those new versions solves this issue

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ executors:
     docker:
       - image: 'cimg/node:<<parameters.node_tag>>'
       - image: 'circleci/redis:<<parameters.redis_tag>>'
-      - image: 'localstack/localstack:0.11.2'
+      - image: 'localstack/localstack:0.13.2'
         environment:
           SERVICES: sqs
           DATA_DIR: /tmp/localstack/data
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -141,17 +141,17 @@ jobs:
             export AWS_ACCESS_KEY_ID=foo
             export AWS_SECRET_ACCESS_KEY=bar
 
-            aws --endpoint-url=http://localhost:4576 sqs create-queue --queue-name create_and_vary_a_licence_prison_events_queue
-            aws --endpoint-url=http://localhost:4576 sqs create-queue --queue-name create_and_vary_a_licence_prison_events_queue_dl
-            aws --endpoint-url=http://localhost:4576 sqs set-queue-attributes --queue-url "http://localhost:4576/queue/create_and_vary_a_licence_prison_events_queue" --attributes '{"RedrivePolicy":"{\"maxReceiveCount\":\"3\", \"deadLetterTargetArn\":\"arn:aws:sqs:eu-west-2:000000000000:create_and_vary_a_licence_prison_events_queue_dl\"}"}'
+            aws --endpoint-url=http://localhost:4566 sqs create-queue --queue-name create_and_vary_a_licence_prison_events_queue
```
smcveigh941, line 144: Port 4576 is deprecated in `v0.11.2` and removed in `v0.13.2`, use port 4566

docker-compose-dev.yml
```diff
@@ -2,7 +2,7 @@ version: '3.1'
 services:
 
   redis:
-    image: 'bitnami/redis:5.0'
+    image: 'redis:5.0'
```
smcveigh941, line 5: The bitnami distribution does not build containers with ARM support. Recommended to use redis from the official source. This is only used for running locally anyway

helm_deploy/create-and-vary-a-licence/values.yaml
```diff
@@ -96,7 +96,7 @@ gotenberg:
 
   image:
     repository: thecodingmachine/gotenberg
-    tag: 6.3.0
+    tag: 7.5.0
```
smcveigh941, line 99: gotenberg is built with ARM support from `7.5.0` onwards

timharrison-moj (commented):
LGTM.

timharrison-moj (approved):
LGTM

PR #235, opened 2022-03-18
CVSL-443: Contact and about us pages

timharrison-moj:
Also fixes health endpoint for docker-compose Gotenberg container.

### DIFF
--- a/server/views/layout.njk
+++ b/server/views/layout.njk
@@ -91,6 +91,25 @@
   <span class="govuk-visually-hidden" id="{{ pageId }}"></span>
 {% endblock %}
 
+{% block footer %}
+  {{ govukFooter({
+    meta: {
+      items: [
+        {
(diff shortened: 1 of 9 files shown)


server/views/layout.njk
```diff
@@ -91,6 +91,25 @@
...
+          href: "/about",
+          text: "About",
+          attributes: { 'target': '_blank' }
+        },
+        {
+          href: "/contact",
+          text: "Contact",
+          attributes: { 'target': '_blank' }
+        }
+      ]
```
smcveigh941, line 108: Accessibility page still to come?

smcveigh941 (approved):
LGTM

server/views/layout.njk
timharrison-moj: And still no actual content for the About us, so a placeholder for now.

PR #244, opened 2022-03-24
CVSL-396: Convert match licences endpoint to a POST method

smcveigh941:


server/services/licenceService.test.ts
```diff
@@ -577,9 +577,9 @@ describe('Licence Service', () => {
     expect(licenceApiClient.matchLicences).toBeCalledWith(
       ['SUBMITTED'],
       ['MDI'],
-      [],
-      [],
-      [],
+      null,
```
smcveigh941, line 580: `[]` would be like saying "Find me all licences which match any of the things in this empty list", which would always be an empty set. Null makes more sense. Beforehand, superagent querystring builder would have removed empty array anyway but a post does not do that.

timharrison-moj: When they landed in the API, there is code to ignore nulls/empty lists anyway ie. `.let? { do some stuff }` so either would do I think? Fine with nulls here.

timharrison-moj (approved):
LGTM

PR #245, opened 2022-03-24
CVSL-467: Updated version of minimist to fix security vulnerabilities

timharrison-moj:
I ran through both create and vary journeys to check this as not altered behaviour.


yasinmustafa (approved):
LGTM

PR #246, opened 2022-03-24
CVSL-471: Fix docker cacheing

smcveigh941:
See template change here https://github.com/ministryofjustice/hmpps-template-typescript/pull/74


timharrison-moj (approved):
LGTM

PR #249, opened 2022-03-24
CVSL-471: Fix dockerfile to update before installing make and python

smcveigh941:


Dockerfile
```diff
@@ -22,10 +22,12 @@ RUN apt-get update && \
 
 # Stage: build assets
 FROM base as build
-ARG BUILD_NUMBER
-ARG GIT_REF
 
-RUN apt-get install -y make python g++
+ARG BUILD_NUMBER=1_0_0
+ARG GIT_REF=not-available
+
```
timharrison-moj, line 28: Do these get overwritten with the proper values in the second build stage? Just checking that the dps-monitor will retain the actual Git ref and build version

Dockerfile
```diff
@@ -22,10 +22,12 @@ RUN apt-get update && \
...
+RUN apt-get update && \
+    apt-get install -y make python g++
```
timharrison-moj, line 30: Why python and g++ here?

timharrison-moj (approved):
Couple of questions but LGTM

Dockerfile
smcveigh941: These were here already, unsure what its for... I just added `apt-get update`

Dockerfile
smcveigh941: Yes seems to be ok, this is copied direct from the template repo

Screenshot 2022-03-24 at 18 54 37

PR #251, opened 2022-03-25
CVSL-468: Upgrade NPM dependencies

smcveigh941:
Upgraded NPM packages, some with breaking changes


bin/build-css.sh
```diff
@@ -5,6 +5,7 @@
 
 # Build the application stylesheet
 ./node_modules/.bin/sass $@ \
+     --load-path=. \
```
smcveigh941, line 8: New version of mon-frontend refers to node_modules directory which is in the root of the project

server/middleware/validationMiddleware.ts
```diff
@@ -1,4 +1,4 @@
-import { ClassConstructor, plainToClass } from 'class-transformer'
+import { ClassConstructor, plainToInstance } from 'class-transformer'
```
smcveigh941, line 1: `plainToClass` was renamed to `plainToInstance` to better reflect what is does

timharrison-moj (approved):
LGTM

PR #259, opened 2022-03-31
CVSL-359: Store expanded conditions on the licence directly

smcveigh941:
* This is simply a revert of a revert. A lot of files but it is identical to a previous PR which was already reviewed and approved


timharrison-moj (approved):
LGTM

PR #265, opened 2022-04-04
CVSL-478: Correcting the names of user properties for GA

timharrison-moj:
* Two properties had changed name but not updated in the layout.njk where data is pushed to GA.


smcveigh941 (approved):
LGTM

PR #271, opened 2022-04-05
CVSL-454: Split banner links into two

smcveigh941:


timharrison-moj (approved):
LGTM. Would like to check this in DEV though before promoting to prod.

PR #272, opened 2022-04-06
CVSL-496: Reduce number of offenders searched in delius from OMU caseview

smcveigh941:
- by filtering to only those cases leaving within 13 weeks


timharrison-moj (approved):
LGTM. Would like a confirmation that 13 weeks is the right period though!

timharrison-moj (commented):
LGTM

PR #273, opened 2022-04-06
CVSL-498: Fix POM being incorrectly identified as the COM

smcveigh941:
- Identify the COM based on a cross-reference from probation-offender-search, rather than trusting the isResponsibleOfficer flag from community API


server/listeners/eventHandlers/probationEvents/offenderManagerChangedEventHandler.ts
```diff
@@ -7,8 +7,12 @@ export default class OffenderManagerChangedEventHandler {
 
   handle = async (event: ProbationEventMessage): Promise<void> => {
     const { crn } = event
+    const deliusRecord = await this.communityService.getProbationer({ crn })
```
smcveigh941, line 10: This is the endpoint which is more accurate.. but still need to cross reference with `/allOffenderManagers` because it has the required data fields (username)

timharrison-moj (approved):
LGTM.

PR #277, opened 2022-04-08
CVSL-501: Fix HDC eligibility check to acknowledge all statuses

smcveigh941:


server/@types/HdcStatus.ts
```diff
@@ -6,7 +6,13 @@ class HdcStatus {
...
   ) {
     this.bookingId = bookingId
-    this.eligibleForHdc = homeDetentionCurfewEligibilityDate && checksPassed === true && approvalStatus !== 'REJECTED'
+    this.eligibleForHdc =
+      homeDetentionCurfewEligibilityDate &&
+      checksPassed === true &&
+      approvalStatus !== 'REJECTED' &&
+      approvalStatus !== 'INELIGIBLE' &&
+      approvalStatus !== 'OPT_OUT' &&
+      approvalStatus !== 'PRES UNSUIT'
```
smcveigh941, line 15: No existing unit tests for this class, although there ought to be. This class is in the `@types` folder which doesn't feel right, so probably a bit of a refactor due to put this logic elsewhere and cover it with unit tests

timharrison-moj: Yes agreed. Its an odd mix of "data" class with a bit of logic setting a virtual field `eligible_for_hdc` in its constructor.

timharrison-moj (approved):
LGTM

PR #278, opened 2022-04-08
CVSL-500 filter rule changes

timharrison-moj:
* Removes the 18th April CRD cut-off from the case lists - we already have cases before this e.g. a release planned on 14th.
* No longer filters out empty/missing `conditionalReleaseDate` - it now allows for the `conditionalReleaseDate` or `releaseDate` or `confirmedReleaseDate` to be present - we've seen cases which should appear being filtered for this.
* Recalls unaffected by this change - there is a rule commented that we may decide to put in later.
* Tests updated - including vary caseloads with ACTIVE IN status altered to INACTIVE OUT - and still passing.

### DIFF
--- a/server/services/caseloadService.ts
+++ b/server/services/caseloadService.ts
@@ -203,13 +203,13 @@ export default class CaseloadService {
       .filter(offender => !offender.nomisRecord.paroleEligibilityDate)
       .filter(offender => offender.nomisRecord.legalStatus !== 'DEAD')
       .filter(offender => !offender.nomisRecord.indeterminateSentence)
-      .filter(offender => offender.nomisRecord.conditionalReleaseDate)
-      // TODO: Following filter rule can be removed after 18th April 2022
-      .filter(offender =>
-        moment(offender.nomisRecord.conditionalReleaseDate, 'YYYY-MM-DD').isSameOrAfter(
-          moment('2022-04-18', 'YYYY-MM-DD'),
-          'day'
-        )
+      // TODO - Check with Jon on rules for filtering recalls
+      // .filter(offender => !offender.nomisRecord.recall)
+      .filter(
+        offender =>
+          offender.nomisRecord.conditionalReleaseDate ||
+          offender.nomisRecord.releaseDate ||
+          offender.nomisRecord.confirmedReleaseDate
       )
 
     const hdcStatuses = await this.prisonerService.getHdcStatuses(
--- a/server/services/caseloadService.ts
+++ b/server/services/caseloadService.ts
@@ -211,6 +211,25 @@ export default class CaseloadService {
           offender.nomisRecord.releaseDate ||
           offender.nomisRecord.confirmedReleaseDate
       )
+      // TODO: Following filter rules can be removed after 18th April 2022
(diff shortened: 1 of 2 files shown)


server/services/caseloadService.ts
```diff
@@ -203,13 +203,13 @@ export default class CaseloadService {
...
       .filter(offender => !offender.nomisRecord.indeterminateSentence)
-      .filter(offender => offender.nomisRecord.conditionalReleaseDate)
-      // TODO: Following filter rule can be removed after 18th April 2022
-      .filter(offender =>
-        moment(offender.nomisRecord.conditionalReleaseDate, 'YYYY-MM-DD').isSameOrAfter(
-          moment('2022-04-18', 'YYYY-MM-DD'),
-          'day'
-        )
+      // TODO - Check with Jon on rules for filtering recalls
+      // .filter(offender => !offender.nomisRecord.recall)
```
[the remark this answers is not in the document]
timharrison-moj: I think one of the cases was a `recall==true` - but really it was not on the case list because it had no CRD, and instead had a confirmedReleaseDate and releaseDate.

server/services/caseloadService.ts
```diff
@@ -203,13 +203,13 @@ export default class CaseloadService {
...
+      .filter(
+        offender =>
+          offender.nomisRecord.conditionalReleaseDate ||
+          offender.nomisRecord.releaseDate ||
+          offender.nomisRecord.confirmedReleaseDate
```
smcveigh941, line 212: Can you check `licenceService` and `datesChangedEventHandler`? `confirmedReleaseDate` isnt used in these places, and maybe it should be

server/services/caseloadService.ts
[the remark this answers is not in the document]
smcveigh941: Yea the recall flag was a red herring I think. AFAIK the only things that were causing problems was the lack of CRD and the HDC filter rules

server/services/caseloadService.ts
timharrison-moj: I'll look when I get back in next hour.

timharrison-moj: It looks fine - the datesChangedEventHandler receives either a `nomisId` or a `bookingId` (which it looks up the nomisId), then gets the matching, in-progress (creation only) licences by `nomisId` and then sets the licence to in-progress, pulls the sentence dates from prison-api and updates the dates on the licence (conditionalReleaseDate and actualReleaseDate).

timharrison-moj: One improvement might be to use the `sentenceDetail.releaseDate` IF there is no `conditionalReleaseDate` present in the setence detail.

timharrison-moj: Or even if there is - and do the same thing that the licenceStartDate is set ti - use releaseDate if present, otherwise conditionalReleaseOverrideDate, and otherwise conditionalReleaseDate.

smcveigh941 (dismissed):
LGTM

smcveigh941 (approved):
LGTM

PR #281, opened 2022-04-11
CVSL-505: Fix 404 throwing for favicon during PDF render

smcveigh941:
Gotenberg automatically searches for a favicon at `/favicon.ico`


joelstobart-moj (approved):
LGTM

PR #282, opened 2022-04-11
CVSL-485: Assign the COM role on offender allocation

smcveigh941:
- Tested using delius wiremock
- Also simplified the endpoint verification task during integration tests by reusing the `success` logic in one place


timharrison-moj (approved):
LGTM

PR #283, opened 2022-04-11
CVSL-500: Caseload rules 

timharrison-moj:
* Remove the use of `releaseDate` as a driver for showing cases - this could bring in other types of licence (DTOs)
* Rely on `conditionalReleaseDate` or `confirmedReleaseDate` only
* Show recalls on case lists but make them unclickable - cannot create a licence - out of scope.
* Show BOTUS (breach of top-up supervision orders) cases on case lists but make them unclickable - cannot create a licence.
* Case list tests to confirm the above.

### DIFF
--- a/server/services/caseloadService.test.ts
+++ b/server/services/caseloadService.test.ts
@@ -139,21 +139,6 @@ describe('Caseload Service', () => {
           },
         ],
       },
-      {
-        deliusRecord: {
-          offenderCrn: 'X12352',
-        },
-        nomisRecord: {
-          prisonerNumber: 'AB1234M',
-          releaseDate: '2022-06-20',
-        },
-        licences: [
-          {
-            status: 'NOT_STARTED',
-            type: 'AP',
-          },
-        ],
-      },
     ])
   })
 
@@ -287,6 +272,7 @@ describe('Caseload Service', () => {
     ])
 
     const result = await serviceUnderTest.getTeamCreateCaseload(user)
+
     expect(communityService.getManagedOffendersByTeam).toHaveBeenNthCalledWith(1, 'teamA')
     expect(communityService.getManagedOffendersByTeam).toHaveBeenNthCalledWith(2, 'teamB')
     expect(result).toMatchObject([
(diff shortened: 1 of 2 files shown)


server/services/caseloadService.ts
```diff
@@ -79,10 +79,16 @@ export default class CaseloadService {
...
+          .filter(
+            offender =>
+              offender.conditionalReleaseDate &&
+              moment(offender.conditionalReleaseDate, 'YYYY-MM-DD').isSameOrAfter(moment(), 'day')
+          )
+          .filter(
+            offender =>
+              offender.conditionalReleaseDate &&
+              moment(offender.conditionalReleaseDate, 'YYYY-MM-DD').isBefore(moment().add(13, 'weeks'), 'day')
           )
```
smcveigh941, line 92: I know this is a temporary block... but people will still be filtered from OMU view if they dont have a CRD by these rules

smcveigh941:
The email prompt job filters by `conditionalReleaseDate` being within 13 weeks.. probably ought to also check `confirmedReleaseDate` in the job...

server/services/caseloadService.ts
smcveigh941: Maybe set a variable called date which == `offender.confirmedlReleaseDate || offender.conditionalReleaseDate` before doing the date checks

timharrison-moj: I saw this bit and did wonder - it duplicate the rules, but its to cut-down the numbers going into the next  functions temporarily.  Probably does need opening up in a similar way to the eligibility filter i.e. conditionalReleaseDate OR confirmedReleaseDate.

timharrison-moj: On it now.

server/services/caseloadService.ts
```diff
@@ -203,14 +213,7 @@ export default class CaseloadService {
...
       .filter(offender => !offender.nomisRecord.indeterminateSentence)
-      // TODO - Check with Jon on rules for filtering recalls
-      // .filter(offender => !offender.nomisRecord.recall)
-      .filter(
-        offender =>
-          offender.nomisRecord.conditionalReleaseDate ||
-          offender.nomisRecord.releaseDate ||
-          offender.nomisRecord.confirmedReleaseDate
-      )
+      .filter(offender => offender.nomisRecord.conditionalReleaseDate || offender.nomisRecord.confirmedReleaseDate)
```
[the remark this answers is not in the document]
timharrison-moj: I'll do this as part of change below.

server/services/caseloadService.test.ts
```diff
@@ -139,21 +139,6 @@ describe('Caseload Service', () => {
...
-          prisonerNumber: 'AB1234M',
-          releaseDate: '2022-06-20',
-        },
-        licences: [
-          {
-            status: 'NOT_STARTED',
-            type: 'AP',
-          },
-        ],
-      },
```
smcveigh941, line 156: Now filtered because of the recall rule :+1:

server/services/caseloadService.ts
```diff
@@ -78,12 +78,15 @@ export default class CaseloadService {
...
-          )
+        return caseload.filter(
+          offender =>
+            (offender.conditionalReleaseDate &&
+              moment(offender.conditionalReleaseDate, 'YYYY-MM-DD').isSameOrAfter(moment(), 'day') &&
+              moment(offender.conditionalReleaseDate, 'YYYY-MM-DD').isBefore(moment().add(13, 'weeks'), 'day')) ||
+            (offender.confirmedReleaseDate &&
+              moment(offender.confirmedReleaseDate, 'YYYY-MM-DD').isSameOrAfter(moment(), 'day') &&
+              moment(offender.confirmedReleaseDate, 'YYYY-MM-DD').isBefore(moment().add(13, 'weeks'), 'day'))
+        )
```
smcveigh941, line 89: Instead of having to check both dates everywhere, I am thinking of putting a `releaseDate` field on the `ManagedCase` type which is a composite of the various dates (i.e. the most relevent one). This should simplify the code a bit. Its a future improvement though

smcveigh941 (approved):
LGTM

PR #284, opened 2022-04-12
CVSL-500: Checks for not in pilot delegated to the caseload service

timharrison-moj:
This PR:
* Delegates the not in pilot checks to the caseload service - no longer re-checks the rollout prisons
* The caseload service also excluded recalls and BOTUS type cases - returns these as NOT_IN_PILOT too
* Updated tests with an example of NOT_IN_PILOT for caseload checks
* Removed route tests which were re-checking the NOT_IN_PILOT for rollout areas/prisons - so now trusts the caseload service.
* Removed the `recall: true` flag from integration test mocked endpoints - causing links to be unclickable.


server/routes/creatingLicences/handlers/caseload.test.ts
```diff
@@ -225,73 +251,13 @@ describe('Route Handlers - Create Licence - Caseload', () => {
...
-        teamView: true,
-        search: 'rogan',
-      })
-      expect(caseloadService.getTeamCreateCaseload).toHaveBeenCalledWith(res.locals.user)
-      expect(caseloadService.getStaffCreateCaseload).not.toHaveBeenCalled()
-    })
-
-    it('should identify offenders outside the pilot areas for the single officer view', async () => {
-      config.rollout.restricted = true
-      config.rollout.prisons = ['LEI']
```
timharrison-moj, line 257: Removing the tests which check for cases outside the pilot areas - these checks are already done in the caseload service, so now trusting the status returned for NOT_IN_PILOT - they could be either location restrictions, or because they are recalls or BOTUS type cases.

smcveigh941 (approved):
LGTM

PR #286, opened 2022-04-19
CVSL-512: Updated the recall rules 

timharrison-moj:
This change alters a few things:

* Eligibility for the service now requires the presence of a `conditionalReleaseDate`  (`confirmedReleaseDate` is no good now)
* The Nomis recall flag is sometimes overridden where the `postRecallReleaseDate` is <= CRD - and makes cases `Not started`
* Breach of topup supervision cases are now not clickable - though show on case lists as `Breach of supervision`
* Valid recalls also show on case lists as `Recall`
* The about page has some new words around recalls.
* Tests for eligibility, case lists, recall and BOTUS cases.


server/services/caseloadService.ts
```diff
@@ -381,7 +381,27 @@ export default class CaseloadService {
...
+    const postRecallReleaseDate = offender.nomisRecord?.postRecallReleaseDate
+      ? moment(offender.nomisRecord?.postRecallReleaseDate, 'YYYY-MM-DD')
+      : undefined
+
+    // If the post-recall release date is AFTER the release date, then it is a genuine recall.
+    // If the post-recall release date is BEFORE or EQUAL to the release date - we ignore the recall flag.
+    // This catches the situation where multiple sentences exist and the recall applies to a previous sentence.
+    if (recall && releaseDate && postRecallReleaseDate) {
+      return postRecallReleaseDate.isAfter(releaseDate)
+    }
```
smcveigh941, line 401: Still a question about when the PRRD is EQUAL to the CRD. This implementation would bring single-sentence recalls into scope of CVL.

timharrison-moj: It would. Jon is having a think about this and talking to Louise. Trying to figure out if there's any other piece of information we could check to signify that there is only one sentence and that the PRRD applies to the same sentence as the release date
.

timharrison-moj: The rules are written out on this ticket and now implemented here - https://dsdmoj.atlassian.net/browse/CVSL-518

server/licences/licenceStatus.ts
```diff
@@ -72,6 +72,16 @@ const statusConfig: Record<LicenceStatus, LicenceStatusConfig> = {
     description: 'Outside pilot area',
     colour: 'grey',
   },
+  OOS_BOTUS: {
+    label: 'Breach of supervision',
```
smcveigh941, line 76: Is this text too long for the badge? How does it look?

server/services/caseloadService.ts
```diff
@@ -197,7 +204,7 @@ export default class CaseloadService {
       .filter(offender => !offender.nomisRecord.paroleEligibilityDate)
       .filter(offender => offender.nomisRecord.legalStatus !== 'DEAD')
       .filter(offender => !offender.nomisRecord.indeterminateSentence)
-      .filter(offender => offender.nomisRecord.confirmedReleaseDate || offender.nomisRecord.conditionalReleaseDate)
```
smcveigh941, line 200: Why remove this? There are legitimate cases which do not have a CRD which would not show up?

server/views/pages/about.njk
```diff
@@ -63,6 +63,12 @@
                 <li>licences requiring parole board approval (including extended determinate sentences and special custodial sentences)</li>
             </ul>
 
+            <p class="govuk-body"><strong>Please note:</strong> Where somebody has been recalled to prison and stays in custody beyond the
```
smcveigh941, line 66:
```suggestion
            <p class="govuk-body"><span class="govuk-!-font-weight-bold">Please note:</span> Where somebody has been recalled to prison and stays in custody beyond the
```

server/views/pages/about.njk
```diff
@@ -77,6 +83,16 @@
 
             <hr class="govuk-section-break govuk-section-break--visible govuk-!-margin-bottom-3"/>
 
+            <h1 class="govuk-heading-m">Licences for a person both recalled and serving another sentence</h1>
```
smcveigh941, line 86: Not sure, but I think multiple `H1` on a page causes an accessibility error?

timharrison-moj: Thanks - not noticed these, should be h2's I think. Will change now.

server/services/caseloadService.ts
timharrison-moj: This was the outcome of many discussions - as we covered at stand-up.

server/licences/licenceStatus.ts
timharrison-moj: It is 21 chars, same as `Variation in progress`.

server/views/pages/about.njk
```diff
@@ -72,18 +72,28 @@
...
 
             <hr class="govuk-section-break govuk-section-break--visible govuk-!-margin-bottom-3"/>
 
-            <h1 class="govuk-heading-m">Getting new licences onto nDelius and NOMIS</h1>
+            <h2 class="govuk-heading-m">Licences for a person both recalled and serving another sentence</h2>
+            <p class="govuk-body">
+                Use this service for a release when someone has been recalled to prison and stays beyond the recall
+                period because they are also serving a sentence for a new offence. This is not classed as a re-release
+                after recall, but is a standard release from the new sentence. This means you will see them in your
+                Create and vary a licence case list..
```
smcveigh941, line 85: Typo here of `..`?

timharrison-moj: Thanks. Fixed now.

smcveigh941 (approved):
LGTM

PR #287, opened 2022-04-19
CVSL-513: Search by release date

smcveigh941:


jobs/promptLicenceCreation.ts
```diff
@@ -1,6 +1,6 @@
 import 'reflect-metadata'
-import moment from 'moment'
 import _ from 'lodash'
+import moment, { Moment } from 'moment'
```
yasinmustafa, line 3: minor one and not specifically related to this PR, but moment is now a legacy library and their advice is to use one of the other JS date libraries

https://momentjs.com/docs/#/-project-status/

yasinmustafa (approved):
LGTM - one general comment about perhaps considering using another date library instead of moment

PR #295, opened 2022-04-26
CVSL-531: Avoid updating licence status for dates-changed events

timharrison-moj:
This is part 1 of a 2-port change. Part 2 is in the API coming soon.

This change avoids updating the licence status whenever we get a dates-changed event.
It will be replaced by a notification to the COM (via the API) if any of the date changes are `material` to the licence.
This avoids the situation where licences are changed from APPROVED to IN_PROGRESS just prior to release, as it looks like the OMU staff update the release date near, or on, the day of release.


smcveigh941 (approved):
LGTM

PR #296, opened 2022-04-28
refs CVSL-540

mgreen-mojhub:
Add the postgresql container to docker-compose-dev.yaml - so will run this and other required containers when using DEV for auth, prison, community APIs etc.

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -69,7 +69,7 @@ services:
     container_name: licences-db
     restart: always
     ports:
-      - "5433:5432"
+      - "5432:5432"
     environment:
       - POSTGRES_PASSWORD=create-and-vary-a-licence
       - POSTGRES_USER=create-and-vary-a-licence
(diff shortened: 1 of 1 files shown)


docker-compose-dev.yml
```diff
@@ -61,5 +61,18 @@ services:
...
+  # Optional - only when running API locally
+  licences-db:
+    image: postgres
+    networks:
+      - hmpps
+    container_name: licences-db
+    restart: always
+    ports:
+      - "5433:5432"
+    environment:
```
timharrison-moj, line 73: The port might be an issue - as the API expects it on 5432. Is this because you also have another postgres locally? Also, 5433 is used by other devs to connect to remote K8s db's via port-forwards. Any change we could change to default 5432 here and stop/start other competing pgSQLs please?

timharrison-moj (commented):
Looks fine apart from the port - other things expect it on localhost:5432 when running in docker I'm afraid.

docker-compose-dev.yml
mgreen-mojhub: Yes this makes sense, it should be the default port. I'll make a push now.

timharrison-moj (approved):
LGTM

PR #297, opened 2022-04-28
CVSL-542: Increase the polling interval for SQS message consumer to 1 minute - allowing time for elastic search indexes to index themselves

smcveigh941:


timharrison-moj (dismissed):
LGTM. Only thought is that 5 seconds to 5 mins is a big jump? Santhosh's testing times in DEV will go up quite a lot. What do you think about 40-50 seconds instead as a first try?

timharrison-moj (approved):
LGTM

PR #298, opened 2022-04-29
CVSL-545: Correct the audit event time format when raised from the UI service.

timharrison-moj:
This bug affects audit events raised by the UI service only.
The moment time parsing was incorrectly using the 12-hour clock form `hh:mm:ss`.
It should (and now does) use the 24-hour clock format `HH:mm:ss` + tests updated.

### DIFF
--- a/server/services/licenceService.ts
+++ b/server/services/licenceService.ts
@@ -388,8 +388,8 @@ export default class LicenceService {
   ): Promise<void> {
     const requestBody = {
       username: user.username,
-      eventTime: moment(eventTime).format('DD/MM/YYYY hh:mm:ss'),
-      eventType: user ? 'USER_EVENT' : 'SYSTEM_EVENT',
+      eventTime: moment(eventTime).format('DD/MM/YYYY HH:mm:ss'),
+      eventType: user && user?.username !== 'SYSTEM' ? 'USER_EVENT' : 'SYSTEM_EVENT',
       licenceId,
       fullName: `${user.firstName} ${user.lastName}`,
       summary,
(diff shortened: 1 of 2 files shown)


server/services/licenceService.ts
```diff
@@ -388,8 +388,8 @@ export default class LicenceService {
   ): Promise<void> {
     const requestBody = {
       username: user.username,
-      eventTime: moment(eventTime).format('DD/MM/YYYY hh:mm:ss'),
-      eventType: user ? 'USER_EVENT' : 'SYSTEM_EVENT',
+      eventTime: moment(eventTime).format('DD/MM/YYYY HH:mm:ss'),
+      eventType: user && user?.username !== 'SYSTEM' ? 'USER_EVENT' : 'SYSTEM_EVENT',
```
smcveigh941, line 392: When is the username 'SYSTEM'?

smcveigh941: I don't think this is ever the case?

smcveigh941 (dismissed):
LGTM with a question

server/services/licenceService.ts
timharrison-moj: You're right. When called from the UI service we always pass the current `user` which will set the event type to USER_EVENT. All other auditing is done inside the API. I'll remove this extra check.

timharrison-moj: I think it was the update status after a date change (now removed) that might have been why I added this - because we used to record the edited licence as a USER EVENT there.

timharrison-moj: Pushed again now.

smcveigh941 (approved):
LGTM

PR #301, opened 2022-05-03
CVSL-541: Update the probation team on the OFFENDER_MANAGER_CHANGED event

smcveigh941:


server/listeners/eventHandlers/probationEvents/offenderManagerChangedEventHandler.ts
```diff
@@ -26,13 +26,25 @@ export default class OffenderManagerChangedEventHandler {
...
+            probationAreaCode: newCom.probationArea?.code,
+            probationAreaDescription: newCom.probationArea?.description,
+            probationPduCode: newCom.team?.borough?.code,
+            probationPduDescription: newCom.team?.borough?.description,
+            probationLauCode: newCom.team?.district?.code,
+            probationLauDescription: newCom.team?.district?.description,
+            probationTeamCode: newCom.team?.code,
+            probationTeamDescription: newCom.team?.description,
+          }),
+        ])
```
timharrison-moj, line 47: These are very likely to update the same row, so just a thought about concurrent update? It's only a thought - as I'm certain that JPA will deal with it as separate transactions and all will be fine - just a thought. Two sequential awaits would take the question away, but I'm 97% certain it'll be fine as is.

timharrison-moj (approved):
LGTM.

timharrison-moj (approved):
LGTM

PR #310, opened 2022-05-09
CVAL 105 audit trail for licence

mgreen-mojhub:


server/routes/support/handlers/offenderAudit.ts
```diff
@@ -0,0 +1,32 @@
...
+  GET = async (req: Request, res: Response): Promise<void> => {
+    const { user } = res.locals
+    const { nomsId, licenceId } = req.params
+
+    const prisonerDetail = _.head(await this.prisonerService.searchPrisonersByNomisIds([nomsId], user))
+    const audit = await this.licenceServer.getAuditEvents(
+      parseInt(licenceId, 10),
+      null,
+      moment().subtract(2, 'years').toDate(),
+      moment().toDate(),
```
timharrison-moj, line 20: Not one for now, but could be made into a configuration value for how many years of audit information is returned.

timharrison-moj (approved):
Very good, Matt. No issues at all with that. Only comment was one for the future that we can configure how wide the audit period is in months/years.

PR #313, opened 2022-05-10
CVSL-243: Fix actualReleaseDate field when confirmedReleaseDate is null on the confirmCreatePage

smcveigh941:
- When confirmedReleaseDate was null, the original code was setting actualReleaseDate == todays date


timharrison-moj (approved):
LGTM

PR #314, opened 2022-05-13
CVSL 530 hdc filter logic

mgreen-mojhub:


integration_tests/integration/createLicence.spec.ts
```diff
@@ -31,12 +31,16 @@ context('Create a licence', () => {
...
 
-    const additionalConditionsQuestionPage = appointmentTimePage
-      .enterDate(moment().add(1, 'year'))
+    /*
+     select the next wednesday one year from now in order to avoid weekends and
+     bank holidays, with the possible exceptions of Xmas and New Year
+     */
+    const additionalConditionsPage = appointmentTimePage
+      .enterDate(moment().add(1, 'year').add(1, 'week').day(3))
       .enterTime(moment())
```
timharrison-moj, line 40: This will fail if Xmas, Boxing Day or NYD falls on a Weds - if the test runs in mid December! 
It might need to detect if the month is December, and if so, add a month? OK for now though, just ideas.

mgreen-mojhub: Good idea! We can simply skip to January to avoid this issue. I'll push the change once complete.

server/services/prisonerService.ts
```diff
@@ -68,16 +68,18 @@ export default class PrisonerService {
...
-      return new HdcStatus(`${h?.bookingId}`, hdcEligibilityDate, h?.passed, h?.approvalStatus)
-    })
+
+    return hdcList.map(
+      h =>
+        <HdcStatus>{
+          bookingId: h.bookingId.toString(),
+          approvalStatus: h?.approvalStatus || 'Not found',
+          checksPassed: h?.passed || 'Not found',
+        }
```
timharrison-moj, line 81: checksPassed a boolean value? Should it be set to 'Not found' or just false?

timharrison-moj (dismissed):
LGTM. Minor question only.

timharrison-moj (approved):
LGTM

PR #318, opened 2022-05-29
CVSL-575

mgreen-mojhub:
Remove IN_PROGRESS licences when checking for licences by date. The standard list of licences for an OMU caseload now contains IN_PROGRESS records, so should not be returned again using the date range.


GurnankCheema (approved):
LGTM

PR #332, opened 2022-07-14
CVSL 502 select caseload approver

GurnankCheema:
Approver can change their caseload and view cases relevant to those prisons.


server/routes/approvingLicences/handlers/approvalCases.ts
```diff
@@ -2,16 +2,23 @@ import { Request, Response } from 'express'
...
   GET = async (req: Request, res: Response): Promise<void> => {
     const search = req.query.search as string
 
     const { user } = res.locals
-    const cases = await this.caseloadService.getApproverCaseload(user)
+    const { caseloadsSelected = [] } = req.session
+    const hasMultipleCaseloadsInNomis = user.prisonCaseload.length > 1
+    const allPrisons = await this.prisonerService.getPrisons()
+    const activeCaseload = allPrisons.filter(p => p.agencyId === user.activeCaseload)
+    const prisonCaseloadToDisplay = caseloadsSelected.length ? caseloadsSelected : [activeCaseload[0].agencyId]
```
andrewrlee, line 19: This is a lot of stuff to be doing in a controller and presumably you have similar stuff in the CA flow.

do you have a userservice you could push a lot of this down to?  getSelectedPrisons or something?

server/routes/approvingLicences/handlers/approvalCases.ts
```diff
@@ -2,16 +2,23 @@ import { Request, Response } from 'express'
...
+  constructor(private readonly caseloadService: CaseloadService, private readonly prisonerService: PrisonerService) {}
 
   GET = async (req: Request, res: Response): Promise<void> => {
     const search = req.query.search as string
 
     const { user } = res.locals
-    const cases = await this.caseloadService.getApproverCaseload(user)
+    const { caseloadsSelected = [] } = req.session
+    const hasMultipleCaseloadsInNomis = user.prisonCaseload.length > 1
+    const allPrisons = await this.prisonerService.getPrisons()
```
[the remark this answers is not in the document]
GurnankCheema: ah yes.
The reason I created the getPrisons method  in prisonerService is the service uses the prisonApiClient. 
There is already another method in there, getPrisonInformation, and I felt the two are closely related so grouped them together.

server/routes/approvingLicences/handlers/approvalCases.ts
andrewrlee: Actually it might be good to enrich the user and their caseloads with prison info in the `populateCurrentUser` middleware.
It already called prison api and it only does it once per session rather than on every request.

mgreen-mojhub (dismissed):

mgreen-mojhub (approved):

PR #335, opened 2022-07-20
Cvsl 610 update std conditions on save

mgreen-mojhub:
If the current policy version is different to the that of the current licence, force and update to the standard conditions so they always appear in the PDF. This functionality can be superseded once the proper policy versioning work is ready.


GurnankCheema (approved):
LGTM

PR #336, opened 2022-07-21
CVCL-582 case list page

GurnankCheema:
COM vary journey

1. Change column header ‘Status’ to ‘Variation status’
2. No variation in progress – No badge/blank . This will be when status is Active. 
3. Attention needed badge – change colour to red. Change text too. The VARIATION_REJECTED status displays the Attention needed.

### DIFF
--- a/server/routes/varyingLicences/handlers/caseload.ts
+++ b/server/routes/varyingLicences/handlers/caseload.ts
@@ -29,7 +29,7 @@ export default class CaseloadRoutes {
           crnNumber: c.deliusRecord.otherIds.crn,
           licenceType: licence.type,
           releaseDate: moment(c.nomisRecord.releaseDate, 'YYYY-MM-DD').format('DD MMM YYYY'),
-          licenceStatus: licence.status === LicenceStatus.ACTIVE ? null : licence.status,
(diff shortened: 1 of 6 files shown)


server/routes/varyingLicences/handlers/caseload.ts
```diff
@@ -29,7 +29,7 @@ export default class CaseloadRoutes {
           crnNumber: c.deliusRecord.otherIds.crn,
           licenceType: licence.type,
           releaseDate: moment(c.nomisRecord.releaseDate, 'YYYY-MM-DD').format('DD MMM YYYY'),
-          licenceStatus: licence.status,
+          licenceStatus: licence.status === LicenceStatus.ACTIVE ? null : licence.status,
```
GurnankCheema, line 32: Setting to null will prevent the statusBadge macro rendering the badge

mgreen-mojhub: Could this check be placed within the Macro itself? The only issue I can see here is if something now or in the future references the licenceStatus property, downstream from the handler.

mgreen-mojhub (approved):
LGTM

PR #339, opened 2022-07-25
CVSL-582 error, question and button text changed.

GurnankCheema:
There is no change to the underlying logic.
The page title could not be presented on one line, even with full width column. 
The image in the ticket shows the title on one line but at the expense of the content below the grey line not being aligned with the other content in the page. 

The revised page: 
Screenshot 2022-07-25 at 08 58 37


mgreen-mojhub (approved):
LGTM

PR #355, opened 2022-08-02
CVSL-607 prevent user from adding more than 2 digits 

GurnankCheema:
in any hours and minute input fields

This applies to the time fields within all pages
eg /initial-meeting-time , /additional-licence-conditions


michalnawrockidj (approved):
LGTM

PR #362, opened 2022-08-24
CVSL-442 support user can get licence stats

GurnankCheema:


server/data/licenceApiClient.ts
```diff
@@ -79,6 +80,17 @@ export default class LicenceApiClient extends RestClient {
...
   }
 
+  async getlicenceStatistics(startDate: string, endDate: string, user: User): Promise<LicenceStatistics[]> {
+    try {
+      return (await this.get(
+        { path: `/support/licence-statistics?startDate=${startDate}&endDate=${endDate}` },
+        { username: user.username }
+      )) as Promise<LicenceStatistics[]>
+    } catch (error) {
+      return error.status >= 400 && error.status < 500 ? null : error
```
andrewrlee, line 90: I don't think there are any expected errors for this endpoint so don't need to try/catch

The other calls wrap as absent OMU is modelled as a 404.

server/routes/support/index.ts
```diff
@@ -11,7 +11,9 @@ import validationMiddleware from '../../middleware/validationMiddleware'
 import prisonIdCurrent from './types/prisonIdCurrent'
 import prisonIdAndEmail from './types/prisonIdAndEmail'
 import prisonIdDelete from './types/prisonIdDelete'
+import dateRange from './types/dateRange'
```
andrewrlee, line 14: This is all a bit weird.
These are references to the classes (types), so should all be upper cased.

server/routes/support/types/customValidators.ts
```diff
@@ -0,0 +1,58 @@
...
+function isStartBeforeEnd(startDate: string, endDate: string) {
+  const parsedStartDate = startDate.split('/')
+  const parsedEndDate = endDate.split('/')
+  const StartDateDate = parseInt(parsedStartDate[0], 10)
+  const StartDateMonth = parseInt(parsedStartDate[1], 10)
+  const StartDateYear = parseInt(parsedStartDate[2], 10)
+
+  const EndDateDate = parseInt(parsedEndDate[0], 10)
+  const EndDateMonth = parseInt(parsedEndDate[1], 10)
+  const EndDateYear = parseInt(parsedEndDate[2], 10)
```
andrewrlee, line 53: Is there a utility method in date-fns to parse the date?  

Will this always be called or only if `IsDateFormat` is true?

server/services/licenceService.ts
```diff
@@ -151,6 +151,11 @@ export default class LicenceService {
     return this.licenceApiClient.createLicence(licence, user)
   }
 
+  async getlicenceStatistics(startDate: string, endDate: string, user: User) {
+    const data = await this.licenceApiClient.getlicenceStatistics(startDate, endDate, user)
```
andrewrlee, line 155: Are you converting from `d/mm/yyyy` -> `yyyy-mm-dd` somewhere?

GurnankCheema: yes

server/data/licenceApiClient.ts
GurnankCheema: ok

server/routes/support/index.ts
GurnankCheema: ok

server/routes/support/types/customValidators.ts
GurnankCheema: I think I need to revisit this

PR #364, opened 2022-08-30
CVSL-608 check PED is set and in the future

mgreen-mojhub:
check PED is set and in the future before excluding record from the caseload service.


andrewrlee (approved):
LGTM

PR #370, opened 2022-09-21
CVSL-657 - Adds temporary notice of updating policy framework version

SebNegahbanMoJ:
Adds a notice explaining that the policy framework is currently being updated, to prevent unnecessary support queries.
This should be removed once the in-flight versioning work has been completed.

image


mgreen-mojhub (approved):
LGTM

PR #375, opened 2022-10-04
CVSL-619 refactor cron job exception handling

GurnankCheema:
moved exception handling to parent function and added the missing 'await'


andrewrlee (approved):
LGTM

PR #376, opened 2022-10-06
CVSL-674 Updating node, orbs and moving alert/release channels

andrewrlee:


louisenorris (dismissed):
LGTM

.circleci/config.yml
```diff
@@ -1,46 +1,36 @@
...
 parameters:
   alerts-slack-channel:
     type: string
-    default: farsight-alerts
+    default: cvl_alerts_security
 
   releases-slack-channel:
     type: string
-    default: farsight-alerts
+    default: the_elephants_devs
```
andrewrlee, line 15: bit torn about this, not a big fan of alerts going into channels that non-robots use

mgreen-mojhub (approved):
LGTM

PR #377, opened 2022-10-06
CVSL-674 Removing dotenv as a prod dependency and using new app insights env var

andrewrlee:


package.json
```diff
@@ -10,7 +10,7 @@
     "compile-sass": "./bin/build-css.sh -q --quiet-deps --no-source-map",
     "watch-ts": "tsc -w",
     "watch-views": "nodemon --watch server/views -e html,njk -x npm run copy-views",
-    "watch-node": "DEBUG=gov-starter-server* nodemon --watch dist/ dist/server.js | bunyan -o short",
+    "watch-node": "DEBUG=gov-starter-server* nodemon -r dotenv/config --watch dist/ dist/server.js | bunyan -o short",
```
andrewrlee, line 13: I didn't add dotenv/config preload to `watch-node-feature` as all the env vars should be present in features.env. 

This should help if a dev adds new integration tests that inadvertently rely on local config. Previously this would only be detected when pushed to CI

helm_deploy/create-and-vary-a-licence/values.yaml
```diff
@@ -27,6 +27,8 @@ generic-service:
     REDIS_TLS_ENABLED: "true"
     TOKEN_VERIFICATION_ENABLED: "true"
     LICENCE_WATERMARK: "false"
+    APPLICATIONINSIGHTS_CONNECTION_STRING: "InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY);IngestionEndpoint=https://northeurope-0.in.applicationinsights.azure.com/;LiveEndpoint=https://northeurope.livediagnostics.monitor.azure.com/"
```
andrewrlee, line 30: App insights wants users to move to the connection string approach rather than use raw tokens

package.json
```diff
@@ -161,6 +160,7 @@
     "cypress": "^8.7.0",
     "cypress-axe": "^0.14.0",
     "cypress-multi-reporters": "^1.5.0",
+    "dotenv": "^16.0.3",
```
andrewrlee, line 163: now a dev dependency

server/utils/azureAppInsights.ts
```diff
@@ -16,9 +15,7 @@ function version(): string {
 }
 
 export function initialiseAppInsights(): void {
-  // Loads .env file contents into | process.env
-  config()
-  if (process.env.APPINSIGHTS_INSTRUMENTATIONKEY) {
+  if (process.env.APPLICATIONINSIGHTS_CONNECTION_STRING) {
```
andrewrlee, line 18: Moving test to use new env var

mgreen-mojhub (approved):
LGTM

PR #379, opened 2022-10-06
CVSL-674 try isolated modules

andrewrlee:


andrewrlee:
makes no difference switching on isolated modules for cypress tests to test run time

PR #384, opened 2022-10-12
Cvsl 348 multiple exclusion zones

mgreen-mojhub:


server/data/licenceApiClient.ts
```diff
@@ -130,6 +133,25 @@ export default class LicenceApiClient extends RestClient {
     )
   }
 
+  async addAdditionalCondition(
+    licenceId: string,
+    type: LicenceType,
+    additionalCondition: AddAdditionalConditionRequest,
+    user: User
+  ): Promise<AdditionalCondition> {
+    return (await this.post(
```
andrewrlee, line 142: just wondering if there was a practical reason for awaiting here but not on the delete?

server/routes/creatingLicences/handlers/additionalLicenceConditionUploadsHandler.ts
```diff
@@ -0,0 +1,57 @@
...
+    const sequence = currentOrNextSequenceForCondition(licence.additionalLicenceConditions, conditionCode)
+
+    const request = {
+      conditionCode,
+      conditionCategory: condition?.categoryShort || condition?.category,
+      conditionText: condition.text,
+      conditionType: type,
+      expandedText: condition.tpl,
+      sequence,
+    } as AddAdditionalConditionRequest
```
andrewrlee, line 49: meh.. I think this `as` stuff is a bit like casting, you're telling the compiler it's different from what it's able to infer. 

Might be better to ` const request: AddAdditionalConditionRequest = ` to ensure it matches the type, unless there's a legitimate reason to!

server/routes/creatingLicences/handlers/additionalLicenceConditionUploadsHandler.ts
```diff
@@ -0,0 +1,57 @@
...
+    const { uploadFile, conditionCode } = req.body
+
+    const condition = getAdditionalConditionByCode(conditionCode)
+    const type = getAdditionalConditionType(conditionCode)
+
+    if (uploadFile !== 'yes') {
+      return res.redirect(`/licence/create/id/${licenceId}/check-your-answers`)
+    }
+
+    const sequence = currentOrNextSequenceForCondition(licence.additionalLicenceConditions, conditionCode)
```
andrewrlee, line 40: I'd tend to shove stuff like this and the calls to things like  the `getAdditionalConditionByCode`calls into the service.
Bit of a personal preference around having very thin controllers which are solely responsible with binding to requests and creating responses.

server/views/pages/create/additionalLicenceConditionInput.njk
```diff
@@ -19,49 +19,59 @@
...
-    </form>
+    {% if fileUpload == false %}
+        <form action="{{ "/licence/create/id/" + licence.id + "/additional-licence-conditions/condition/" + additionalCondition.id + "/delete" + fromReview }}"
+              method="POST">
+            <input type="hidden" name="_csrf" value="{{ csrfToken }}">
+            {{ govukButton({
+                text: "Delete this condition",
+                type: "submit",
+                classes: 'govuk-button--secondary',
+                attributes: { 'data-qa': 'delete' }
```
andrewrlee, line 73: Might be worth generally adding `preventDoubleClick` attributes to buttons that do stuff?

andrewrlee (approved):
LGTM

PR #385, opened 2022-10-17
CVSL-645 Exposing exclusion information

andrewrlee:
Added a new wrapper object that acts like an array but `filter` takes an additional parameter as an exclusion reason. 
Items that are filtered out are kept in a separate array internally and passed along through maps, filters and concats so they can be inspected at the end of the pipeline. 

This is then exposed on a `/licence/view/cases-with-exclusions.json` url, which just dumps out the cases, cases that were excluded and their reason for exclusion.

All the caseload methods have been altered to use this structure as they share a lot of the same functionality but only currently exposed this for the OMU caseload

### DIFF
--- a/server/routes/viewingLicences/handlers/viewCases.ts
+++ b/server/routes/viewingLicences/handlers/viewCases.ts
@@ -25,6 +25,7 @@ export default class ViewAndPrintCaseRoutes {
 
     const cases = await this.caseloadService.getOmuCaseload(user, prisonCaseloadToDisplay, view as string)
     const caseloadViewModel = cases
+      .unwrap()
       .map(c => {
         return {
           licenceId: _.head(c.licences).id,
@@ -70,6 +71,19 @@ export default class ViewAndPrintCaseRoutes {
       hasMultipleCaseloadsInNomis,
       probationView,
     })
+  }
+
+  GET_WITH_EXCLUSIONS = async (req: Request, res: Response): Promise<void> => {
+    const { user } = res.locals
+    const { caseloadsSelected = [] } = req.session
+    const view = req.query.view || 'prison'
+    const allPrisons = await this.prisonerService.getPrisons()
+    const activeCaseload = allPrisons.filter(p => p.agencyId === user.activeCaseload)
+    const prisonCaseloadToDisplay = caseloadsSelected.length ? caseloadsSelected : [activeCaseload[0].agencyId]
+
+    const cases = await this.caseloadService.getOmuCaseload(user, prisonCaseloadToDisplay, view.toString())
+    res.header('Content-Type', 'application/json')
+    res.send(JSON.stringify(cases, null, 4))
   }
 
   selectReleaseDate(nomisRecord: Prisoner) {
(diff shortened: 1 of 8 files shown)


server/services/container.ts
```diff
@@ -0,0 +1,40 @@
+type FilterParams<T> = (value: T, index: number, array: T[]) => unknown
+type MapParams<T, R> = (t: T, index: number, array: T[]) => R
+type ExcludedItem = [item: unknown, reason: string]
+
+export default class Container<T> {
```
andrewrlee, line 5: Didn't really know what to call this.. `Caseload` was tempting but it can contain anything really.
It's used quite a few places so wanted to call it something short... rather than `ExclusionRecordingList` or something

server/services/container.ts
```diff
@@ -0,0 +1,40 @@
+type FilterParams<T> = (value: T, index: number, array: T[]) => unknown
+type MapParams<T, R> = (t: T, index: number, array: T[]) => R
+type ExcludedItem = [item: unknown, reason: string]
```
andrewrlee, line 3: `item` is of unknown type as if you have a pipeline of operations, different steps could produce excluded items in different structures.

server/services/caseloadService.ts
```diff
@@ -139,14 +143,17 @@ export default class CaseloadService {
             },
           }
         }
-        return {}
+        return { nomisRecord: offender }
```
andrewrlee, line 146: this is thrown away after the subsequent filter but passing it through so we store some additional contextual information along with exclusion

server/services/caseloadService.ts
```diff
@@ -79,24 +80,26 @@ export default class CaseloadService {
...
+            ].some(status => c.licences.find(l => l.status === status)),
+          'Has no licence in NOT_STARTED, NOT_IN_PILOT, OOS_RECALL, OOS_BOTUS'
         )
       })
 
-    const combinedCases = await Promise.all([casesWithLicences, casesPendingLicence]).then(c => c.flat())
-
-    return this.mapResponsibleComsToCases(combinedCases)
+    const [withLicence, pending] = await Promise.all([casesWithLicences, casesPendingLicence])
+    return this.mapResponsibleComsToCasesWithExclusions(withLicence.concat(pending))
```
andrewrlee, line 102: `flat` doesn't work with none-array types so reimplemented this with concat

server/routes/viewingLicences/handlers/viewCases.ts
```diff
@@ -67,6 +68,18 @@ export default class ViewAndPrintCaseRoutes {
...
+  GET_WITH_EXCLUSIONS = async (req: Request, res: Response): Promise<void> => {
+    const { user } = res.locals
+    const { caseloadsSelected = [] } = req.session
+    const allPrisons = await this.prisonerService.getPrisons()
+    const activeCaseload = allPrisons.filter(p => p.agencyId === user.activeCaseload)
+    const prisonCaseloadToDisplay = caseloadsSelected.length ? caseloadsSelected : [activeCaseload[0].agencyId]
+
+    const cases = await this.caseloadService.getOmuCaseload(user, prisonCaseloadToDisplay)
+    res.header('Content-Type', 'application/json')
+    res.send(JSON.stringify(cases, null, 4))
```
andrewrlee, line 80: just dumping it out as json - might be a bit tricky rendering as it will currently produce objects at varying stages so will have different structures

GurnankCheema (dismissed):
LGTM

mgreen-mojhub (approved):
Looks good

PR #386, opened 2022-10-18
Cvsl 138 for andy

andrewrlee:


server/routes/varyingLicences/handlers/policyChange.ts
```diff
@@ -0,0 +1,184 @@
...
+    const conditionHintText = policyChangeHintText[condition.code][this.formatVersionNumber('2.1')]
+
+    // Overriding session changeCounter for when the back link is clicked
+    req.session.changedConditionsCounter = conditionCounter
+
+    if (conditionCounter === 0) {
+      return res.redirect(`/licence/vary/id/${licenceId}/policy-changes`)
+    }
+
+    if (!condition.currentText) {
```
andrewrlee, line 27: @SebNegahbanMoJ  
Does this mean: if the condition is not present in the new conditions and all of the replacements existed in the previous set, then consider the condition deleted?

SebNegahbanMoJ: I've put more info in Slack, but for an explanation here too:
This check determines whether a condition has been `deleted` or `replaced`.
`deleted` and `replaced` are largely the same behaviour (a condition has been deleted and new conditions have been suggested to replace it), but the distinction lies in the content of the pages. Either way, the previously selected condition no longer exists.
`deleted` conditions are those whose replacements contain one or more existing conditions, in other words, the replacements are suggestions of existing conditions that may be appropriate, or may not.
`replaced` conditions are those whose replacements are all new conditions, in other words, the replacements have been added as explicit replacements for the deleted condition.


The check that determines whether a condition has been removed from policy is the `!condition.currentText` on line 27, as it was assumed that if the description of the change happening to a condition in the policy resulted in there being no new text, the condition had been removed.

PR #388, opened 2022-10-19
CVSL-348-UR-feedback

mgreen-mojhub:


server/routes/creatingLicences/handlers/additionalLicenceConditionUploadsHandler.ts
```diff
@@ -16,13 +17,13 @@ export default class AdditionalLicenceConditionUploadsHandler {
     const { user } = res.locals
     const licence = await this.licenceService.getLicence(licenceId, user)
 
-    const conditions = licence.additionalLicenceConditions.filter(c => c.code === conditionCode)
+    const conditions = licence.additionalLicenceConditions.filter((c: AdditionalCondition) => c.code === conditionCode)
```
andrewrlee, line 20: is this just a stylistic thing? typescript should be able to infer the types anyway?

server/routes/creatingLicences/handlers/additionalLicenceConditionInput.ts
```diff
@@ -61,12 +61,20 @@ export default class AdditionalLicenceConditionInputRoutes {
     const { conditionId } = req.params
     const { user } = res.locals
 
+    const condition = licence.additionalLicenceConditions.find(
+      (c: AdditionalCondition) => c.id === parseInt(conditionId, 10)
+    ) as AdditionalCondition
+
+    // if the exclusion zone name outOfBoundArea is present, redirect to file uploads dialog
+    const redirect =
+      condition.expandedText.indexOf('{outOfBoundsArea}') > 1
```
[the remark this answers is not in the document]
mgreen-mojhub: Its horrid, but the only way I could think of to redirect to the file upload summary page if the condition has the corresponding placeholder. I'll try and think of another way.

server/routes/creatingLicences/handlers/additionalLicenceConditionUploadsHandler.ts
```diff
@@ -54,4 +63,11 @@ export default class AdditionalLicenceConditionUploadsHandler {
       `/licence/create/id/${licenceId}/additional-licence-conditions/condition/${conditionResult.id}?fromReview=true`
     )
   }
+
+  DELETE = async (req: Request, res: Response): Promise<void> => {
```
andrewrlee, line 67: missing test coverage for these as well

server/routes/creatingLicences/handlers/additionalLicenceConditionUploadsHandler.ts
mgreen-mojhub: I was getting type errors within IntelliJ when missing the cast. It seems at bit fickle within the IDE to be fair.

server/views/formBuilder.njk
```diff
@@ -209,7 +209,7 @@
                 {% for item in formValues %}
                     {# When submitting an empty form, the formValues is an array containing an empty string #}
                     {# This is to cater for other input types which are part of the Add-Another pattern #}
-                    {% if item.length > 1 %}
+                    {% if (item.length > 1) and (additionalCondition.uploadSummary[0].thumbnailImage) %}
```
andrewrlee, line 212: not sure why this is `0` rather than `i`?

mgreen-mojhub: This is existing code present before the change. It looks like FarSight intended the uploadSummary to contain multiple files against a condition. But then hardcoded reference to multiple exclusions zones to look for the first index, as its only one file per condition.

The problem with uploading more than one file against the same condition relates to the other data belonging to the condition. The file name region is a text area. So you cannot uploads more than one file using this method, because the file name region can only be added once.

server/routes/creatingLicences/handlers/additionalLicenceConditionUploadsHandler.ts
mgreen-mojhub: I missed this test file completely, i'll add one now.

andrewrlee (approved):
LGTM

PR #389, opened 2022-10-19
CVSL 164 UI display prison and probation tabs in ca caseload view

GurnankCheema:
Acceptance Criteria
CA caseload to be divided into two tabs:  People in prison, People on probation

The “people in prison” tab should only include licences for the prison(s) associated to the CA user, and where the licence state is “Not started”, “In progress”, “Submitted”, “Approved” AND should NOT include “Active” licences

 The “people in prison” tab excludes “not_started” cases if the release date is more than 4 weeks in the future

The “people on probation” tab should only include licences for the prison(s) associated to the CA user, and where the licence state is “active”, “variation_in_progress”, “variation_Submitted”, “variation_Approved” AND “Active”


server/routes/changeLocation/handlers/changeLocation.ts
```diff
@@ -11,15 +11,17 @@ export default class ChangeLocationRoutes {
       const prisonCaseloadFromNomis = await this.userService.getPrisonUserCaseloads(user)
       const caseload = prisonCaseloadFromNomis.map(c => ({ value: c.caseLoadId, text: c.description }))
       const checked = req.session.caseloadsSelected
-      const cancelLink = role === AuthRole.CASE_ADMIN ? '/licence/view/cases' : '/licence/approve/cases'
+      const cancelToUrl = req.query.view ? '/licence/view/cases?view=probation' : '/licence/view/cases'
+      const cancelLink = role === AuthRole.CASE_ADMIN ? cancelToUrl : '/licence/approve/cases'
       res.render('pages/changeLocation', { caseload, checked, cancelLink })
```
GurnankCheema, line 16: Change location and cancel link are not requested as part of the ticket's AC's. However the change location and cancel link need to be updated as now they serve 2 different views and  following their execution, the user needs to to be returned to their original view.

server/routes/changeLocation/handlers/changeLocation.ts
```diff
@@ -11,15 +11,17 @@ export default class ChangeLocationRoutes {
       const prisonCaseloadFromNomis = await this.userService.getPrisonUserCaseloads(user)
       const caseload = prisonCaseloadFromNomis.map(c => ({ value: c.caseLoadId, text: c.description }))
       const checked = req.session.caseloadsSelected
```
GurnankCheema, line 13: Change location and cancel link are not requested as part of the ticket's AC's. However the change location and cancel link need to be updated as now they serve 2 different views and following their execution, the user needs to to be returned to their original view.

server/routes/viewingLicences/handlers/viewCases.ts
```diff
@@ -14,15 +14,16 @@ export default class ViewAndPrintCaseRoutes {
 
   GET = async (req: Request, res: Response): Promise<void> => {
     const search = req.query.search as string
-
+    const view = req.query.view || 'prison'
```
GurnankCheema, line 17: default view is prison

server/routes/viewingLicences/handlers/viewCases.ts
```diff
@@ -38,7 +39,10 @@ export default class ViewAndPrintCaseRoutes {
             _.head(c.licences).status !== LicenceStatus.NOT_IN_PILOT &&
             _.head(c.licences).status !== LicenceStatus.OOS_RECALL &&
             _.head(c.licences).status !== LicenceStatus.OOS_BOTUS &&
-            _.head(c.licences).status !== LicenceStatus.IN_PROGRESS,
+            _.head(c.licences).status !== LicenceStatus.IN_PROGRESS &&
+            _.head(c.licences).status !== LicenceStatus.VARIATION_IN_PROGRESS &&
+            _.head(c.licences).status !== LicenceStatus.VARIATION_APPROVED &&
+            _.head(c.licences).status !== LicenceStatus.VARIATION_SUBMITTED,
```
GurnankCheema, line 45: the variation types should not be clickable as OMU (prison) staff can't do variations

server/services/caseloadService.ts
```diff
@@ -95,8 +95,30 @@ export default class CaseloadService {
       })
 
     const combinedCases = await Promise.all([casesWithLicences, casesPendingLicence]).then(c => c.flat())
```
GurnankCheema, line 97: filterByPrisonOrProbationView is the key function. It is called here because all the existing logic such as for ensuring results only include releases within 4 weeks, is already executed.
Essentially, we are splitting the single view user previously saw into 2 parts based on the licence status.
The Variation type statuses are however an addition to what the user could previously see.

server/services/licenceService.ts
```diff
@@ -355,6 +355,9 @@ export default class LicenceService {
       LicenceStatus.APPROVED.valueOf(),
       LicenceStatus.SUBMITTED.valueOf(),
       LicenceStatus.IN_PROGRESS.valueOf(),
+      LicenceStatus.VARIATION_APPROVED.valueOf(),
+      LicenceStatus.VARIATION_IN_PROGRESS.valueOf(),
+      LicenceStatus.VARIATION_SUBMITTED.valueOf(),
     ]
```
GurnankCheema, line 361: need to also get licences with these statuses from DB

server/views/pages/view/cases.njk
```diff
@@ -63,32 +63,51 @@
...
-            <div class="govuk-form-group">
-                <label class="govuk-visually-hidden" for="search">
-                    Search for a case
-                </label>
-                <input class="govuk-input moj-search__input govuk-!-font-size-14" id="search" name="search" placeholder="You can search by name, prison number or probation practitioner" value="{{ search }}"></div>
-                <button class="govuk-button moj-search__button" data-module="govuk-button">
-                    Search
-                </button>
-            </form>
-        </div>
```
GurnankCheema, line 83: splitting into tabs

server/services/caseloadService.ts
```diff
@@ -95,8 +95,30 @@ export default class CaseloadService {
...
+    const casesFilteredByView = this.filterByPrisonOrProbationView(view, combinedCases)
```
GurnankCheema, line 98: filterByPrisonOrProbationView is the key function. It is called here because all the existing logic such as ensuring results only include releases within 4 weeks, is already executed.
Essentially, we are splitting the single view user previously saw into 2 parts based on the licence status.
The Variation type statuses are however an addition to what the user could previously see.

server/services/caseloadService.ts
```diff
@@ -68,9 +68,9 @@ export default class CaseloadService {
       .then(caseload => this.mapResponsibleComsToCases(caseload))
   }
 
-  async getOmuCaseload(user: User, prisonCaseload: string[]): Promise<ManagedCase[]> {
-    // Get cases with a licence in ACTIVE, APPROVED, or SUBMITTED state
-    const casesWithLicences = this.licenceService
+  async getOmuCaseload(user: User, prisonCaseload: string[], view: string): Promise<ManagedCase[]> {
+    // Get cases with a licence in ACTIVE, APPROVED, SUBMITTED, IN_PROGRESS or VARIATION_IN_* state
+    const casesWithLicences = await this.licenceService
```
GurnankCheema, line 73: NOT_STARTED status does not exist in the licences DB which means the NOT_STARTED licences are obtained via the call to prisonerService.searchPrisonersByReleaseDate on L81
This already filters out those with release dates going past 4 weeks in future

integration_tests/pages/changeLocationPage.ts
```diff
@@ -11,8 +11,8 @@ export default class ChangeLocationPage extends Page {
     return cy.get('[data-qa=continue]').click()
   }
 
-  clickCancelForCA = (): ViewCasesPage => {
-    cy.get('[href="/licence/view/cases"]').click()
+  clickCancelLink = (): ViewCasesPage => {
+    cy.get('[data-qa=cancel-link]').click()
     return Page.verifyOnPage(ViewCasesPage)
```
GurnankCheema, line 16: can't use href to target element anymore because it has a query param that is different for each view

server/routes/changeLocation/handlers/changeLocation.ts
```diff
@@ -11,15 +11,17 @@ export default class ChangeLocationRoutes {
...
-      const cancelLink = role === AuthRole.CASE_ADMIN ? '/licence/view/cases' : '/licence/approve/cases'
+      const cancelToUrl = req.query.view ? '/licence/view/cases?view=probation' : '/licence/view/cases'
```
SebNegahbanMoJ, line 14: Non-blocking, but I'm not sure if I agree with the approach of having a generic `view` param on the query and only checking if it's defined before setting the view to `probation`.
I understand that `prison` is the default, so if a view is defined then it is currently safe to assume that it will only be `probation`, but I think having `req.query.view === 'probation' ? ...` would make the code easier to follow (without the assumption of prison being the default) and also would allow us to expand the `view` param more easily in future.

server/views/pages/view/cases.njk
```diff
@@ -63,32 +63,51 @@
...
+        <nav class="moj-sub-navigation" aria-label="Sub navigation">
+            <ul class="moj-sub-navigation__list">
+                <li class="moj-sub-navigation__item">
+                    <a class="moj-sub-navigation__link" data-qa="prison-view-link"  {% if not probationView %}aria-current="page"{% endif %} href="?view=prison">People in prison</a>
```
SebNegahbanMoJ, line 79: This perhaps links to the comment I put above, where we are passing `view=prison` as a query param, but if the `view` is present previously, we assume the link is to probation. It may be that these lines do not interact with each other in a way that causes an issue, but in my opinion it makes the code less clear.

SebNegahbanMoJ (approved):
Happy to approve with the non-blocking comments above for consideration

server/routes/changeLocation/handlers/changeLocation.ts
GurnankCheema: good point.

PR #391, opened 2022-10-20
CVSL-348

mgreen-mojhub:
Update licenceComparator.ts to handle differences between multiple exclusions zones. Relying on the condition code no longer works with multiple conditions with the same type.


GurnankCheema (approved):
LGTM

PR #392, opened 2022-10-21
CVSL-662 rollout November 2022

mgreen-mojhub:
Update config to permit North East, North West and West Midlands access.

Add temporary filter to exclude cases for release prior to 6th November 2022. This should be removed on the 7th November 2022.


andrewrlee (approved):
LGTM

SebNegahbanMoJ (approved):

PR #397, opened 2022-10-24
CVSL-138 Adding version locking to ensure returned policy conditions are correct for the licence.

SebNegahbanMoJ:


andrewrlee (approved):
LGTM

PR #400, opened 2022-10-24
CVSL-138 Moving delete and replaced conditions to the start of the policy changes array to prevent a bug

SebNegahbanMoJ:


andrewrlee (approved):
needs tests and LGTM

PR #405, opened 2022-10-28
Adding specs to cover inflight licences work

SebNegahbanMoJ:
Also replaces `mockReturnValue(Promise<>)` with `mockResolvedValue()` and some minor refactoring to make better use of new fields returned by the API


server/routes/creatingLicences/handlers/additionalLicenceConditions.test.ts
```diff
@@ -8,14 +8,12 @@ import ConditionService from '../../../services/conditionService'
...
 
-jest.spyOn(conditionService, 'getGroupedAdditionalConditions').mockReturnValue(
-  Promise.resolve([
-    {
-      category: 'group1',
-      conditions: [{ text: 'Condition 1', code: 'condition1', category: 'group1', requiresInput: false }],
-    },
-  ])
-)
+jest.spyOn(conditionService, 'getGroupedAdditionalConditions').mockResolvedValue([
```
andrewrlee, line 11: you usually don't need to manually create spies, you can add:  

`jest.mock('../../../services/conditionService')` 

after the import and it means if your instantiated instances will be mocked and you can just call:

```
conditionService.getGroupedAdditionalConditions.mockResolvedValue(...)
```

server/routes/varyingLicences/handlers/policyChange.ts
```diff
@@ -118,19 +115,28 @@ export default class PolicyChangeRoutes {
...
-      req.body.additionalConditions?.forEach(async (code: string) => {
-        licenceConditionCodes.push(code)
-        const replacement = await this.conditionService.getAdditionalConditionByCode(code, licence.version)
-        if (replacement.requiresInput) {
-          inputs.push(code)
-        }
-      })
+      if (req.body.additionalConditions?.length > 0) {
+        const replacementArray = await Promise.all(
+          req.body.additionalConditions.map(async (code: string) => {
```
andrewrlee, line 120: meh don't need the async

andrewrlee (dismissed):
LGTM

server/routes/varyingLicences/handlers/policyChange.test.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  describe('POST', () => {
+    it('should redirect to the policyChangesInputCallback', async () => {
+      await handler.POST(req, res)
+      expect(res.redirect).toHaveBeenCalledWith('/licence/vary/id/1/policy-changes/input/callback/1')
+    })
+
+    it('sets the changeConditionInputs in session storage', async () => {
+      req.body.additionalConditions = ['code 5', 'code 7']
+      condition1 = { ...condition1, suggestions: [{ code: 'code 5' }, { code: 'code 7' }] } as LicenceConditionChange
```
mgreen-mojhub, line 165: I would assign a new variable here, rather than overwrite the top level variable. It's fine here, but if other tests are added at a later point, they may rely on the initial state of the object.

server/routes/varyingLicences/handlers/policyChange.test.ts
```diff
@@ -0,0 +1,186 @@
...
+jest.mock('../../../services/conditionService')
+
+describe('Route handlers', () => {
+  const handler = new PolicyChangeRoutes(licenceService, conditionService)
+  let req: Request
+  let res: Response
+
+  licenceService.getParentLicenceOrSelf.mockResolvedValue({ id: 1, version: 'version', typeCode: 'AP_PSS' } as Licence)
+
+  let condition1 = {
```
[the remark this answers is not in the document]
SebNegahbanMoJ: Apologies, I did look into this but I later pass in conditions1-4 as mocked responses/ session objects, which I would need to redefine if I were to create a new variable in the individual tests. It can definitely be done, but I did it this way to avoid repetition of code.

server/routes/varyingLicences/handlers/policyChange.test.ts
SebNegahbanMoJ: Copying and pasting response to the other related comment for completion's sake

> Apologies, I did look into this but I later pass in conditions1-4 as mocked responses/ session objects, which I would need to redefine if I were to create a new variable in the individual tests. It can definitely be done, but I did it this way to avoid repetition of code.

mgreen-mojhub (approved):

PR #412, opened 2022-11-03
CVSL-537

mgreen-mojhub:
Wrap dateFormat in try catch to avoid page crash when viewing null dates. Longer term fix should filter out none dated records, or handle according to business requirment.

### DIFF
--- a/server/routes/approvingLicences/handlers/approvalCases.ts
+++ b/server/routes/approvingLicences/handlers/approvalCases.ts
@@ -4,8 +4,6 @@ import { format, getUnixTime } from 'date-fns'
 import CaseloadService from '../../../services/caseloadService'
 import PrisonerService from '../../../services/prisonerService'
 import { convertToTitleCase, selectReleaseDate } from '../../../utils/utils'
-import { Prisoner } from '../../../@types/prisonerSearchApiClientTypes'
-import logger from '../../../../logger'
 
 export default class ApprovalCaseRoutes {
   constructor(private readonly caseloadService: CaseloadService, private readonly prisonerService: PrisonerService) {}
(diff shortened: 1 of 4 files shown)


server/utils/utils.ts
```diff
@@ -157,6 +160,26 @@ const isBankHolidayOrWeekend = (date: Moment, bankHolidays: Holiday[]) => {
...
+
+  if (nomisRecord.confirmedReleaseDate) {
+    dateString = nomisRecord.confirmedReleaseDate
+  }
+
+  if (nomisRecord.conditionalReleaseOverrideDate) {
+    dateString = nomisRecord.conditionalReleaseOverrideDate
+  }
+
+  try {
```
andrewrlee, line 174: might be best to check if date is defined or not - try catch suggests the date is an invalid format which should never be the case.

andrewrlee (approved):
LGTM

server/routes/approvingLicences/handlers/approvalCases.ts
```diff
@@ -27,7 +26,7 @@ export default class ApprovalCaseRoutes {
           name: convertToTitleCase(`${c.nomisRecord.firstName} ${c.nomisRecord.lastName}`.trim()),
           prisonerNumber: c.nomisRecord.prisonerNumber,
           probationPractitioner: c.probationPractitioner,
-          releaseDate: format(new Date(this.selectReleaseDate(c.nomisRecord)), 'dd MMM yyyy'),
```
andrewrlee, line 30: worth adding a test for this?

I quite like the practice of adding a failing test case and then fixing it

andrewrlee (approved):
LGTM

PR #413, opened 2022-11-03
Fixing a bug where variations would not trigger the validation to ensure all conditions had data inputs

SebNegahbanMoJ:
When viewing the check-answers page, there should be a validation that prevents a user from submitting a licence with placeholders still within the condition text (ie, they haven't entered the required data in the child page).
This validation triggers on newly created licences, but was not triggering on variations.

This PR makes it trigger on variations.


mgreen-mojhub (approved):
LGTM

PR #415, opened 2022-11-04
CVSL-710 initial PP email sent only for NOT_STARTED status cases

GurnankCheema:
It was previously being sent for for both NOT_STARTED and IN_PROGRESS statuses


andrewrlee (approved):
LGTM

PR #418, opened 2022-11-07
CVSL-656 - Auto-populate end date for condition 14b

SebNegahbanMoJ:
Adds a new input type called `info` that allows the user to review an input in a read-only fashion.
Also removes the validations for 14b as all versions of the policy are changing to be `info` inputs.

The change to `nunjucksSetup` is to prevent the value of `infoInputReviewed` from appearing in the summary boxes that show the inputs associated with a condition.


server/routes/creatingLicences/handlers/additionalLicenceConditionsCallback.ts
```diff
@@ -36,4 +36,10 @@ export default class AdditionalLicenceConditionsCallbackRoutes {
     }
     return res.redirect(`/licence/create/id/${licenceId}/bespoke-conditions-question`)
   }
+
+  needsReview = (condition: AdditionalCondition) => {
+    return condition.data.some(data => {
+      return data.field === 'infoInputReviewed' && data.value !== 'true'
+    })
```
andrewrlee, line 43: nice to have some test coverage for this?

server/views/formBuilder.njk
```diff
@@ -231,6 +231,25 @@
...
+                    <input type="hidden" name="{{input.name}}" value="{{formValues}}" />
+                    <input hidden name="infoInputReviewed" value="true" />
+                    <h2 class="govuk-heading-m">Help with this licence condition</h2>
+                    {{ govukDetails({
+                        summaryText: input.helpLink.summary,
+                        text: input.helpLink.text,
+                        classes: 'govuk-!-margin-bottom-3'
+                    }) }}
+                </div>
+            {% endif %}
```
andrewrlee, line 252: should there be an end to end test to cover the behaviour of this kind of condition?

server/routes/creatingLicences/handlers/additionalLicenceConditionsCallback.ts
andrewrlee: Sorry, I didn't mean a unit test for this function - more a supertest test to cover how this impacts navigation

server/views/formBuilder.njk
andrewrlee: By end to end I meant cypress test. Just as this has a lot of moving parts that affect the wider add conditions flow.

Does this change impact vary and licence version migration?

SebNegahbanMoJ: There's now a cypress test to cover this!

SebNegahbanMoJ: > Does this change impact vary and licence version migration?
It shouldn't, as we're just updating the existing licence policies to use the new condition config, rather than creating a new version of the policy. This shouldn't be a problem as we're removing an input, rather than adding a new one.

PR #422, opened 2022-11-10
CVSL 709 refactor for tests

GurnankCheema:
1, The logic within excludeCasesNotAssignedToPpWithinPast7Days was wrong. It was thought the allocationDate referred to the date a PP was allocated. In fact the allocationDate is part of the offenderCrn object. 
A call is made to probation-offender-search to get the offender details. The returned object contains the offenderManagers array exampled below. The relevant data from the offenderManagers array is in bold. 
"offenderManagers": [
           {
               "staff": {
                   "code": "A123456B",
                   "forenames": "some",
                   "surname": "practioner"
               },
               "partitionArea": "National Data",
               "softDeleted": false,
               "team": {
                   "code": "someTeamCode",
                   "description": "Durham & Darlington IOM",
                   "district": {
                       "code": "someDistrictCode",
                       "description": "Durham"
                   },
                   "borough": {
                       "code": "someBoroughCode",
                       "description": "County Durham and Darlington"
                   }
               },
               "probationArea": {
                   "code": "N54",
                   "description": "North East Region",
                   "nps": false
               },
               **"fromDate": "2022-11-02",
               "active": true,**
               "allocationReason": {
                   "code": "someReason",
                   "description": "Caseload Adjustment"
               }
           }
       ],

2, some logic extracted out of promptLicenceCreation.ts  for code simplification and to enable testing
3, calls to promptLicenceCreationService changed to be in a more natural chronological order.

### DIFF
--- a/jobs/promptLicenceCreation.ts
+++ b/jobs/promptLicenceCreation.ts
@@ -1,6 +1,6 @@
 import 'reflect-metadata'
 import _ from 'lodash'
-import { add, subDays, startOfISOWeek, endOfISOWeek } from 'date-fns'
+import { add, startOfISOWeek, endOfISOWeek } from 'date-fns'
 import { buildAppInsightsClient, flush, initialiseAppInsights } from '../server/utils/azureAppInsights'
 import logger from '../logger'
 import { Prisoner } from '../server/@types/prisonerSearchApiClientTypes'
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "record-build-info": "node ./bin/record-build-info",
     "lint": "eslint . --cache --max-warnings 0",
     "typecheck": "tsc",
-    "test": "jest --collectCoverage=true --testPathPattern=/server/.*",
+    "test": "jest --collectCoverage=true --detectOpenHandles",
```
andrewrlee, line 24: I think detectOpenHandles has a bit of a performance cost

jobs/promptLicenceCreation.ts
```diff
@@ -10,38 +10,20 @@ import { ManagedCase } from '../server/@types/managedCase'
...
+const promptLicenceCreationService = new PromptLicenceCreationService(prisonerService, caseloadService)
 
 const buildEmailGroups = async (
-  initialPromptCases: ManagedCase[],
-  urgentPromptCases: ManagedCase[]
+  urgentPromptCases: ManagedCase[],
+  initialPromptCases: ManagedCase[]
 ): Promise<EmailContact[]> => {
-  const managedCases = [...initialPromptCases, ...urgentPromptCases]
+  const managedCases = [...urgentPromptCases, ...initialPromptCases]
```
andrewrlee, line 26: probably doesn't matter but don't think this needed to change

GurnankCheema: I just put them in the same order just so the code flows nicely. I know it doesn't make any actual diff as the all the values just get added to the array .

package.json
GurnankCheema: remove it?

jobs/promptLicenceCreationService.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      deliusRecord: { offenderId: 1, offenderManagers: [{ active: true, fromDate: format(new Date(), 'yyyy-MM-dd') }] },
+      nomisRecord: {
+        restrictedPatient: false,
+        prisonId: 'someId',
+        conditionalReleaseDate: format(new Date(), 'yyyy-MM-dd'),
+        status: 'someStatus',
+      },
+      licences: [{ status: LicenceStatus.NOT_STARTED, type: LicenceType.AP }],
+      probationPractitioner: { name: 'jim' },
+    },
```
andrewrlee, line 41: meh... you could have used the same pattern that we talked about last time -  extract out a function so you make the bits of the test your interested in pop out in the test. 

This shared data between tests is a bit gnarly and you can't change it for one test without breaking other tests and it becomes difficult to change

andrewrlee (dismissed):
few comments, but (probably) no blockers?
LGTM

jobs/promptLicenceCreation.ts
andrewrlee: yes, that makes sense

jobs/promptLicenceCreationService.test.ts
GurnankCheema: test refactored. 
The two test do use the same data though

andrewrlee (approved):
LGTM

PR #423, opened 2022-11-14
CVSL-723 Capturing user info in app insights at the request level

andrewrlee:
Also deleted some bits I'm pretty sure aren't being used any longer:
* condition list
* prison register calls to get emails


GurnankCheema (approved):
LGTM

PR #428, opened 2022-11-17
Bumping TS to v4.9.3 and govuk-frontend to v4.4.0

SebNegahbanMoJ:
Dependency bumps to stop errors in circle


mgreen-mojhub (approved):
LGTM

PR #430, opened 2022-11-22
CVSL-730 chunk args passed in api calls

GurnankCheema:
The apis were throwing exceptions because too much data was passed in one go.  This was preventing promptLicenceCreation cronjob from completing.  Args chunked to 500.

app insights:
prison-api
Screenshot 2022-11-22 at 11 35 38

probation-offender-search
Screenshot 2022-11-22 at 11 42 20


mgreen-mojhub (approved):
LGTM

PR #433, opened 2022-11-24
CVSL-732 displays correct date and description for vary journey

GurnankCheema:
The date in header when displaying offender records is different for vary journey pages. 

**Jira ticket:**

Screenshot 2022-11-24 at 11 18 31

### DIFF
--- a/server/routes/varyingLicences/index.ts
+++ b/server/routes/varyingLicences/index.ts
@@ -1,4 +1,4 @@
-import { RequestHandler, Router } from 'express'
+import { RequestHandler, Request, Response, NextFunction, Router } from 'express'
 import asyncMiddleware from '../../middleware/asyncMiddleware'
 import fetchLicence from '../../middleware/fetchLicenceMiddleware'
 import roleCheckMiddleware from '../../middleware/roleCheckMiddleware'
@@ -27,6 +27,13 @@ import PolicyChangesCallbackRoutes from './handlers/policyChangesCallback'
 import PolicyChangeRoutes from './handlers/policyChange'
 import PolicyChangesInputCallbackRoutes from './handlers/policyChangesInputCallback'
 
+function alterResObject() {
+  return (req: Request, res: Response, next: NextFunction) => {
+    res.locals.isVaryJourney = true
+    next()
+  }
+}
+
 export default function Index({
   licenceService,
   caseloadService,
@@ -48,6 +55,7 @@ export default function Index({
       routePrefix(path),
       roleCheckMiddleware(['ROLE_LICENCE_RO']),
       fetchLicence(licenceService),
+      alterResObject(),
       asyncMiddleware(handler)
     )
 
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -1,13 +1,21 @@
 /* eslint-disable no-param-reassign */
 import nunjucks, { Environment } from 'nunjucks'
+import { isToday, format } from 'date-fns'
 import express from 'express'
 import path from 'path'
 import moment from 'moment'
 import { filesize } from 'filesize'
 import { FieldValidationError } from '../middleware/validationMiddleware'
 import config from '../config'
-import { formatAddress, jsonDtTo12HourTime, jsonDtToDate, jsonDtToDateShort, jsonDtToDateWithDay } from './utils'
-import { AdditionalCondition, AdditionalConditionData } from '../@types/licenceApiClientTypes'
+import {
+  formatAddress,
+  jsonDtTo12HourTime,
+  jsonDtToDate,
+  jsonDtToDateShort,
+  jsonDtToDateWithDay,
+  toDate,
+} from './utils'
+import { AdditionalCondition, AdditionalConditionData, Licence } from '../@types/licenceApiClientTypes'
 import SimpleTime from '../routes/creatingLicences/types/time'
 import SimpleDate from '../routes/creatingLicences/types/date'
 import Address from '../routes/creatingLicences/types/address'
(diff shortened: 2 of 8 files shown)


server/views/pages/create/checkAnswers.test.ts
```diff
@@ -403,4 +404,16 @@ describe('Create a Licence Views - Check Answers', () => {
     expect($('.check-answers-header__change-link').length).toBe(0)
     expect($('[data-qa="send-licence-conditions"]').length).toBe(0)
   })
+
```
GurnankCheema, line 407: non-vary journeys should still show 'Release date'

server/views/pages/vary/timeline.test.ts
```diff
@@ -51,4 +51,13 @@ describe('Timeline', () => {
     const $ = cheerio.load(compiledTemplate.render(viewContext))
     expect($('.moj-timeline__date').text()).not.toContain('Last update: ')
   })
+  it('should display correct date description for "vary" routes', () => {
+    viewContext = {
+      isVaryJourney: true,
+      licence: { typeCode: 'AP', licenceExpiryDate: '01/01/2022' },
+    }
+    const $ = cheerio.load(compiledTemplate.render(viewContext))
+    expect($('[data-qa=date]').text()).not.toContain('Release date: ')
```
GurnankCheema, line 60: Vary journeys should not show Release date but one of the other date descriptions eg 'Licence end date'

server/views/partials/licenceDetailsBanner.njk
```diff
@@ -8,7 +8,13 @@
                 {% if user.authSource == 'nomis' %}
                     <span class="pipe-separated__item">Prison number: {{ licence.nomsId }}</span>
                 {% endif %}
-                <span class="pipe-separated__item">Release date: {{ licence.actualReleaseDate | datetimeToDateShort or licence.conditionalReleaseDate | datetimeToDateShort }}</span>
+                    <span class="pipe-separated__item"  data-qa="date">
+                        {% if isVaryJourney %}
```
GurnankCheema, line 12: isVaryJourney was added to res.locals object in the vary journey route index page so that it is accessible to all vary routes

server/views/partials/licenceDetailsBanner.njk
```diff
@@ -8,7 +8,13 @@
                 {% if user.authSource == 'nomis' %}
                     <span class="pipe-separated__item">Prison number: {{ licence.nomsId }}</span>
                 {% endif %}
-                <span class="pipe-separated__item">Release date: {{ licence.actualReleaseDate | datetimeToDateShort or licence.conditionalReleaseDate | datetimeToDateShort }}</span>
```
GurnankCheema, line 11: isVaryJourney was added to res.locals object in the vary journey route index page so that it is accessible to all vary routes

server/utils/nunjucksSetup.ts
```diff
@@ -245,5 +253,35 @@ export function registerNunjucks(conditionService: ConditionService, app?: expre
...
+    let textToDisplay = ''
+
+    if (licenceType === 'AP') {
+      textToDisplay = 'Licence end date'
+      dateToDisplay = led
+    }
+
+    if (licenceType === 'AP_PSS') {
+      textToDisplay = isToday(tussd) ? 'PSS end date' : 'Licence end date'
+      dateToDisplay = led
```
SebNegahbanMoJ, line 271: Should this not be `isToday(tussd) ? tussd: led`?

GurnankCheema: that's what I thought at first but when I re-read it. i thought it meant the title should change, not the actual date

SebNegahbanMoJ: I think that's something we'd need to check, otherwise surely we would be showing the wrong date with the wrong label? Eg LED with the label of PSS End Date?

GurnankCheema: changes made following clarification from BA.

server/utils/nunjucksSetup.ts
```diff
@@ -245,5 +253,36 @@ export function registerNunjucks(conditionService: ConditionService, app?: expre
...
+    let dateToDisplay: Date
+    let textToDisplay = ''
+
+    if (licenceType === 'AP') {
+      textToDisplay = 'Licence end date'
+      dateToDisplay = led
+    }
+
+    if (licenceType === 'AP_PSS') {
+      textToDisplay = isToday(tussd) ? 'PSS end date' : 'Licence end date'
```
mgreen-mojhub, line 271: Does this function accept null values? Its possible tussd could be null here

server/routes/varyingLicences/index.ts
```diff
@@ -27,6 +27,13 @@ import PolicyChangesCallbackRoutes from './handlers/policyChangesCallback'
 import PolicyChangeRoutes from './handlers/policyChange'
 import PolicyChangesInputCallbackRoutes from './handlers/policyChangesInputCallback'
 
+function alterResObject() {
```
mgreen-mojhub, line 30: I'm not sure you need this middleware. Could you check the licence.isVariationOf value instead?

GurnankCheema: I don't think it will work with licence.isVariationOf or even licence.isVariation as this won't be present on first run of vary journey. 
    await this.licenceService.updateStatus(licence.variationOf, LicenceStatus.INACTIVE, user) is only executed on POST

server/utils/nunjucksSetup.ts
GurnankCheema: fair point. Have split functionality into 2 if blocks so can can check tussd exists. 
It does bring into question though of what should be rendered if there is a tused but no tussd !

mgreen-mojhub (approved):

PR #434, opened 2022-11-25
CVSL-713 inflight policy improvements

SebNegahbanMoJ:
Summarised list of changes:
- New 'Confirm deletion' page added when deleting a condition via the inflight licence journey
- Moved the 'Delete this condition' button to be alongside 'Continue' button. Was not previously possible due to needing to be in a separate form, but with the new 'Confirm deletion' page, the buttons can be side-by-side.
- Reduced spacing between bulletpointed hint text and new licence condition
- Added space above the 'Continue' button on all pages
- Policy change tracker added to child pages
- Added space between reporting times if two or more are conditionally present (eg hidden behind radio button)
- Minor grammar changes


integration_tests/mockApis/licence.ts
```diff
@@ -1,12 +1,17 @@
 import { SuperAgentRequest } from 'superagent'
 import { stubFor } from '../wiremock'
 import LicenceStatus from '../../server/licences/licenceStatus'
-import Policy from './polices/v2-0'
+// eslint-disable-next-line camelcase
+import Policyv2_0 from './polices/v2-0'
```
andrewrlee, line 5: usual style guide for js is:
data/variables = lower camel case
types and classes = upper camel case

andrewrlee: that would remove the need for eslint wrangling

server/routes/varyingLicences/handlers/policyConfirmDelete.ts
```diff
@@ -0,0 +1,11 @@
+import { Request, Response } from 'express'
+
+export default class PolicyConfirmDelete {
```
andrewrlee, line 3: missing test?

server/routes/varyingLicences/handlers/policyChange.ts
```diff
@@ -163,34 +169,38 @@ export default class PolicyChangeRoutes {
...
 
-    const counter = +changeCounter
-    const conditionCode = req.session.changedConditions[counter - 1].code
-    const conditionType: LicenceType = await this.conditionService.getAdditionalConditionType(
-      conditionCode,
-      (
-        await this.licenceService.getParentLicenceOrSelf(licenceId, user)
-      ).version
-    )
+    if (req.body['confirm-delete'] === 'yes') {
```
andrewrlee, line 172: think you're missing test coverage here

integration_tests/mockApis/licence.ts
SebNegahbanMoJ: I can fix that, but I believe it also complains about the underscore in the variable name. Not a lot I can do about that though as otherwise it would be `v20` and `v21`, which can lead to some predicable issues with future versioning!

assets/sass/components/_add-another.scss
```diff
@@ -4,3 +4,7 @@
     position: relative;
   }
 }
+
+.govuk-radios__conditional .govuk-date-input {
+  margin-bottom: 15px;
+}
```
andrewrlee, line 10: meh missing new line

andrewrlee (dismissed):
LGTM

server/routes/varyingLicences/handlers/policyConfirmDelete.ts
SebNegahbanMoJ: Added now!

integration_tests/mockApis/licence.ts
SebNegahbanMoJ: Unfortunately even dropping the case to make it `policyV2_x` still causes the camelcase error, so I've had to leave the eslint tomfoolery in

andrewrlee (approved):
LGTM

PR #437, opened 2022-11-29
CVSL-626 Extracting out formatting code

andrewrlee:
Extracting out formatting code and fixing test names/locations.

This is in preparation for reimplementing in the backend

### DIFF
--- a/server/services/conditionFormatter.test.ts
+++ b/server/services/conditionFormatter.test.ts
@@ -1,0 +1,697 @@
+import { AdditionalCondition, AdditionalConditionAp } from '../@types/licenceApiClientTypes'
+import ConditionFormatter from './conditionFormatter'
+
+describe('Conditions Provider - expansions', () => {
+  const conditionFormatter = new ConditionFormatter()
+
+  describe('formats', () => {
+    it('Will return text verbatim when no placeholders exist', async () => {
+      const condition: AdditionalCondition = {
+        id: 1,
+        code: 'ed607a91-fe3a-4816-8eb9-b447c945935c',
+        category: 'Possession, ownership, control or inspection of specified items or documents',
+        sequence: 1,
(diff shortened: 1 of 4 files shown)


server/services/conditionFormatter.ts
```diff
@@ -0,0 +1,200 @@
+import {
```
andrewrlee, line 1: This is just a lift and shift of existing code

server/services/conditionFormatter.test.ts
```diff
@@ -0,0 +1,697 @@
+import { AdditionalCondition, AdditionalConditionAp } from '../@types/licenceApiClientTypes'
```
andrewrlee, line 1: This tests formatting code in isolation

server/services/conditionService.test.ts
```diff
@@ -0,0 +1,730 @@
+import { AdditionalCondition, AdditionalConditionAp } from '../@types/licenceApiClientTypes'
```
andrewrlee, line 1: This is the existing test - an integration test, testing basically the same code at a higher level proving it still works.
This can be rewritten to mock out (so decouple) the formatter

We're lacking test coverage for other public methods on this service

server/services/conditionService.ts
```diff
@@ -141,58 +143,7 @@ export default class ConditionService {
...
-          conditionText = this.replacePlaceholderWithValue(ph, conditionText, value)
-        }
-      })
-
-      // Remove any unmatched placeholders that remain
-      return this.removeAllPlaceholders(conditionText)
-    }
-    // No input was required - keep the condition verbatim to print on the licence
-    return configCondition.text as string
+    return this.conditionFormatter.format(configCondition, enteredData)
```
andrewrlee, line 146: we now delegate to the formatter

mgreen-mojhub (approved):
LGTM

PR #447, opened 2022-12-07
CVSL-639 make input accessibility compliant

GurnankCheema:
AC: Implement the code in the ticket to make it easier for users of speech recognition software to navigate the page

Could have removed the H1 altogether and retained the govukInput nunjucks component and set isPageHeader to true. However there is an govukWarningText component (only displays if dates aren't compatible) that then wouldn't render in the correct place. It would sit above the H1. So have changed from using nunjucks component to straight HTML. 

**Retaining the nunjucks component would have presented like this**
would have looked

**instead of** 
required look


andrewrlee (approved):
LGTM

PR #448, opened 2022-12-07
CVSL-640 CSS for signout link. 

GurnankCheema:
Becomes black on click for accessibility colour contrast and underline on hover, like other services.


andrewrlee (approved):
lgtm

PR #450, opened 2022-12-07
CVSL-636 styling change and setting of h2

GurnankCheema:
Changes made in collaboration with the design team
The only way to set the section headings as h2s was to embed them inside the legend


andrewrlee (approved):
LGTM

PR #452, opened 2022-12-09
Updating licence api type imports and refactoring conditionService

SebNegahbanMoJ:
Replacing manually-written type definitions with those created by openapi
Refactors ConditionService to not return the full licence policy when not needed


server/data/licenceApiClient.ts
```diff
@@ -393,17 +393,17 @@ export default class LicenceApiClient extends RestClient {
...
     try {
-      return this.get({ path: `/licence-policy/version/${version}` }) as Promise<LicencePolicy>
+      return this.get({ path: `/licence-policy/version/${version}` }) as Promise<LicencePolicyResponse>
     } catch (error) {
       return error.status >= 400 && error.status < 500 ? null : error
     }
   }
 
-  async getActiveConditions(): Promise<LicencePolicy> {
+  async getActiveConditions(): Promise<LicencePolicyResponse> {
```
andrewrlee, line 404: would generally expect a method called getConditions to return a list of conditions.

maybe `getLicencePolicy` and `getLicencePolicyForVersion` would be better names for these?

server/services/conditionService.ts
```diff
@@ -86,42 +92,30 @@ export default class ConditionService {
...
-        }
+    additionalConditions[licenceType].forEach((condition: AdditionalConditionAp | AdditionalConditionPss) => {
+      const collection = map.get(condition.category)
+      if (!collection) {
+        map.set(condition.category, [condition])
+      } else {
+        collection.push(condition)
       }
-    )
+    })
```
andrewrlee, line 104: meh.. if you're looping over something to build up an object, a `reduce` would be a bit more idiomatic.

server/services/conditionService.ts
```diff
@@ -67,17 +73,17 @@ export default class ConditionService {
...
 
+  async getAdditionalConditions(version: string): Promise<AdditionalConditionsConfig> {
+    const policy = await this.getConditions(version)
+    return this.parseResponse(policy.additionalConditions)
+  }
+
   async getAdditionalConditionByCode(
     searchCode: string,
     version: string = null
   ): Promise<AdditionalConditionAp | AdditionalConditionPss> {
```
andrewrlee, line 84: meh.. you could add a `typealias` for `AdditionalConditionAp | AdditionalConditionPss` as its used in a few places

server/services/conditionService.ts
```diff
@@ -86,42 +92,30 @@ export default class ConditionService {
...
     version: string = null
   ): Promise<{ category: string; conditions: AdditionalConditionAp[] | AdditionalConditionPss[] }[]> {
-    let policyConditions: LicencePolicy
-    if (version) {
-      policyConditions = await this.getConditions(version)
-    } else {
-      policyConditions = await this.getActiveConditions()
-    }
+    const additionalConditions = await this.getAdditionalConditions(version)
     const map = new Map()
```
andrewrlee, line 96: meh... Not your change but the Map should have generics for `Map<Key, Value>()`

andrewrlee (approved):
LGTM

server/services/conditionService.ts
SebNegahbanMoJ: This isn't actually anything I've changed in this PR, just a formatting change making GH think it's new code. I can certainly look into changing it to a reduce though!

andrewrlee: ah cool, no worries I misread it

andrewrlee (approved):
LGTM

PR #457, opened 2022-12-14
CVSL-746

mgreen-mojhub:
Ensure bookingId is present within HDC payload, otherwise return null from the function call.


andrewrlee (approved):
lgtm

PR #458, opened 2022-12-14
Fixing url generation bug for additional condition input delete button

SebNegahbanMoJ:
Fixing some unexpected behaviour where 'undefined' was being added to urls when trying to delete an additional condition from its input page.


andrewrlee (approved):
lgtm

PR #461, opened 2022-12-15
Cvsl 738 override licence status

mgreen-mojhub:
Add ability to override licence status using the support pages.


server/enumeration/licenceStatus.ts
```diff
@@ -16,4 +16,15 @@ enum LicenceStatus {
   OOS_RECALL = 'OOS_RECALL',
 }
 
+const SelectableLicenceStatus = Object.values(LicenceStatus).filter(
```
andrewrlee, line 19: think this should be lower camel case as its data not a Type

mgreen-mojhub: now replaced with snake-case

andrewrlee (approved):
LGTM

PR #464, opened 2022-12-19
update-redis-client-v4

mgreen-mojhub:
Update the redis client from v3 to v4. Requires code refactoring to make use of new functions provided by the library.


GurnankCheema (approved):
LGTM

PR #465, opened 2022-12-20
Adding IS91 to the allowed legal statuses to be activated.

SebNegahbanMoJ:
Adding 'IMMIGRATION_DETAINEE' as a valid legal status for licences that are to be activated by the automatic job


mgreen-mojhub (approved):
LGTM

PR #467, opened 2022-12-22
Changing confirmedReleaseDate from prisoner profile to ARD on the licence

SebNegahbanMoJ:
Changing licence activation check to use the ARD on the licence rather than confirmedReleaseDate in NOMIS

Explanation:
When an offender has an IS91 warrant as their primary conviction (resulting in the legalStatus being `IMMIGRATION_DETAINEE`), the confirmedReleaseDate in NOMIS will always be `NULL`. As such, we need to use the actualReleaseDate/ARD that's stored (and preserved) on the licence.

Also changing datesChangedEventHandler to use legalStatus rather than imprisonmentStatus for consistency, and to cover very rare edge cases.


mgreen-mojhub (approved):
LGTM

andrewrlee (approved):
LGTM

PR #472, opened 2022-12-29
Wrapping validation middleware in async error handling

andrewrlee:
This should stop a source of pages hanging if there are issues with the backend API. Also short circuited call to the backend when no condition code is present which would happen on every submit whether there are conditions or not


server/routes/support/index.ts
```diff
@@ -42,9 +42,9 @@ export default function Index({
 
   get('/', supportHomeHandler.GET)
   get('/manage-omu-email-address', manageOmuEmailAddressHandler.GET)
-  post('/manage-omu-email-address/add-or-edit', manageOmuEmailAddressHandler.ADD_OR_EDIT, prisonIdAndEmail)
-  post('/manage-omu-email-address/delete', manageOmuEmailAddressHandler.DELETE, prisonIdDelete)
-  post('/manage-omu-email-address', manageOmuEmailAddressHandler.CURRENT, prisonIdCurrent)
+  post('/manage-omu-email-address/add-or-edit', manageOmuEmailAddressHandler.ADD_OR_EDIT, PrisonIdAndEmail)
+  post('/manage-omu-email-address/delete', manageOmuEmailAddressHandler.DELETE, PrisonIdDelete)
+  post('/manage-omu-email-address', manageOmuEmailAddressHandler.CURRENT, PrisonIdCurrent)
```
andrewrlee, line 47: types should be Capitalised

server/middleware/validationMiddleware.ts
```diff
@@ -1,24 +1,32 @@
...
+async function getValidator(
+  conditionService: ConditionService,
+  licenceVersion: string | undefined,
+  conditionCode: string,
+  type: new () => object
+): Promise<new () => object> {
+  if (!conditionCode) return type
+  const condition = await conditionService.getAdditionalConditionByCode(conditionCode, licenceVersion)
+  return condition?.validatorType || type
+}
```
andrewrlee, line 22: no short circuiting if no condition code. 

Don't know why the type fall back exists - really horrible as this is used for every POST in the entire application and would need to check every use to see if it actually is used

server/middleware/validationMiddleware.ts
```diff
@@ -1,24 +1,32 @@
...
+
 function validationMiddleware(conditionService: ConditionService, type?: new () => object): RequestHandler {
-  return async (req, res, next) => {
-    const { licence } = res.locals
+  return asyncMiddleware(async (req, res, next) => {
```
andrewrlee, line 25: wrapping in async middleware which will forward any unexpected async errors to the next function (and so pipe through to the error handler)

PR #480, opened 2023-01-13
Fixing 'invalid date' issue caused by trying to load CRD and ARD from NOMIS rather than licence

SebNegahbanMoJ:
When on approver/OMU caseload view screens, an error could occur when rendering the release date field, saying 'invalid date'. This appears to have been caused by trying to load the CRD/ARD on cases where it has been removed from NOMIS.
As these dates should be available on the licence in these cases, this change makes these pages use the dates on the CVL licence (if one is available). It will only revert to NOMIS if there is no licence (eg if a licence hasn't been started yet).


mgreen-mojhub (approved):
LGTM

server/utils/utils.ts
```diff
@@ -193,6 +193,23 @@ const selectReleaseDate = (nomisRecord: Prisoner) => {
...
+  let dateString = licence.actualReleaseDate || licence.conditionalReleaseDate
+
+  if (!dateString) {
+    logger.error(`No release date found for NOMIS ID: ${licence.nomsId}`)
+    return 'not found'
+  }
+
+  try {
+    dateString = format(new Date(dateString), 'dd MMM yyyy')
+  } catch (e) {
```
andrewrlee, line 206: not sure about this error handling -could be wrong but seems unnecessarily defensive considering the licence date from CVL will always be `dd/MM/yyyy`? 

https://github.com/ministryofjustice/create-and-vary-a-licence-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Licence.kt#L80

server/utils/utils.test.ts
```diff
@@ -414,94 +416,174 @@ describe('Get prisoner release date from Nomis', () => {
...
+    const nomisRecord = {
+      confirmedReleaseDate: '2035-10-22',
+    } as Prisoner
+
+    expect(releaseDateLabel(null, nomisRecord)).toBe('Confirmed release date')
+  })
+
+  it('returns "CRD" if the release date from the licence is a CRD, regardless of NOMIS record', () => {
+    const licence = {
+      conditionalReleaseDate: '2035-11-22',
```
andrewrlee, line 565: Think the date format is dd/MM/YYYY

also be wary that javascript new date interprets `MM/dd/YYYY`
Screenshot 2023-02-03 at 08 18 05

mgreen-mojhub (approved):
LGTM

server/utils/utils.ts
SebNegahbanMoJ: I believe it's there in case the date comes back in a malformed state from NOMIS or elsewhere. We have specs to  test for that scenario, so my assumption is that it's something that does occasionally happen or has happened in the past.

SebNegahbanMoJ: I had written a comment about this being in case the date comes back malformed from NOMIS, but in hindsight I think it's actually leftover from me copying the method from the selectReleaseDate method in the same file.
It may be overly defensive, but I'm also not sure it does any harm to leave it in place, to make sure the date in the CVL db isn't also malformed somehow?

server/utils/utils.test.ts
SebNegahbanMoJ: You're exactly right and I've added a step to avoid the `new Date` issue (and also to add consistency to date formatting if the date is somehow corrupted/invalid)

server/routes/approvingLicences/handlers/approvalCases.test.ts
```diff
@@ -112,6 +117,17 @@ describe('Route Handlers - Approval - case list', () => {
...
+
+    licenceService.getLicence.mockResolvedValue({
+      id: 1,
+      typeCode: 'AP',
+      additionalLicenceConditions: [],
+      additionalPssConditions: [],
+      bespokeConditions: [],
+      isVariation: false,
+      conditionalReleaseDate: '2022-1-5',
+      actualReleaseDate: '2022-1-3',
```
andrewrlee, line 129: wouldn't the licence service return dates in `dd/MM/yyy` format?

server/utils/utils.ts
andrewrlee: Unnecessarily defensive code just leads to more and more defensive code. 

Code like this is a statement that the format of dates from a licence is not known - which makes everywhere that doesn't express similar uncertainty look like a possible bug

The dates in the CVL DB are typed at `DateTime` and the API has consistent (if annoying) formatting rules so the uncertainty is unnecessary.

server/utils/utils.ts
```diff
@@ -206,6 +230,13 @@ const isPassedArdOrCrd = (licence: LicenceSummary, prisoner: Prisoner | PrisonAp
   return false
 }
 
+const releaseDateLabel = (licence: Licence, nomisRecord: Prisoner): string => {
+  if (licence) {
+    return licence.actualReleaseDate ? 'Confirmed release date' : 'CRD'
+  }
+  return nomisRecord.confirmedReleaseDate ? 'Confirmed release date' : 'CRD'
```
andrewrlee, line 237: is this right? the labels are the same?

server/utils/utils.ts
```diff
@@ -192,6 +197,25 @@ const selectReleaseDate = (nomisRecord: Prisoner) => {
   return dateString
 }
 
+const selectReleaseDateFromLicence = (licence: Licence) => {
```
andrewrlee, line 200: util libraries should generally be completely agnostic to the domain you are working on. 

These are really business/domain rules that should sit with `services`, granted they could live in a module of functions inside there.

server/utils/utils.test.ts
andrewrlee: should these tests be updated then?

server/utils/utils.ts
SebNegahbanMoJ: Yes, `actualReleasedate` on the licence and `confirmedReleaseDate` in NOMIS are the same thing.
The purpose of this ticket is to read the licence dates if they exist, and if not to fall back to NOMIS. Otherwise we end up in a situation where the dates on the licence can end up being different to the dates we show in CVL, eg if the dates are deleted in NOMIS and we retain them for IS91 cases (as the licence generation reads from the CVL database)

server/utils/utils.ts
SebNegahbanMoJ: If we have consistent formatting enforced in the API, then I'm happy to remove the uncertainty! I only kept it as there appears to be uncertainty in the data received from NOMIS and I wasn't aware that we enforced structure in the API.

server/utils/utils.test.ts
SebNegahbanMoJ: Yes... yes they should 😅 
I changed a handful, didn't realise I'd missed so many!

server/routes/approvingLicences/handlers/approvalCases.test.ts
SebNegahbanMoJ: Looking at other examples, yes, but I can't seem to follow the code path that handles this? In the DB they're stored as yyyy-mm-dd, and I can't seem to see where the conversion happens on the 'get licence' side.

PR #483, opened 2023-01-16
Cvsl 647 case filter probation

mgreen-mojhub:
Add team filter when viewing team cases. Only allow cases for one team to be viewed.
If user has more than one team, force the user to select an available team from their allocated teams.


andrewrlee (approved):
LGTM

PR #488, opened 2023-01-24
refs cvsl-768

mgreen-mojhub:
Replace moment with date-fns and use the correct date format specified in the API schema. The correct date format for use with the CVL API is 'dd/MM/yyyy HH:mm:ss'


GurnankCheema (approved):
LGTM

PR #493, opened 2023-02-13
Removing the workaround to preserve dates for IS91 cases

SebNegahbanMoJ:
Despite working in some cases, it unfortunately resulted in the wrong dates being saved in most cases.
As such it has been determined that it's best to remove the workaround entirely so that it's clearer when IS91 case dates are incorrect (as they'll be missing entirely)


andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

justin-gardiner (approved):

PR #494, opened 2023-02-15
CVSL-820 - Adding username to API call when overriding licence status so it's visible in the audit log

SebNegahbanMoJ:
Fixes issue where CVL-staff usernames were not visible in the audit log when overriding a licence status.


andrewrlee (approved):
LGTM

PR #495, opened 2023-02-21
Adding batching for crns call to Delius to prevent errors

SebNegahbanMoJ:
The delius CRNs endpoint has a limit of being able to accept 512 CRNs at once. In some cases, we were sending more than this limit, so the call has been batched into chunks of 500.
If this still causes issue for any reason, the chunk size can be reduced.


andrewrlee (approved):
LGTM

PR #496, opened 2023-03-01
CVSL-720 - Swapping probation search for Community API

SebNegahbanMoJ:
The `OFFENDER_MANAGER_CHANGED` event handler is failing due to outdated data being returned by the Probation Offender Search API. From speaking to the Delius team, the long-term plan is to create a new API bespoke for CVL, but for the interim period, the Search API call is being swapped for a slower Community API call that will return the correct data instantly.


andrewrlee (approved):
LGTM

PR #497, opened 2023-03-02
Revert "CVSL-820 - Adding username to API call when overriding licence status so it's visible in the audit log"

SebNegahbanMoJ:
Reverts ministryofjustice/create-and-vary-a-licence#494

Unfortunately the changes made in CVSL-820 didn't have the intended effect, and with more pressing issues that need fixing, it seems best to remove the code for now (to make sure current functionality is not impacted) and fix it at a later date.


andrewrlee (approved):
LGTM

PR #504, opened 2023-04-03
CVSL-799

SebNegahbanMoJ:
CRO field | Change where we get CRO from

### DIFF
--- a/server/services/licenceService.ts
+++ b/server/services/licenceService.ts
@@ -124,7 +124,7 @@ export default class LicenceService {
       probationTeamDescription: responsibleOfficerDetails.team?.description,
       crn: deliusRecord.otherIds?.crn,
       pnc: deliusRecord.otherIds?.pncNumber,
-      cro: deliusRecord.otherIds?.croNumber || this.getCRONumber(prisonerNumber, user),
+      cro: deliusRecord.otherIds?.croNumber || (await this.getCRONumber(prisonerNumber, user)),
       standardLicenceConditions: [LicenceType.AP, LicenceType.AP_PSS].includes(licenceType)
         ? await this.conditionService.getStandardConditions(LicenceType.AP)
         : [],
--- a/server/services/licenceService.ts
+++ b/server/services/licenceService.ts
@@ -124,7 +124,7 @@ export default class LicenceService {
       probationTeamDescription: responsibleOfficerDetails.team?.description,
       crn: deliusRecord.otherIds?.crn,
       pnc: deliusRecord.otherIds?.pncNumber,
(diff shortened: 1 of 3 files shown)


server/services/licenceService.ts
```diff
@@ -149,6 +149,14 @@ export default class LicenceService {
     return this.licenceApiClient.createLicence(licence, user)
   }
 
+  private async getCRONumber(prisonerNumber: string, user: User): Promise<string> {
```
SebNegahbanMoJ, line 152: I think a name like `getCroNumberFromNomis` would be more appropriate, not only for TS-friendly casing but also to show it's coming from NOMIS

SebNegahbanMoJ: Also very small point that's more a preference than a rule, I'd prefer this to be towards the bottom of the file, to keep the `private` methods together and also to keep the regularly-used methods (like `getLicence`) closer to the top for easy of reading

server/services/licenceService.ts
```diff
@@ -149,6 +149,14 @@ export default class LicenceService {
...
+    const prisoners = await this.prisonerService.searchPrisonersByNomisIds([prisonerNumber], user)
+    if (prisoners != null) {
```
SebNegahbanMoJ, line 154: I don't think this will catch if `prisoners` is `undefined` rather than `null`.
I believe `if (prisoners)` should cover it.

server/services/licenceService.ts
```diff
@@ -149,6 +149,14 @@ export default class LicenceService {
...
+      return prisoners[0].croNumber
+    }
+    return ''
```
SebNegahbanMoJ, line 157: We've already talked about it off GitHub, but this should just be `return null`. We don't want to be setting it to an empty string if it's not there, or it will look odd in the service and on the printed PDFs

server/services/licenceService.ts
```diff
@@ -124,7 +124,7 @@ export default class LicenceService {
       probationTeamDescription: responsibleOfficerDetails.team?.description,
       crn: deliusRecord.otherIds?.crn,
       pnc: deliusRecord.otherIds?.pncNumber,
-      cro: deliusRecord.otherIds?.croNumber,
+      cro: deliusRecord.otherIds?.croNumber || (await this.getCRONumber(prisonerNumber, user)),
```
SebNegahbanMoJ, line 127: I don't think we need the outer brackets anymore, I think `await this.getCRONumber(prisonerNumber, user)` would suffice

andrewrlee (approved):
LGTM

PR #506, opened 2023-04-05
CVSL-893: Update messages on error pages

justin-gardiner:


assets/sass/components/_errors.scss
```diff
@@ -0,0 +1,3 @@
+.error-sub-heading {
+  padding-top: 40px;
+}
```
andrewrlee, line 3: Think its better to use the predefined sass helper functions rather than hardcoding pixel numbers, but most people tend to just use the [override spacing classes](https://design-system.service.gov.uk/styles/spacing/) and avoid separate css rules entirely

justin-gardiner: updated to use the override spacing classes

andrewrlee (approved):
LGTM

PR #510, opened 2023-04-11
Cvsl 978 cro | cro should be blank rather than saying no cro

SKMOJ:


server/services/licenceService.ts
```diff
@@ -124,7 +124,7 @@ export default class LicenceService {
       probationTeamDescription: responsibleOfficerDetails.team?.description,
       crn: deliusRecord.otherIds?.crn,
       pnc: deliusRecord.otherIds?.pncNumber,
-      cro: deliusRecord.otherIds?.croNumber,
+      cro: deliusRecord.otherIds?.croNumber || (await this.getCroNumberFromNomis(prisonerNumber, user)),
```
andrewrlee, line 127: good to have some tests for this?

SKMOJ: Done.

PR #516, opened 2023-04-17
Cvsl 978 cro should be blank rather than saying no cro3

SKMOJ:


server/services/licenceService.ts
```diff
@@ -639,8 +639,8 @@ export default class LicenceService {
   private async getCroNumberFromNomis(prisonerNumber: string, user: User): Promise<string> {
     const prisoners = await this.prisonerService.searchPrisonersByNomisIds([prisonerNumber], user)
     if (prisoners) {
-      return prisoners[0]?.croNumber
+      if (prisoners[0]?.croNumber !== undefined) return prisoners[0]?.croNumber
```
andrewrlee, line 642: you can make this a bit less verbose using `||`

`return prisoners[0]?.croNumber || ''`

andrewrlee (approved):
LGTM

PR #522, opened 2023-04-21
Bank holiday fix

SebNegahbanMoJ:
Fixing issue where bank holiday check was failing as it was comparing datetime rather than just date


justin-gardiner (approved):
LGTM

PR #524, opened 2023-04-24
CVSL-984. Added new fields to the offender page.

SKMOJ:


server/routes/support/handlers/offenderDetail.test.ts
```diff
@@ -122,6 +126,17 @@ describe('Route Handlers - Offender detail', () => {
           team: 'The A Team',
           telephone: '078929482994',
         },
+        license: {
+          led: 'Not found',
+          crn: undefined,
```
SebNegahbanMoJ, line 131: This returning `undefined` is a sign of a problem when the default value that's been set in the `let` is 'Not found'.

server/routes/support/handlers/offenderDetail.ts
```diff
@@ -18,7 +23,33 @@ export default class OffenderDetailRoutes {
...
+    const licenceSummary = await this.licenceServer.getLatestLicenceByNomisIdsAndStatus([nomsId], [], user)
 
+    let licencecrn = 'Not found'
+    let crd = 'Not found'
+    let releaseDate = 'Not found'
+    let sentenceStartDate = 'Not found'
+    let sed = 'Not found'
+    let led = 'Not found'
+    let tussd = 'Not found'
+    let licensetused = 'Not found'
```
SebNegahbanMoJ, line 35: Using a lot of `let`s is generally discouraged, as it created a lot of mutable variables. This reassignment could be handled with ternaries for instantiation instead

server/routes/support/handlers/offenderDetail.ts
```diff
@@ -18,7 +23,33 @@ export default class OffenderDetailRoutes {
...
+
+    if (licenceSummary != null) {
+      const licence = await this.licenceServer.getLicence(licenceSummary.licenceId.toString(), user)
+
+      licencecrn = licence.crn
+      crd = licence.conditionalReleaseDate
+      releaseDate = licence.actualReleaseDate
```
SebNegahbanMoJ, line 42: `releaseDate` is not a particularly accurate name for this, as CRD is also a release date. We tend to refer to this date by `ard`

server/routes/support/handlers/offenderDetail.ts
```diff
@@ -18,7 +23,33 @@ export default class OffenderDetailRoutes {
...
+      sentenceStartDate = licence.sentenceStartDate
+      sed = licence.sentenceEndDate
+      led = licence.licenceExpiryDate
+      tussd = licence.topupSupervisionStartDate
+        ? moment(licence.topupSupervisionStartDate).format('DD MMM YYYY')
```
SebNegahbanMoJ, line 47: This has the DD MMM YYYY format, but if you tried to access it I imagine it will likely fail and raise an error, as the dates stored on the licence are in `dd/mm/yyyy` format, and moment expects them in `mm/dd/yyyy` format. This means that something like `13/1/1970` would be evaluated as 1st of the 13th month, raising an error.
You need to specify `moment(datestring, 'DD/MM/YYYY')` for it to parse correctly.

server/routes/support/handlers/offenderDetail.ts
```diff
@@ -18,7 +23,33 @@ export default class OffenderDetailRoutes {
...
+        : 'Not found'
+      licensetused = licence.topupSupervisionExpiryDate
```
SebNegahbanMoJ, line 49: I think an alternate approach here would mean that we don't have to double-up variable names, so there's no need for `tused` and `licencetused`.
Even if we were to keep it, it would probably be better to be called `licenceTused` to follow camelcase

server/routes/support/handlers/offenderDetail.ts
```diff
@@ -73,6 +104,16 @@ export default class OffenderDetailRoutes {
         pdu: probationPractitioner?.team?.borough?.description,
         region: probationPractitioner?.probationArea?.description,
       },
+      license: {
```
SebNegahbanMoJ, line 107: This should be `licence`, which will affect all other references to the object

server/views/pages/support/offenderDetail.njk
```diff
@@ -92,7 +92,7 @@
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
             <table class="govuk-table">
-                <caption class="govuk-table__caption govuk-table__caption--m">Sentence details</caption>
+                <caption class="govuk-table__caption govuk-table__caption--m">Sentence details<p>This information comes from NOMIS.</caption>
```
SebNegahbanMoJ, line 95: I don't think the `<p>` is meant to be here? If it is, it should be outside of the <caption></caption> and should have a closing tag.

server/views/pages/support/offenderDetail.njk
```diff
@@ -143,6 +143,48 @@
...
     </div>
 
+    <details class="govuk-details" data-module="govuk-details">
+      <summary class="govuk-details__summary">
+        <span class="govuk-details__summary-text">
+          View dates stored by Create and vary a licence
+        </span>
+      </summary>
+      <div class="govuk-details__text">
+<table class="govuk-table">
```
SebNegahbanMoJ, line 153: The tab-spacing of the code in this table seems a bit off, as does the </div> at the end

server/views/pages/support/offenderDetail.njk
```diff
@@ -143,6 +143,48 @@
...
+                <tbody class="govuk-table__body">
+                    <tr class="govuk-table__row">
+                        <th scope="row" class="govuk-table__header">CRD</th>
+                        <td class="govuk-table__cell">{{ license.licencecrn }}</td>
```
SebNegahbanMoJ, line 157: This is using CRN, which is an identifier from delius, not a date
It should be using CRD

SebNegahbanMoJ:
I think a tidier solution for this would be to make a method that builds the CVL licence dates as an object, and passes that directly into the render call. Then in the `res.render` call, replace the current `license: { <building object> }` with a call to the method (eg `licence: methodCall()`)
That would remove the need to have shared var names, as well as the need to have `let`s instead of `const`s, as you can just escape out of the method with a `null` return if there's no licence for the offender, and handle the page rendering differently. You'd need to speak to Steph and Tom for how they'd like to handle it if there's no licence, whether we just don't show the table at all, or replace it with a message is up to them.

I've drafted up an example of what the method could look like, and we can go through it if you'd like to, but I'm happy to let you crack on with it your way.

server/routes/support/handlers/offenderDetail.test.ts
SKMOJ: Done.

server/routes/support/handlers/offenderDetail.ts
SKMOJ: actualReleaseDate

server/routes/support/handlers/offenderDetail.ts
SKMOJ: Done.

server/views/pages/support/offenderDetail.njk
```diff
@@ -26,7 +26,7 @@
                         <td class="govuk-table__cell">{{ prisonerDetail.dob }}</td>
                     </tr>
                     <tr class="govuk-table__row">
-                        <th scope="row" class="govuk-table__header govuk-!-font-weight-bold">Prisoner number</th>
+                        <th scope="row" class="govuk-table__header govuk-!-font-weight-bold">Prison number</th>
```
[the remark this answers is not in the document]
SKMOJ: As per the request from Tom and the offender detail page design.
https://docs.google.com/document/d/1AR0AnzF0i4vNkpqUf2Yn14IGm2ZYrg-cdxOovWXiiPM/edit

server/routes/support/handlers/offenderDetail.ts
```diff
@@ -18,7 +23,33 @@ export default class OffenderDetailRoutes {
...
+    let sed = 'Not found'
+    let led = 'Not found'
+    let tussd = 'Not found'
+    let licensetused = 'Not found'
+
+    if (licenceSummary != null) {
+      const licence = await this.licenceServer.getLicence(licenceSummary.licenceId.toString(), user)
+
+      licencecrn = licence.crn
+      crd = licence.conditionalReleaseDate
```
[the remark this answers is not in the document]
SKMOJ: Done.

server/routes/support/handlers/offenderDetail.test.ts
SebNegahbanMoJ: It still seems to be set to `undefined`. I don't thin CRN is relevant for this section regardless, as it's the Delius identifier rather than a date

server/routes/support/handlers/offenderDetail.test.ts
```diff
@@ -317,6 +343,17 @@ describe('Route Handlers - Offender detail', () => {
...
+        led: 'Not found',
+        crn: undefined,
+        crd: 'Not found',
+        licencecrn: 'Not found',
+        licensetused: 'Not found',
+        releaseDate: 'Not found',
+        sed: 'Not found',
+        sentenceStartDate: 'Not found',
+        tussd: 'Not found',
+      },
```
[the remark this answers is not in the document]
SebNegahbanMoJ: Just bumping this comment

server/routes/support/handlers/offenderDetail.ts
```diff
@@ -18,7 +23,43 @@ export default class OffenderDetailRoutes {
       ? await this.communityService.getStaffDetailByStaffCode(probationPractitioner?.staff.code)
       : undefined
     const hdcStatus = _.head(await this.prisonerService.getHdcStatuses([prisonerDetail], user))
+    const licenceSummary = await this.licenceServer.getLatestLicenceByNomisIdsAndStatus([nomsId], [], user)
 
+    let licenceFromSummary
+    let licencecrn = 'Not found'
```
SebNegahbanMoJ, line 29: I don't think we need CRN at all here, as it's a Delius identifier rather than a licence date.

server/routes/support/handlers/offenderDetail.ts
```diff
@@ -18,7 +23,43 @@ export default class OffenderDetailRoutes {
...
+    let crd = 'Not found'
+    let actualReleaseDate = 'Not found'
+    let sentenceStartDate = 'Not found'
+    let sed = 'Not found'
+    let led = 'Not found'
+    let tussd = 'Not found'
+    let licenceTused = 'Not found'
```
SebNegahbanMoJ, line 36: We're still relying on a lot of `let`s here

server/routes/support/handlers/offenderDetail.ts
```diff
@@ -18,7 +23,43 @@ export default class OffenderDetailRoutes {
...
+
+    if (licenceSummary != null) {
+      licenceFromSummary = await this.licenceServer.getLicence(licenceSummary.licenceId.toString(), user)
+    }
+    if (licenceFromSummary != null) {
+      licencecrn = licenceFromSummary.crn ? licenceFromSummary.crn : 'Not found'
+      crd = licenceFromSummary.conditionalReleaseDate
+        ? moment(licenceFromSummary.conditionalReleaseDate, 'DD/MM/YYYY').format('DD MMM YYYY')
+        : 'Not found'
+      actualReleaseDate = licenceFromSummary.actualReleaseDate ? licenceFromSummary.actualReleaseDate : 'Not found'
```
SebNegahbanMoJ, line 46: This doesn't seem to have the date formatting on it

server/routes/support/handlers/offenderDetail.ts
SebNegahbanMoJ: It still appears to be called `license`

server/views/pages/support/offenderDetail.njk
SebNegahbanMoJ: Just bumping this comment

server/routes/support/handlers/offenderDetail.ts
SKMOJ: Done.

server/routes/support/handlers/offenderDetail.ts
SKMOJ: Done.

server/routes/support/handlers/offenderDetail.ts
SKMOJ: Done.

server/routes/support/handlers/offenderDetail.ts
```diff
@@ -73,6 +88,47 @@ export default class OffenderDetailRoutes {
...
+        tussd: 'Not found',
+        tused: 'Not found',
+      }
+    }
+
+    const licence = await this.licenceServer.getLicence(licenceSummary.licenceId.toString(), user)
+
+    return {
+      crd: licence.conditionalReleaseDate
+        ? moment(licence.conditionalReleaseDate, 'DD/MM/YYYY').format('DD MMM YYYY')
```
andrewrlee, line 114: could extract out a helper function and tidy this up a bit?

```
ard: formatLicenceDate(licence.actualReleaseDate),
ssd: formatLicenceDate(licence.sentenceStartDate),
```

SKMOJ: Done.

SebNegahbanMoJ (approved):

PR #527, opened 2023-04-27
Cvsl 767 layout change  | primary and secondary buttons across the whole service

nishanthmahasamudram:


server/routes/creatingLicences/index.ts
```diff
@@ -170,9 +170,9 @@ export default function Index({
     additionalLicenceConditionUploads.POST
   )
 
-  post('/id/:licenceId/additional-licence-types/condition/delete', additionalLicenceTypesHandler.DELETE)
+  get('/id/:licenceId/additional-licence-types/condition/:conditionCode/delete', additionalLicenceTypesHandler.DELETE)
 
-  post(
+  get(
```
andrewrlee, line 175: is the difference between these two just that one is called from the file upload page and the other by the normal conditions page? 

presumably those are the only places they are called from?

nishanthmahasamudram: Yes, it is called only from those two places.
I tested by adding and deleting all conditions in local env.

nishanthmahasamudram: additionalLicenceConditionInput.ts is for deleting all conditions except file upload condition, It is deleted by using additionalLicenceTypesHandler.ts

nishanthmahasamudram: As other conditions are deleted using **conditionId** while file upload condition is deleted using **conditionCode**.

andrewrlee (approved):
LGTM

PR #528, opened 2023-05-02
CVSL-994-App Insights | Add user to request in App insights

nishanthmahasamudram:


server/utils/azureAppInsights.test.ts
```diff
@@ -47,6 +47,8 @@ describe('azureAppInsights', () => {
         displayName: 'A User',
         nomisStaffId: 'N123',
         prisonCaseload: ['MDI'],
+        username: 'N123',
+        activeCaseLoadId: ['MDI'],
```
andrewrlee, line 51: this isn't really testing the change. 

The test previously checked that you were just copying values across.. we now are doing some additional aliasing

If you remove these fields from here and add them after line 59 then that would be better 
(it's a bit unrealistic at the moment as it makes it seem that username and activeCaseload are part of the user object when they aren't)

andrewrlee (approved):
LGTM

PR #533, opened 2023-05-09
CVSL-1022 Add better caching of assets

andrewrlee:
Currently the assets are only set to cache for 20 seconds. This should drastically reduce the number of requests to the server and also make browsing logs easier to see.

In production the css cache will be linked to the deployment version rather than pod start up time - currently the browser caches a version for each pod

There's also another little Helmet/CSP fix copied across from the template project to prevent user's being stuck on a page if they submit a page and the app detects their session has timed out. 
Previously CSP would have kicked in and an error would have been presented in terminal and the user would be stranded. Now the user will be redirected to auth.

### DIFF
--- a/server/applicationVersion.ts
+++ b/server/applicationVersion.ts
@@ -2,8 +2,8 @@
 import fs from 'fs'
 
 const packageData = JSON.parse(fs.readFileSync('./package.json').toString())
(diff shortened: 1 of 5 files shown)


justin-gardiner (approved):
LGTM

PR #536, opened 2023-05-11
CVSL-1022 Fix version reporting

andrewrlee:
The earlier caching improvement change revealed that the version reporting was broken. This previously relied on running a shell script to generate a file with a json payload in it. The code that read this file to extract out the version for the cache improvement and also setting the application version in app insights was looking in the wrong location

There was another location that looked up the file and read in the details for the health endpoint which was looking in the right place

This change moves to reading the version and git reference into an env var in the docker file instead, which means we can centralise how this info is made available and remove the additional file management

This should be a safe fix as the build info file was previously being generated from the docker build anyway - so the file should be available


server/applicationInfo.ts
```diff
@@ -0,0 +1,12 @@
+import fs from 'fs'
+import path from 'path'
+import config from './config'
+
+const packageJson =
+  process.env.NODE_ENV !== 'test' ? path.join(__dirname, '../../package.json') : path.join(__dirname, '../package.json')
```
andrewrlee, line 6: not very happy with the check to see if we're running a test or not. 

Problem is that ts-jest transpiles js on the fly and the running dir is the ts source folder, whereas when running the application we're one layer down (in the `dist` folder)

Dockerfile
```diff
@@ -16,6 +16,8 @@ WORKDIR /app
 
 # Cache breaking
 ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
+ENV GIT_REF ${GIT_REF:-xxxxxxxxxxxxxxxxxxx}
```
andrewrlee, line 19: bake in the GIT_REF as an env var when building the image. 

This is available at this stage as seen by the old code that used it for writing the build-info json file.

server/config.ts
```diff
@@ -30,6 +30,8 @@ export interface ApiConfig {
 }
 
 export default {
+  buildNumber: get('BUILD_NUMBER', '1_0_0', requiredInProduction),
+  gitRef: get('GIT_REF', 'xxxxxxxxxxxxxxxxxxx', requiredInProduction),
```
andrewrlee, line 34: blow up if env var not set when running the image in docker, default when running locally

server/services/healthCheck.ts
```diff
@@ -25,25 +26,18 @@ function service(name: string, url: string, agentConfig: AgentConfig): HealthChe
...
 }
 
-function getBuild() {
-  try {
-    // eslint-disable-next-line import/no-unresolved,global-require
-    return require('../../build-info.json')
-  } catch (ex) {
-    return null
-  }
-}
```
andrewrlee, line 45: this way of loading the file worked when running in docker

bin/record-build-info
```diff
@@ -1,21 +0,0 @@
-#!/usr/bin/env node
-
-const fs = require('fs')
-const path = require('path')
-
-const target = path.resolve(__dirname, '../build-info.json')
```
andrewrlee, line 6: this way of loading the file failed to work when running in docker

server/utils/azureAppInsights.ts
```diff
@@ -1,20 +1,16 @@
...
 function defaultName(): string {
-  const {
-    packageData: { name },
-  } = applicationVersion
-  return name
+  return applicationInfo.applicationName
 }
 
 function version(): string {
-  const { buildNumber } = applicationVersion
```
andrewrlee, line 16: this was previously passing through a defaulted version `1_0_0`

SebNegahbanMoJ (approved):

PR #537, opened 2023-05-11
Cvsl 1000 change variation status to status and surface active status

nishanthmahasamudram:
We introduced a label "Active" in status column.
This is for initial licences which have been approved and made active (by the system on CRD or manually by our team) and also licence variations which have been made active.

Also, renamed the column header from "Variation status" to just "Status".


server/views/pages/vary/caseload.njk
```diff
@@ -55,7 +55,7 @@
                 id: 'licence-status-' + loop.index,
                 "data-sort-value": offender.licenceStatus | getStatusOrder
             },
-            html: statusBadge({ status: offender.licenceStatus, size: 'small' }, statusConfig) if offender.licenceStatus !== 'ACTIVE'
```
nishanthmahasamudram, line 58: Introduced the Active status.
This is for initial licences which have been approved and made active (by the system on CRD or manually by our team) and also licence variations which have been made active.

SebNegahbanMoJ (approved):
LGTM - need to make sure that the colour of the `ACTIVE` pill is correct in testing

PR #539, opened 2023-05-12
CVSL-759: When confirming a licence variation action check if there is already an existing variation for the licence and if so use that instead of creating a new variation.

justin-gardiner:


server/services/licenceService.ts
```diff
@@ -583,6 +583,18 @@ export default class LicenceService {
     return this.licenceApiClient.getLicenceById(licence.variationOf.toString(), user)
   }
 
+  async getLicenceVariations(nomisId: string): Promise<LicenceSummary[]> {
```
SebNegahbanMoJ, line 586: I think this name could be easily misconstrued to be 'get all the licences that are variations of the given licence' (ie including those that are active).
Perhaps a more fitting name would be something like `getIncompleteLicenceVariations`? I'm hesitant to say 'in progress' or 'non-active' as those are licence statuses in themselves, but I think 'incomplete' covered both not active and not inactive.

server/routes/varyingLicences/handlers/confirmVaryAction.ts
```diff
@@ -12,13 +13,19 @@ export default class ConfirmVaryActionRoutes {
...
 
     if (answer === YesOrNo.NO) {
       return res.redirect(`/licence/vary/id/${licenceId}/view-active`)
     }
 
-    const newLicence = await this.licenceService.createVariation(licenceId, user)
+    let newLicence: LicenceSummary
+    const licenceVariations = await this.licenceService.getLicenceVariations(licence.nomsId)
+    if (licenceVariations?.length > 0) {
+      ;[newLicence] = licenceVariations
```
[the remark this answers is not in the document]
justin-gardiner: the semi-colon is for eslint, the build fails without it. It's because we don't terminate lines with semi-colons so the array at the start of the line could be a continuation of the previous line without the semi-colon. See https://eslint.org/docs/latest/rules/semi

Anyway I've updated to use _.head as you suggested and it's no longer needed :)

server/services/licenceService.ts
justin-gardiner: good point, of course other status can be variations of a licence. Updated.

SebNegahbanMoJ (approved):

andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

PR #543, opened 2023-05-16
CVSL-996: Add recall status to support offender detail page.

justin-gardiner:


server/routes/support/handlers/offenderDetail.ts
```diff
@@ -43,6 +43,11 @@ export default class OffenderDetailRoutes {
     const licenceExpiryDate = this.formatNomisDate(prisonerDetail.licenceExpiryDate)
     const paroleEligibilityDate = this.formatNomisDate(prisonerDetail.paroleEligibilityDate)
 
+    let isRecall
```
andrewrlee, line 46: is it possible for recall to be undefined? looks [like it's a primitive?](https://github.com/ministryofjustice/prison-api/blob/main/src/main/java/uk/gov/justice/hmpps/prison/api/model/RecallCalc.java#L41)?

justin-gardiner: from that it looks like it can't be so I've updated the code. I was going off the type definition which says it is optional

justin-gardiner: looks like it can't be from that, I was going from the type definition which has it as optional,

andrewrlee (approved):
LGTM

PR #545, opened 2023-05-17
CVSL-1034 Update redis library

andrewrlee:


server/data/index.ts
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable import/first */
```
andrewrlee, line 1: this is more in line with the template project now - lazily creates redis connection which means that the tests don't continually try to create connections with redis

SebNegahbanMoJ (approved):

PR #551, opened 2023-05-23
CVSL-1024-Bespoke condition cannot delete just one easily

nishanthmahasamudram:


server/views/pages/create/bespokeConditions.njk
```diff
@@ -40,7 +40,7 @@
                     <div class="moj-button-action">
                         {{ govukButton({
                             text: 'Add another bespoke condition',
-                            classes: 'govuk-button--secondary moj-add-another__add-button govuk-!-margin-bottom-4',
+                            classes: 'govuk-button--secondary moj-add-another__add-button custom-add-button govuk-!-margin-bottom-7',
```
andrewrlee, line 43: don't think this custom-add-button class is used?

assets/js/bespoke.js
```diff
@@ -0,0 +1,9 @@
+window.onclick = function () {
+  var $addAnothers = document.querySelectorAll('[data-module="moj-add-another"]')
+  const addAnother = new MOJFrontend.AddAnother($addAnothers)
+  const buttons = document.getElementsByClassName('delete-condition-button')
+  if (buttons.length) {
```
andrewrlee, line 5: do you need to do this check ? can you not just rely on `getItems` here?

assets/js/bespoke.js
```diff
@@ -0,0 +1,9 @@
+window.onclick = function () {
```
andrewrlee, line 1: seems a bit hacky to trigger this on every button click on the page.

assets/js/bespoke.js
```diff
@@ -0,0 +1,9 @@
...
+  var $addAnothers = document.querySelectorAll('[data-module="moj-add-another"]')
+  const addAnother = new MOJFrontend.AddAnother($addAnothers)
+  const buttons = document.getElementsByClassName('delete-condition-button')
+  if (buttons.length) {
+    var button = buttons[0]
+    button.textContent = addAnother.getItems().length > 1 ? 'Delete these conditions' : 'Delete this condition'
+  }
+}
```
andrewrlee, line 9: would be good to add a cypress test for this if possible as this is tightly coupled to the internal implementation of the add another component (ideally we'd ask them to add some hooks into their API).

A cypress test would help ensure it doesn't break during updates to the library - I think it would be very easy to miss, during a manual regression test, if this behaviour was broken

server/routes/creatingLicences/handlers/bespokeConditions.ts
```diff
@@ -29,4 +30,11 @@ export default class BespokeConditionsRoutes {
 
     return res.redirect(`/licence/create/id/${licenceId}/check-your-answers`)
   }
+
+  DELETE = async (req: Request, res: Response): Promise<void> => {
```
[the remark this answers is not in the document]
nishanthmahasamudram: test added.

server/views/pages/create/bespokeConditions.njk
nishanthmahasamudram: removed unused class name.

assets/js/bespoke.js
nishanthmahasamudram: Yes, we need it.
It is not remove button count, it is to get the delete condition button to rename the text content.

assets/js/bespoke.js
nishanthmahasamudram: we have total four buttons on this page.
1) Remove - trigger is needed
2) Add - trigger is needed
3) Continue and 4) Delete -> This will redirect any way to check answers page so there won't be much impact

nishanthmahasamudram: changed the implementation using overwrite method of existing class.

assets/js/bespoke.js
```diff
@@ -0,0 +1,24 @@
...
+    oldButtonClick.call(this, e)
+    changeButtonTextContent()
+  }
+
+  MOJFrontend.AddAnother.prototype.onRemoveButtonClick = function (e) {
+    oldRemoveClick.call(this, e)
+    changeButtonTextContent()
+  }
+
+  var $addAnothers = document.querySelectorAll('[data-module="moj-custom-add-another"]')
```
andrewrlee, line 14: can you change these `var`s to `const`s?

integration_tests/pages/bespokeConditions.ts
```diff
@@ -21,6 +21,16 @@ export default class BespokeConditionsPage extends Page {
     return this
   }
 
+  checkDeleteThisCondition = (): BespokeConditionsPage => {
+    cy.get('.delete-condition-button').should('have.text'.trim().replace(/[\r\n]+/g, ''), '\n  Delete this condition\n')
+    return this
+  }
+
+  checkDeleteTheseConditions = (): BespokeConditionsPage => {
+    cy.get('.delete-condition-button').should('have.text'.trim().replace(/[\r\n]+/g, ''), 'Delete these conditions')
```
andrewrlee, line 30: not sure this is right?  
`'have.text'.trim().replace(/[\r\n]+/g, '')`

you're trimming the text `have.text`

nishanthmahasamudram: have.text will get the text dynamically form button and then we are comparing.
I am getting some spaces and new lines so I did it.

nishanthmahasamudram: It is just to trim the dynamic string.

nishanthmahasamudram: removed it and replaced have.text to contain.text

andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

PR #552, opened 2023-05-23
Test1

andrewrlee:


feature.env
```diff
@@ -13,3 +13,6 @@ API_CLIENT_SECRET=clientsecret
 SYSTEM_CLIENT_ID=clientid
 SYSTEM_CLIENT_SECRET=clientsecret
 SQS_POLLING_WAIT_TIME_MS=2000
+
+AWS_ACCESS_KEY_ID=foo
+AWS_SECRET_ACCESS_KEY=bar
```
andrewrlee, line 18: There is a chain of credential providers that is checked when the server starts up and starts to consume from the queues. 

In production, the "web token credential provider" checks for the `AWS_WEB_IDENTITY_TOKEN_FILE` and then retrieves credential from that. 

When running integration tests run we do not have that mechanism available to us so we need to set these vars to trigger a different credential provider. If these aren't set then you get a message complaining no credentials have been provided. 

In integration tests we use local stack which does validate the credentials so these values are arbitrary

PR #553, opened 2023-05-25
CVSL-1046 using custom service account

andrewrlee:
This service account has the permission to retrieve credentials to communicate with the queue


justin-gardiner (approved):
LGTM

PR #557, opened 2023-06-05
Cvsl 465 dont allow pp to vary licence conditions if led less than current date less than tused

nishanthmahasamudram:


server/routes/varyingLicences/handlers/viewActiveLicence.ts
```diff
@@ -18,10 +22,21 @@ export default class ViewActiveLicenceRoutes {
     const { conditionsWithUploads, additionalConditions } = this.conditionService.additionalConditionsCollection(
       licence.additionalLicenceConditions
     )
+    parentOrSelfAdditionalConditions = additionalConditions
```
andrewrlee, line 25: should this be filtered to only show the AP conditions belonging to the parent? 

If so then maybe `apConditions` might be better?
  
if it's not just `apConditions` then calling it `previousLicenceConditions` might be more meaningful?

server/utils/utils.ts
```diff
@@ -206,6 +206,19 @@ const isPassedArdOrCrd = (licence: LicenceSummary, prisoner: Prisoner | PrisonAp
   return false
 }
 
+const isInPssPeriod = (licence: Licence): boolean => {
+  const { licenceExpiryDate, topupSupervisionExpiryDate } = licence
```
andrewrlee, line 210: I don't think domain logic belongs in utils but it feels a bit like I'm fighting a losing battle :-D 

We should be looking at moving all stuff like this down in to the API layer anyway but I'd suggest leaving this here for now..

Could do with a test though?

andrewrlee, line 210: I don't think domain logic belongs in utils but it feels a bit like I'm fighting a losing battle :-D 

We should be looking at moving all stuff like this down in to the API layer anyway but I'd suggest leaving this here for now..

Could do with a test though?

server/routes/creatingLicences/handlers/checkAnswers.test.ts
```diff
@@ -90,8 +92,39 @@ describe('Route Handlers - Create Licence - Check Answers', () => {
...
+            deliusStaffIdentifier: 999,
+          },
+        },
+      } as unknown as Response
+
+      await handler.GET(req, res)
+
+      expect(res.render).toHaveBeenCalledWith('pages/create/checkAnswers', {
+        inPssPeriod: true,
+        parentOrSelfAdditionalConditions: [],
```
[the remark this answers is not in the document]
nishanthmahasamudram: Yes, it is.

server/utils/utils.ts
nishanthmahasamudram: I added the test case specific for utils function as well.

server/views/pages/create/checkAnswers.njk
```diff
@@ -28,15 +28,15 @@
...
+                    <h1 class="govuk-heading-l">Vary post sentence supervision order details</h1>
                 {% elseif licence.typeCode === 'PSS' %}
                     <h1 class="govuk-heading-l">Vary post sentence supervision order details</h1>
                 {% endif %}
             {% else %}
                 {% if licence.typeCode === 'AP' %}
                     <h1 class="govuk-heading-l">Check licence details</h1>
                 {% elseif licence.typeCode === 'AP_PSS' %}
-                    <h1 class="govuk-heading-l">Check licence and post sentence supervision order details</h1>
+                    <h1 class="govuk-heading-l">Check post sentence supervision order details</h1>
```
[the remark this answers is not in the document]
nishanthmahasamudram: ok I will check with Tom and update.

server/routes/creatingLicences/handlers/checkAnswers.ts
```diff
@@ -30,8 +32,23 @@ export default class CheckAnswersRoutes {
...
+    if (inPssPeriod) {
+      const { additionalConditions: parentAdditionalConditions } = this.conditionService.additionalConditionsCollection(
+        (await this.licenceService.getParentLicenceOrSelf(licence.id, user))?.additionalLicenceConditions
+      )
+      parentOrSelfAdditionalAPConditions = parentAdditionalConditions
+    }
+
+    res.render('pages/create/checkAnswers', {
+      parentOrSelfAdditionalAPConditions,
+      conditionsWithUploads,
```
SebNegahbanMoJ, line 48: I _think_ we need to get the `conditionsWithUploads` from the parent licence too for the `inPssPeriod` situation, in the same way you have with `additionalConditions` on line 35.
In other words, on line 40 it'll need to be `{ conditionsWithUploads, additionalConditions }` (the same as line 32), and you'll need to create a `parentOrSelfConditionsWithUploads` on line 44, which you pass in instead of `conditionsWithUploads` on line 48.

Alternatively rather than creating the `parentOrSelf-` versions of the vars, line 32 could be changed to a `let` and simply reassign them in the `inPssPeriod` condition if necessary.
Not sure if `let` is problematic to use in this case or not, @andrewrlee what do you think?

server/views/partials/additionalPssConditionsSummary.njk
```diff
@@ -17,7 +17,7 @@
 
     <div class="govuk-grid-row check-answers-header">
         <div class="govuk-grid-column-three-quarters">
-            <h2 id="additional-pss-conditions-heading" class="govuk-heading-m">{{ title or 'Additional post sentence supervision requirements' }}</h2>
+            <h2 id="additional-pss-conditions-heading" class="govuk-heading-m">{{ title or 'Additional post sentence requirements details' }}</h2>
```
SebNegahbanMoJ, line 20: If this is the wording in the prototype, I'd double-check this with Tom as PSS stands for Post Sentence Supervision, and removing the 'Supervision' seems odd

server/routes/varyingLicences/handlers/viewActiveLicence.ts
nishanthmahasamudram: refactored the code so removed extra variable

server/routes/creatingLicences/handlers/checkAnswers.ts
nishanthmahasamudram: ok. If we need conditionsWithUploads from the parent then we can completely refactor it.
I removed the parentOrSelf- variable and i am reassigning to the same destructured object.

SebNegahbanMoJ: I've just realised I didn't actually explain why
Basically, one of the conditions (the one about exclusion zones were you upload a map of an area that the offender cannot enter) works a bit differently to the others. The requirement was to add multiple exclusion zones (this is what people mean if you hear the acronym 'MEZ') to one condition, but the with the templating system that we use this proved to be very difficult. The solution that was found was to duplicate the licence condition for each map (so an offender can have multiple entries of the same condition), then when rendering, combine all of the instances of the MEZ condition into one (this probably makes more sense if you see an example of how it looks!)

Currently, only the exclusion zone condition has uploads, so whenever you see something like `conditionsWithUploads`, it basically means `instances of the exclusion zone condition`. This means that by not including it here, the result would be that the exclusion zone condition would never appear in the new 'expired conditions' section that you've added, which is very easy to miss in testing if you happen to just not select that condition during your tests!

server/views/partials/additionalPssConditionsSummary.njk
nishanthmahasamudram: Ok I will check with Tom.

server/views/pages/create/checkAnswers.njk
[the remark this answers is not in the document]
nishanthmahasamudram: Yes, Seb you are right we need to change the heading only if it is in PSS period, got confirmation from Tom and I did changes accordingly

server/views/partials/additionalPssConditionsSummary.njk
nishanthmahasamudram: image
I checked with Tom and the heading should be as in prototype itseems.

server/routes/varyingLicences/handlers/viewActiveLicence.ts
```diff
@@ -18,10 +22,21 @@ export default class ViewActiveLicenceRoutes {
...
+    const inPssPeriod = isInPssPeriod(licence)
+
+    if (inPssPeriod) {
+      const { additionalConditions: parentAdditionalConditions } = this.conditionService.additionalConditionsCollection(
+        (await this.licenceService.getParentLicenceOrSelf(licence.id, user))?.additionalLicenceConditions
+      )
+      parentOrSelfAdditionalAPConditions = parentAdditionalConditions
+    }
 
     return res.render('pages/vary/viewActive', {
```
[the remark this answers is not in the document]
nishanthmahasamudram: Yes, it is in the demo.

server/routes/creatingLicences/handlers/checkAnswers.test.ts
```diff
@@ -113,6 +116,59 @@ describe('Route Handlers - Create Licence - Check Answers', () => {
       expect(res.render).toHaveBeenCalledWith('pages/create/checkAnswers', {
         additionalConditions: [],
         conditionsWithUploads: [],
+        isInPssPeriod: false,
+        backLink: req.session.returnToCase,
+      })
+      expect(licenceService.recordAuditEvent).toHaveBeenCalled()
+    })
+
+    it('should render view, record audit event (not owner) and PSS period should be true', async () => {
```
andrewrlee, line 125: not sure you need this test anymore? 

pssPeriod presumably will be read from the licence directly rather than calculated so don't need a specific test for checking its value?

server/routes/creatingLicences/handlers/checkAnswers.ts
```diff
@@ -27,11 +28,22 @@ export default class CheckAnswersRoutes {
...
     }
 
-    const { conditionsWithUploads, additionalConditions } = this.conditionService.additionalConditionsCollection(
+    let { conditionsWithUploads, additionalConditions } = this.conditionService.additionalConditionsCollection(
       licence.additionalLicenceConditions
     )
 
-    res.render('pages/create/checkAnswers', { additionalConditions, conditionsWithUploads, backLink })
+    if (isInPssPeriod) {
+      ;({ conditionsWithUploads, additionalConditions } = this.conditionService.additionalConditionsCollection(
```
andrewrlee, line 36: you can avoid the mutable reference (`let`) by restructuring? 
```ts
const conditionsToDisplay  = !isInPssPeriod 
  ?  licence.additionalLicenceConditions
  :  (await this.licenceService.getParentLicenceOrSelf(licence.id.toString(), user))?.additionalLicenceConditions
```

andrewrlee: You're tests aren't exercising this conditional though

server/routes/varyingLicences/handlers/viewActiveLicence.ts
```diff
@@ -15,13 +17,20 @@ export default class ViewActiveLicenceRoutes {
...
-    const { conditionsWithUploads, additionalConditions } = this.conditionService.additionalConditionsCollection(
+    let { conditionsWithUploads, additionalConditions } = this.conditionService.additionalConditionsCollection(
       licence.additionalLicenceConditions
     )
 
+    if (isInPssPeriod) {
+      ;({ conditionsWithUploads, additionalConditions } = this.conditionService.additionalConditionsCollection(
+        (await this.licenceService.getParentLicenceOrSelf(licence.id.toString(), user))?.additionalLicenceConditions
+      ))
+    }
```
andrewrlee, line 28: as you're doing this twice and it's a bit complicated - maybe better to push this down in to the condition service and give it a nice name?

server/routes/creatingLicences/handlers/checkAnswers.ts
nishanthmahasamudram: added tests to check this condition

server/routes/creatingLicences/handlers/checkAnswers.test.ts
```diff
@@ -113,6 +125,59 @@ describe('Route Handlers - Create Licence - Check Answers', () => {
...
+
+      await handler.GET(req, res)
+
+      expect(res.render).toHaveBeenCalledWith('pages/create/checkAnswers', {
+        additionalConditions: [],
+        conditionsWithUploads: [],
+        isInPssPeriod: false,
+        backLink: req.session.returnToCase,
+      })
+      res = {
```
andrewrlee, line 161: This is testing 2 different scenarios so should really be two different tests

server/data/licenceApiClient.ts
```diff
@@ -446,4 +446,13 @@ export default class LicenceApiClient extends RestClient {
       data: offenderDetails,
     })
   }
+
+  async getParentLicenceOrSelf(licenceId: string, user: User): Promise<Licence> {
```
justin-gardiner, line 450: Is there a reason this was moved here from LicenceService?

nishanthmahasamudram: Yes, In order to remove circular dependency move getParentOrSelfAdditionalLicenceConditions this functionality to conditionalService we decided to move(Andy suggested)

justin-gardiner (approved):
lgtm

PR #562, opened 2023-06-09
CVSL-1041-Not setting the username correctly in app insights

nishanthmahasamudram:


SebNegahbanMoJ (dismissed):
Looks reasonable to me!

server/@types/CvlUserDetails.ts
```diff
@@ -5,6 +5,9 @@ export default class CvlUserDetails {
   // e.g. smith
   lastName?: string
 
+  // e.g. John_Smith
+  username?: string
```
andrewrlee, line 9: I can't see where this is being set on the CVL User Details type?

nishanthmahasamudram: removed the username for cvluserdetail class

PR #566, opened 2023-06-19
Adding COM details stored in CVL to support page

SebNegahbanMoJ:
Allows easier recognition of cases where CVL has an incorrect COM assigned to a case


justin-gardiner (approved):
lgtm

PR #571, opened 2023-06-22
Add override dates support page

justin-gardiner:


server/routes/support/handlers/offenderLicenceDates.ts
```diff
@@ -0,0 +1,80 @@
...
+      licenceStartDate: this.simpleDateToLicenceDate(formData.lsd),
+      licenceExpiryDate: this.simpleDateToLicenceDate(formData.led),
+      topupSupervisionStartDate: this.simpleDateToLicenceDate(formData.tussd),
+      topupSupervisionExpiryDate: this.simpleDateToLicenceDate(formData.tused),
+    }
+  }
+
+  simpleDateToLicenceDate = (date: SimpleDate): string | undefined => {
+    if (!date) {
+      return undefined
```
SebNegahbanMoJ, line 74: What happens if you only enter a new value for one date (eg CRD) and leave the rest blank? Will this not override the values stored in the DB to `undefined` (or `null`)?

SebNegahbanMoJ: Nevermind, just seen that this is handled in the njk file with `formResponses.<date> or licenceDates.<date>`

server/views/pages/support/offenderLicenceDates.njk
```diff
@@ -0,0 +1,219 @@
...
+                            {{ licence.licenceStartDate | datetimeToDateShort }}
+                        {% else %}
+                            Not found
+                        {% endif %}
+                    </td>
+                    <td class="govuk-table__cell">
+                        {{ datePicker({
+                            id: "lsd",
+                            errorMessage: validationErrors | findError('lsd'),
+                            formResponses: formResponses.led or licenceDates.lsd
```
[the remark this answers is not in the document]
justin-gardiner: good spot! fixed

server/@types/licenceDates.ts
```diff
@@ -0,0 +1,10 @@
+export type LicenceDates = {
```
andrewrlee, line 1: There's [scripts](https://github.com/ministryofjustice/create-and-vary-a-licence#generating-types-from-openapi) to update the swagger definitions and you should be able to reference this from there? (I think)

andrewrlee (approved):
LGTM

PR #573, opened 2023-06-28
Fix policy types

andrewrlee:
Now policy stubs have additional typesafety of input types


server/@types/licenceApiClientTypes.ts
```diff
@@ -38,3 +38,4 @@ export type AdditionalConditionApResponse = components['schemas']['AdditionalCon
 export type AdditionalConditionPssResponse = components['schemas']['AdditionalConditionPss']
 export type Input = components['schemas']['Input']
 export type LicenceConditionChange = components['schemas']['LicenceConditionChanges']
+export type OverrideLicenceDatesRequest = components['schemas']['OverrideLicenceDatesRequest']
```
andrewrlee, line 41: @justin-gardiner I thought we might be able to reuse this instead of the custom type you added but this includes the reason whereas the custom type assumes its separate. not a big deal really!

justin-gardiner (approved):
lgtm

PR #575, opened 2023-06-29
Adding new handler for changes to offender name and DOB

SebNegahbanMoJ:
Paired with API PR [here](https://github.com/ministryofjustice/create-and-vary-a-licence-api/pull/264)


andrewrlee (dismissed):
LGTM

server/listeners/eventHandlers/prisonEvents/index.ts
```diff
@@ -21,6 +23,10 @@ export default function buildEventHandler({ licenceService, prisonerService }: S
         case 'CONFIRMED_RELEASE_DATE-CHANGED':
           datesChangedEventHandler.handle(eventMessage).catch(error => logger.error(error))
           break
+        case 'OFFENDER-UPDATED':
+        case 'OFFENDER_DETAILS-CHANGED':
+          offenderDetailsChangedEventHandler.handle(eventMessage).catch(error => logger.error(error))
```
andrewrlee, line 28: we do have cypress tests for the other event types, we should probably have one for this?

SebNegahbanMoJ: Ah of course we do, will do!

integration_tests/integration/eventHandlers.cy.ts
```diff
@@ -148,5 +148,53 @@ context('Event handlers', () => {
...
+              "Value": "OFFENDER_DETAILS-CHANGED"
+            }
+          }
+         }`
+      )
+
+      cy.task('verifyEndpointCalled', {
+        verb: 'PUT',
+        path: '/offender/nomisid/G9786GC/update-offender-details',
+        times: 1,
```
andrewrlee, line 196: would be nice if we could verify the payload that is sent rather than just that the endpoint was called, but granted none of the other examples do that

andrewrlee (approved):
LGTM

PR #577, opened 2023-06-30
CVSL-1143 fixing return to caseload links

andrewrlee:
Currently the "return to caseload" links rely on storing the destination location in the session.

When the application session times out the user will be redirected to auth and assuming the user's auth session is still active they will be redirected back to the application where a new application session will be created. 

This happens transparently and the user will returned to the page where they were on.

The lack of a return destination in the session breaks the session.

This change puts in a sensible default location for each link. (Might be nicer to move to a central middleware or better different paths for each destination - which would massively improve observability but that would require quite a lot of restructuring)

### DIFF
--- a/run-local.sh
+++ b/run-local.sh
@@ -1,4 +1,5 @@
-#
+#!/bin/bash
+
 # This script is used to run the Create and Vary a licence UI locally, to sets up the containers and installs the
 # necessary dependencies required for the UI.
 #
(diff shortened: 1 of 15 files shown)


run-local.sh
```diff
@@ -1,4 +1,5 @@
-#
+#!/bin/bash
```
andrewrlee, line 1: My shellcheck plugin had suggestions..

server/routes/creatingLicences/handlers/caseload.test.ts
```diff
@@ -1,21 +1,18 @@
...
 import { ManagedCase } from '../../../@types/managedCase'
 
-const licenceService = new LicenceService(null, null, null, null) as jest.Mocked<LicenceService>
 const caseloadService = new CaseloadService(null, null, null) as jest.Mocked<CaseloadService>
 
-jest.mock('../../../services/licenceService')
 jest.mock('../../../services/caseloadService')
 
 describe('Route Handlers - Create Licence - Caseload', () => {
-  const handler = new CaseloadRoutes(licenceService, caseloadService)
```
andrewrlee, line 18: unused dependency

server/views/pages/create/checkAnswers.njk
```diff
@@ -130,7 +130,7 @@
                         
                     {% endif %}
                     
-                    <a href={{ backLink }} class="govuk-link govuk-!-font-size-19">Return to caselist</a>
+                    <a href="{{ backLink }}" class="govuk-link govuk-!-font-size-19">Return to caselist</a>
```
andrewrlee, line 133: when no backlink was present this was generating a url like: `/licence/create/id/123/class=%22govuk-link` due to lack of quotes

server/routes/creatingLicences/handlers/confirmCreate.ts
```diff
@@ -18,7 +18,7 @@ export default class ConfirmCreateRoutes {
   GET = async (req: Request, res: Response): Promise<void> => {
     const { nomisId } = req.params
     const { user } = res.locals
-    const backLink = req.session.returnToCase
+    const backLink = req.session.returnToCase || '/licence/create/caseload'
```
justin-gardiner, line 21: would they be able get here without a valid session?

justin-gardiner (approved):
lgtm

PR #580, opened 2023-07-03
CVSL-1145 Adding retries and fixing app insights on update job

andrewrlee:


SebNegahbanMoJ (approved):
Looks reasonable to me!

PR #582, opened 2023-07-03
Moving to latest orb versions

andrewrlee:


.circleci/config.yml
```diff
@@ -70,28 +69,6 @@ jobs:
             - dist
             - .cache/Cypress
 
-  check_outdated:
```
andrewrlee, line 73: There is a version from the orb that we can use, which means we don't need to use the slack orb

.circleci/config.yml
```diff
@@ -1,9 +1,8 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@6
-  slack: circleci/slack@4.4.2
-  codecov: codecov/codecov@3.2.0
+  hmpps: ministryofjustice/hmpps@7
```
andrewrlee, line 4: version 7 is latest and Matt has added some additional debug so we should be able to see what is happening around the app_version issue

SebNegahbanMoJ (approved):

PR #600, opened 2023-07-21
CVSL-465 Don't allow PP to vary licence conditions if LED < current d…

andrewrlee:
…ate < TUSED


nishanthmahasamudram:
https://github.com/ministryofjustice/create-and-vary-a-licence/pull/609 is the latest version.

PR #603, opened 2023-07-24
CVSL-1185 fixed security audit exceptions for request and tough-cookie

louisenorris:


audit-ci.json
```diff
@@ -1,6 +1,5 @@
 {
   "low": true,
```
andrewrlee, line 2: ah, this can go entirely

PR #605, opened 2023-07-25
CVSL-1230 Adding preprod phase banner

andrewrlee:


server/config.ts
```diff
@@ -32,6 +32,7 @@ export interface ApiConfig {
 export default {
   buildNumber: get('BUILD_NUMBER', '1_0_0', requiredInProduction),
   gitRef: get('GIT_REF', 'xxxxxxxxxxxxxxxxxxx', requiredInProduction),
+  phaseName: process.env.SYSTEM_PHASE || 'BETA',
```
justin-gardiner, line 35: do you need to set a value of SYSTEM_PHASE for prod to prevent BETA being displayed there?

PR #609, opened 2023-08-01
Alee/cvsl 465 b

nishanthmahasamudram:


server/routes/varyingLicences/handlers/confirmVaryAction.ts
```diff
@@ -8,7 +8,9 @@ export default class ConfirmVaryActionRoutes {
   constructor(private readonly licenceService: LicenceService) {}
 
   GET = async (req: Request, res: Response): Promise<void> => {
-    res.render('pages/vary/confirmVaryQuestion')
+    const { licence } = res.locals
+    const { isInPssPeriod = false } = licence
+    res.render('pages/vary/confirmVaryQuestion', { isInPssPeriod, licence })
```
SebNegahbanMoJ, line 13: Is the licence object not available on this page from middleware (ie without being passed in explicitly)?

server/routes/varyingLicences/handlers/viewActiveLicence.test.ts
```diff
@@ -73,5 +121,93 @@ describe('Route Handlers - Vary Licence - View active licence', () => {
 
       expect(res.redirect).toHaveBeenCalledWith(`/licence/vary/id/${licence.id}/timeline`)
     })
+
+    it('should display expired section if licence type is AP_PSS, is in pss period, isActivatedInPssPeriod with additional conditions', () => {
```
SebNegahbanMoJ, line 125: The `cheerio` related tests should probably be in the `views` directory, not the `handlers` directory, as they're testing the njk files rather than the ts handlers

server/views/pages/create/checkAnswers.njk
```diff
@@ -28,16 +30,16 @@
                 {% if licence.typeCode === 'AP' %}
                     <h1 class="govuk-heading-l">Vary licence details</h1>
                 {% elseif licence.typeCode === 'AP_PSS' %}
-                    <h1 class="govuk-heading-l">Vary licence and post sentence supervision order details</h1>
+                    <h1 class="govuk-heading-l">Vary post sentence supervision order details</h1>
```
SebNegahbanMoJ, line 33: Is this correct? This doesn't seem to be checking `inInPssPeriod` so it should be possible for users to vary the AP conditions on an AP_PSS licence

server/views/pages/create/checkAnswers.njk
```diff
@@ -78,7 +80,9 @@
                 {{ inductionMeetingDetails(licence, isInProgress, validationErrors) }}
             {% endif %}
 
-            {% if licence.typeCode === 'AP' or licence.typeCode === 'AP_PSS' %}
+
+            {% if (not licence.isInPssPeriod or (licence.isInPssPeriod and (not isInVariation or not licence.isActivatedInPssPeriod))) and
```
SebNegahbanMoJ, line 84: I think this logic can be tidied up to
```
if (licence.typeCode === 'AP' or licence.typeCode === 'AP_PSS' ) and (not licence.isInPssPeriod or not isInVariation or not licence.isActivatedInPssPeriod)
```

Having the AP/AP_PSS check first makes the logic more readable in my opinion, and the check of `(not licence.isInPssPeriod` makes the `licence.isInPssPeriod` part of the `and` irrelevant, because it will either be:
licence.isInPssPeriod is false, which means the logic will short-circuit at `not licence.isInPssPeriod` as it returns `true` in an `or`
The check is true, in which case the logic will be `true and ...`, at which point the `true` is irrelevant and it just relies on the other half of the statement. I'm happy to have a chat about this and if I'm missing something I can retract this statement!

SebNegahbanMoJ: Alternatively, you might be able to grab the logic you use later to display the `expired conditions` section and just stick a `not` in front of it, ie:

`if (licence.typeCode === 'AP' or licence.typeCode === 'AP_PSS') and not (isInPssPeriod and (isInVariation or licence.isActivatedInPssPeriod))`

I haven't thought as much about this one though so I might be missing something

server/views/pages/vary/viewActive.njk
```diff
@@ -28,13 +30,33 @@
 
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-full">
-            {% if licence.typeCode === 'AP' or licence.typeCode === 'AP_PSS' %}
+            {% if (not licence.isInPssPeriod or (licence.isInPssPeriod and (not isInVariation or not licence.isActivatedInPssPeriod))) and
```
SebNegahbanMoJ, line 33: Same as the previous comment with the suggested logic refactoring

server/routes/varyingLicences/handlers/viewActiveLicence.test.ts
nishanthmahasamudram: That file was existing so i used the same to test my scenarios

server/views/pages/create/checkAnswers.njk
nishanthmahasamudram: I don't see any difference between my logic and this if (licence.typeCode === 'AP' or licence.typeCode === 'AP_PSS') and not (isInPssPeriod and (isInVariation or licence.isActivatedInPssPeriod)). Can you explain me what is the difference?

server/views/pages/create/checkAnswers.njk
nishanthmahasamudram: reverted this back, as it is not my change

server/routes/varyingLicences/handlers/viewActiveLicence.test.ts
SebNegahbanMoJ: Was the existing file not just checking the handler rather than the view?

server/views/pages/create/checkAnswers.njk
SebNegahbanMoJ: From a purely code perspective, the only change is removing the `licence.isInPssPeriod` from `or (licence.isInPssPeriod and (...` as I'm not sure it's necessary.
Personally I thought it made the code more readable, but the logic should be unchanged as that looks fine to me. This is one of the things I'd like to discuss with other devs as if they disagree with me then I'm more than happy to back down!

server/services/conditionService.ts
```diff
@@ -254,4 +257,19 @@ export default class ConditionService {
     return mappedConditions
   }
   /* eslint-disable no-param-reassign */
+
+  async getParentOrSelfAdditionalLicenceConditions(licence: Licence, user: User): Promise<AdditionalCondition[]> {
```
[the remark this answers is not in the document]
andrewrlee: Yes @SebNegahbanMoJ , specifically suggested to pull this down (not sure about the name though!)

Controllers should be very light, ideally just convert to and from request/response and then delegate down to the service layer. 
 
There was far too much logic in this situation.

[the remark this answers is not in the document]
andrewrlee: > mainly because it's not really reflective of what's happening inside

We use abstraction to avoid exposing "the how".

Ideally we'd express this in a more domain specific way: What are we showing here?

 "getRelevantConditions",  "getConditionsToCheck"... probably too abstract...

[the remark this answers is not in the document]
andrewrlee: (not a big fan of `getParentLicenceOrSelf` for the same reason)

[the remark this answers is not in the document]
SebNegahbanMoJ: That makes sense, I was very much pushing logic into the controller but if it's better practice to just have a call to the service layer that says 'Get me what I want', that's completely fine.
If we want to go with the hyper-specific service methods, I'd probably advocate for something like
`getAdditionalConditionsForSummaryAndPdf`? That way it's specific in its use but without getting into the detail of what is happening within the method

server/services/licenceService.ts
```diff
@@ -573,15 +573,6 @@ export default class LicenceService {
     return this.licenceApiClient.deleteOmuEmailAddress(prisonId, user)
   }
 
-  async getParentLicenceOrSelf(licenceId: string, user: User): Promise<Licence> {
```
andrewrlee, line 576: Can we keep this here but delegate down to the licenceApiClient? 

Controllers shouldn't be directly talking to the client, can swap those back to use the licence service

server/services/conditionService.ts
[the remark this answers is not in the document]
SebNegahbanMoJ: This is also only getting AP conditions right? The PSS are contained within `licence.additionalPssConditions`? If that's the case, I'd also advocate for specifying `AP` in the method name, something like `getAdditionalAPConditionsForSummaryAndPdf`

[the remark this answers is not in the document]
SebNegahbanMoJ: Are you certain that it's fetching the PSS conditions? The `Licence` object type appears to have:
``` 
additionalLicenceConditions: components['schemas']['AdditionalCondition'][]
additionalPssConditions: components['schemas']['AdditionalCondition'][]
```
and the nunjucks component that renders the PSS conditions appears to read them directly from the licence (which I believe is defined in the njk files by middleware) rather than from the additionalConditions object that's passed in to the checkAnswers view (PSS component is [here](https://github.com/ministryofjustice/create-and-vary-a-licence/blob/27d3529fc716a3fb015003320d600ef0da6a6fbd/server/views/partials/additionalPssConditionsSummary.njk#L49))

server/services/licenceService.ts
SebNegahbanMoJ: I believe this had to be done because otherwise it caused a circular dependency between conditionService and licenceService, but hopefully Nish can confirm (I may be making that up 😅)

nishanthmahasamudram: Only solution is to duplicate the method in both licenceService and licenceApiClient. Can I do that?

nishanthmahasamudram: I am delegating the getParentLicenceOrSelf method through licenceApiClient So the controllers can communicate through service.

server/views/pages/create/checkAnswers.njk
```diff
@@ -87,6 +91,27 @@
...
+                    so the conditions from their standard determinate licence can no longer be varied.</p>
+                    <details class="govuk-details" data-module="govuk-details">
+                        <summary class="govuk-details__summary">
+                            <span class="govuk-details__summary-text">
+                                View conditions from expired licence
+                            </span>
+                        </summary>
+                        <div class="govuk-details__text">
+                            {{ additionalLicenceConditions(licence, additionalConditions, conditionsWithUploads, null, validationErrors) }}
+                            {{ bespokeConditions(licence, null) }}
```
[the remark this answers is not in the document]
nishanthmahasamudram: fetching bespoke conditions from parent if it is a variation in PSS period

server/services/conditionService.test.ts
```diff
@@ -178,4 +180,55 @@ describe('ConditionService', () => {
       expect(conditionService.parseResponse(conditionConfig)).toEqual(expectedOutput)
     })
   })
+
+  describe('Get parent or self licence additional conditions', () => {
```
SebNegahbanMoJ, line 184: Minor, but the description text should probably be updated to reference the new method name rather than the old one

server/routes/varyingLicences/handlers/viewActiveLicence.test.ts
SebNegahbanMoJ: I don't think these specs need to be here anymore do they? They appear to be in a new spec called `viewActive`

server/routes/creatingLicences/handlers/checkAnswers.ts
```diff
@@ -27,11 +27,19 @@ export default class CheckAnswersRoutes {
...
-    const { conditionsWithUploads, additionalConditions } = this.conditionService.additionalConditionsCollection(
-      licence.additionalLicenceConditions
-    )
+    const conditionsToDisplay = await this.conditionService.getAdditionalAPConditionsForSummaryAndPdf(licence, user)
 
-    res.render('pages/create/checkAnswers', { additionalConditions, conditionsWithUploads, backLink })
+    const { conditionsWithUploads, additionalConditions } =
+      this.conditionService.additionalConditionsCollection(conditionsToDisplay)
+
+    const parentBespokeConditions = await this.conditionService.getbespokeConditionsForSummaryAndPdf(licence, user)
```
SebNegahbanMoJ, line 35: This appears to have similar behaviour to the AP condition equivalent, which means if it's not appropriate to get the parentBespokeConditions, it will get the licence bespoke conditions instead. I don't think this is a problem, but I think it makes the naming a bit confusing. It also means that in the njk bespoke conditions macro, `parentBespokeConditions` will always be populated if `licence.bespokeConditions` is, meaning that the check on `licence.bespokeConditions` is somewhat unused, because either you will want the parent conditions (which will be in `parentBespokeConditions`, or you want the licence bespoke conditions, which will also be populated in `parentBespokeConditions`.

This isn't a blocker, but is a simplification to the logic that we could make.

nishanthmahasamudram: why I used parentBespokeConditions is I wanted to differentiate between existing bespokeConditions macro and my variable. As it is not recommended to use the same names for methods and variables. If you feel it need's to be changed please suggest me the name.

Other one is if we are using parentBespokeConditions in bespokeConditionsSummary page then licence.bespokeConditions is useless. Yes, it will be only when we are calling macro from checkAnswers.njk and viewActiveLicence.njk files and for remaining files it will only use licence.bespokeConditions as parentBespokeConditions will be null elsewhere.

SebNegahbanMoJ: Ah, I hadn't realised the bespokeConditions macro was used elsewhere.
In which case, the only change I would possibly suggest would be changing `parentBespokeConditions` to `bespokeConditionsToDisplay`, but again, this is very minor so if you disagree, don't worry.

nishanthmahasamudram: updated the variable name from parentBespokeConditions to bespokeConditionsToDisplay.

server/routes/varyingLicences/handlers/viewActiveLicence.test.ts
nishanthmahasamudram: Yes, removed.

SebNegahbanMoJ (dismissed):
LGTM

gurps-singh (approved):

PR #625, opened 2023-08-09
CVSL-999 add feature flag

gurps-singh:
This PR is part of CVSL-999 and this is to introduce a feature flag to turn the new search off or on dependent on the environment.


andrewrlee (approved):
LGTM

PR #627, opened 2023-08-16
CVSL-1242-Approver view | Configure back link to take people from the…

nishanthmahasamudram:
… DPS profile page to CVL


server/routes/approvingLicences/handlers/approvalView.ts
```diff
@@ -32,6 +32,8 @@ export default class ApprovalViewRoutes {
 
       const comDetails = await this.communityService.getStaffDetailByUsername(comUsername)
 
+      const returnPath = encodeURIComponent(`/licence/approve/id/${licence.id}/view`)
```
nishanthmahasamudram, line 35: Return path has to be URL encoded as it contains special characters.

server/views/pages/approve/view.njk
```diff
@@ -118,7 +120,9 @@
         <div class="govuk-grid-row">
             <h2 class="govuk-heading-l govuk-!-margin-top-9">View more information</h2>
             <p class="govuk-body">
-                <a class="govuk-link govuk-link--no-visited-state" href="{{dpsUrl}}/prisoner/{{licence.nomsId}}">See this person’s full Digital Prison Services profile</a> 
+                <a class="govuk-link govuk-link--no-visited-state back-link" href={{ dpsBackLink }}>
```
nishanthmahasamudram, line 123: This link is to navigate back to CVL from DPS prisoners profile page.

server/routes/approvingLicences/handlers/approvalView.ts
SebNegahbanMoJ: I'm not sure I understand this comment, wouldn't `licence.id` resolve to a numerical value and just be added into the string at the time of creation of `returnPath`, meaning that any time you referenced `returnPath` in future it would return something like `licence/approve/id/3/view`?

nishanthmahasamudram: They explicitly specified us to encode the return path if there is a special character in it, as it might cause problem while they try to read it. There are "\" in our returnPath which is considered as a special character , so I added it.

SebNegahbanMoJ (approved):

PR #629, opened 2023-08-22
Cvsl 1269 comms banner and whats new page

nishanthmahasamudram:


helm_deploy/create-and-vary-a-licence/values.yaml
```diff
@@ -33,6 +33,7 @@ generic-service:
     LICENCE_WATERMARK: "false"
     APPLICATIONINSIGHTS_CONNECTION_STRING: "InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY);IngestionEndpoint=https://northeurope-0.in.applicationinsights.azure.com/;LiveEndpoint=https://northeurope.livediagnostics.monitor.azure.com/"
     SERVICE_NAME: "create-and-vary-a-licence"
+    SHOW_WHATS_NEW_BANNER: "false"
```
nishanthmahasamudram, line 36: make SHOW_WHATS_NEW_BANNER flag true once Version changes are deployed.

server/routes/index.ts
```diff
@@ -17,12 +17,15 @@ import populateCurrentUser from '../middleware/populateCurrentUser'
 import flashMessages from '../middleware/flashMessageMiddleware'
 import fromReviewMiddleware from '../middleware/fromReviewMiddleware'
 import PrisonerController from './prisonerController'
+import WhatsNewPage from './whatsNewPage/whatsNewPage'
```
andrewrlee, line 20: There is a convention that these are called `Controller` not `Page`

(Page is a bit confusing as this is what we call e2e test page objects)

server/routes/index.ts
```diff
@@ -17,12 +17,15 @@ import populateCurrentUser from '../middleware/populateCurrentUser'
...
 
 export default function Index(services: Services): Router {
   const router = Router({ mergeParams: true })
   const prisonerController = new PrisonerController(services.prisonerService)
+  const whatsNewPage = new WhatsNewPage()
 
   router.get('/prisoner/:nomsId/image', prisonerController.getImage())
+  router.get('/whats-new-page', whatsNewPage.GET)
```
andrewrlee, line 28: These should be authenticated right? 

seems like this would be better belonging in the `homeRoutes` section?
This tends to have things that are more meta - accessiblity statements/ ts+cs etc..

server/routes/index.ts
nishanthmahasamudram: renamed it to WhatsNewController

server/routes/index.ts
```diff
@@ -17,12 +17,15 @@ import populateCurrentUser from '../middleware/populateCurrentUser'
...
 import fromReviewMiddleware from '../middleware/fromReviewMiddleware'
 import PrisonerController from './prisonerController'
+import WhatsNewPage from './whatsNewPage/whatsNewPage'
 
 export default function Index(services: Services): Router {
   const router = Router({ mergeParams: true })
   const prisonerController = new PrisonerController(services.prisonerService)
+  const whatsNewPage = new WhatsNewPage()
 
   router.get('/prisoner/:nomsId/image', prisonerController.getImage())
```
andrewrlee, line 27: this should be authenticated and we should have a user in context so should sit below the rest of the middleware along with the other routes

server/routes/whatsNewPage/whatsNewPage.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    res = {
+      header: jest.fn(),
+      send: jest.fn(),
+      render: jest.fn(),
+    } as unknown as Response
+  })
+
+  describe('GET', () => {
+    it('should render cases when user only has 1 caseloaded prison', async () => {
```
andrewrlee, line 23: this description doesn't seem right?

helm_deploy/create-and-vary-a-licence/values.yaml
andrewrlee: to test this you could override this on test1?

Then once this is merged we could see that dev doesn't have the banner but that test1 does.

server/routes/__testutils/appSetup.ts
```diff
@@ -7,24 +7,17 @@ import errorHandler from '../../errorHandler'
 import * as auth from '../../authentication/auth'
 import type { Services } from '../../services'
 import AuthRole from '../../enumeration/authRole'
+import { User } from '../../@types/CvlUserDetails'
```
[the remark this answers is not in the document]
nishanthmahasamudram: > not sure this is right?
> 
> Shouldn't this be of type `LocalsUser` which is a combination of `User` and `CvlUserDetails`?
> 
> Why did you need to change this?

Yes, this User is pointing to LocalsUser only.
export type User = Express.LocalsUser;
Instead of Using LocalUser in all places we decided to assign it to the User as many places in the application is pointing to this, if you remember.

helm_deploy/create-and-vary-a-licence/values.yaml
andrewrlee: to override it in test1 - you can specify it in `values-test1.yml`

andrewrlee (approved):
LGTM

PR #632, opened 2023-08-24
CVSL-999 add probation search page

gurps-singh:
This PR is part of CVSL-999 and adds the search page. This includes functionality to search from the create or vary caseloads. 

Before:
Screenshot 2023-08-09 at 09 29 33

After:
Screenshot 2023-08-24 at 12 45 24

Screenshot 2023-08-24 at 12 45 30


integration_tests/pages/checkAnswers.ts
```diff
@@ -38,4 +39,9 @@ export default class CheckAnswersPage extends Page {
     cy.get(this.discardButtonId).click()
     return Page.verifyOnPage(DiscardPage)
   }
+
+  clickBackToCaseload = (): CaseloadPage => {
+    cy.get('.govuk-button-group > .govuk-link').click()
```
andrewrlee, line 44: meh.. good to add data attributes to identify buttons rather than using non-semantic css selectors

`data-qa="back-to-caseload-link"`

server/routes/index.ts
```diff
@@ -39,6 +40,7 @@ export default function Index(services: Services): Router {
   router.use(supportRoutes(services))
   router.use(changeLocationRoutes(services))
   router.use(changeTeamRoutes(services))
+  router.use(probationSearchRoutes(services))
```
andrewrlee, line 43: Should this sit inside one of the route files defined above? 
does it belong to a single area or is it shared across multiple areas?

integration_tests/pages/checkAnswers.ts
gurps-singh: Just so happened to be reading this on Friday related to this - https://dev.to/christinepinto/data-qa-attribute-a-better-way-to-select-elements-for-ui-test-automation-48lm#:~:text=Solution%3A%20Data%2Dqa%20Attributes&text=These%20HTML%20attributes%20enable%20us,or%20purposes%20on%20the%20page.

server/routes/index.ts
gurps-singh: Shared across create and vary journeys - might be better to have generic `searchRoutes` instead?

server/views/pages/search/probationSearch/probationSearch.njk
```diff
@@ -0,0 +1,245 @@
...
+                            "aria-sort": "none"
+                        }
+                    }
+                ],
+                rows: results
+            }) }}
+        </form>
+    {% else %}
+        {% if setPeopleInPrisonTab %}
+            <div class="govuk-grid-row">
```
andrewrlee, line 83: can you do something like? 
```
{% if setPeopleInPrisonTab %}
            {% set caption = "In Prison Search Results" %}
            <h2 class="govuk-heading-l" id = "tab-heading-prison">{{ countText }}</h2>
{% else %}
            {% set caption = "On Probation Search Results" %}
            <h2 class="govuk-heading-l" id = "tab-heading-probation">{{ countText }}</h2>
{% endif %}

{% if results.length == 0 %}
  ... render message
{% elseif %}
  ... render rows
{% endif %}
```

andrewrlee: would be tempted to pass in caption and headingId as separate macro params and get rid of the initial condition too
```
    <h2 class="govuk-heading-l" id = "tab-heading-{{headingId}}">{{ countText }}</h2>
```

You can wrap the params in an object like in javascript if you want the effect of named params

gurps-singh: Cool, would it be ok if I add these changes to the final PR for the search - got one more to come?

server/routes/search/handlers/probationSearch.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (searchResponse.inPrisonCount === 0 || searchResponse.inPrisonCount > 1) {
+      inPrisonCountText = `People in prison (${searchResponse.inPrisonCount} results)`
+    } else {
+      inPrisonCountText = `People in prison (${searchResponse.inPrisonCount} result)`
+    }
+
+    if (searchResponse.onProbationCount === 0 || searchResponse.onProbationCount > 1) {
+      onProbationCountText = `People on probation (${searchResponse.onProbationCount} results)`
+    } else {
+      onProbationCountText = `People on probation (${searchResponse.onProbationCount} result)`
```
andrewrlee, line 29: these might be better handled with a nunjucks filter, similar to [this](https://stackoverflow.com/a/22336061)

server/views/pages/search/probationSearch/probationSearch.njk
andrewrlee: sure, added another suggstion around using a filter to pluralise which could also be done as a follow up

PR #633, opened 2023-08-24
CVSL-465-bug-fixes

nishanthmahasamudram:


server/views/partials/bespokeConditionsSummary.njk
```diff
@@ -9,13 +9,14 @@
         </div>
     {% endif %}
     {% set bespokeConditions = [] %}
+    {% set bespokeConditionsLength = bespokeConditionsToDisplay and bespokeConditionsToDisplay.length or licence.bespokeConditions.length  %}
```
andrewrlee, line 12: This `bespokeConditionsToDisplay or licence.bespokeConditions` kind of logic in this partial is very esoteric and I think will be difficult to understand when coming back to it.

There's no context here to help me understand what each set of conditions refer to, it feels like working out which set of conditions should be pushed into controller... 
 
What happens if there are 0 bespokeConditionsToDisplay because there are legitimately 0 bespoke conditions to display on that version of the licence? `0` being falsey would mean that we default to `licence.bespokeConditions.length`

This is missing tests too

nishanthmahasamudram: fixed by passing bespoke conditions explicitly from all the function calls.

andrewrlee (approved):
LGTM

PR #634, opened 2023-08-24
CVSL-1315 Adding retries and fixing app insights for job

andrewrlee:
Also removing unnecessary scripts and managing openapi-typescript in package.json so all devs end up using the same version


nishanthmahasamudram (approved):
LGTM

PR #635, opened 2023-08-25
CVSL-1287 Grouping conditions by code to make rendering simpler

andrewrlee:


server/views/partials/additionalLicenceConditionsSummary.njk
```diff
@@ -35,29 +35,24 @@
         }), additionalLicenceConditions) %}
     {% endif %}
 
-    {% for condition in additionalConditions %}
-        {% if condition.uploadSummary.length %}
```
andrewrlee, line 39: need to check - can you access this page with no upload summaries

server/views/partials/additionalLicenceConditionsSummary.njk
```diff
@@ -35,29 +35,24 @@
...
-            {% set actionLink =  "/licence/create/id/" + licence.id + "/additional-licence-conditions/condition/" + condition.code + "/file-uploads" %}
-        {% elseif condition | checkConditionRequiresInput(licence.version) %}
-            {% set actionLink =  "/licence/create/id/" + licence.id + "/additional-licence-conditions/condition/" + condition.id %}
-        {% else %}
-            {% set actionLink =  "/licence/create/id/" + licence.id + "/additional-licence-conditions" %}
```
andrewrlee, line 44: this was never true

server/views/partials/additionalPssConditionsSummary.njk
```diff
@@ -57,7 +57,7 @@
                         visuallyHiddenText: "Change requirement"
                     }
                 ]
-            } if condition | checkConditionRequiresInput(licence.version) and isEditable
+            } if isEditable
```
andrewrlee, line 60: `checkConditionRequiresInput` always returned true

server/services/conditionService.ts
```diff
@@ -136,10 +136,17 @@ export default class ConditionService {
...
-      (c: AdditionalCondition) => !conditionsWithUploads.find((c2: AdditionalCondition) => c.id === c2.id)
-    )
-    return { conditionsWithUploads, additionalConditions }
+    const conditionsByCode = conditions
+      // This filter will need to be removed when the frontend handles conditions with uploads transparantly
+      .filter((c: AdditionalCondition) => {
+        return !conditionsWithUploads.find((c2: AdditionalCondition) => c.id === c2.id)
+      })
+      .reduce((acc, c) => {
+        acc[c.code] = [...(acc[c.code] || []), c]
```
SebNegahbanMoJ, line 145: How does `[...(acc[c.code] || []), c]` work?
My initial interpretation would be that it will always set the first item in the array to be an empty array itself, ie the accumulator will end up with:
```{"abc123": [ [], <condition1>, <condition2> ] }```
That doesn't seem like it would be correct though!

Comment (andrewrlee) [reply server/services/conditionService.ts:145]: `x = (acc[c.code] || [])`

if this is the first time we've seen this condition then `x = []` otherwise `x` is the previously seen conditions in an array.

`[...x, c]` all the previous conditions and this condition added to the end. 

So basically just makes sure there is an array, and then keeps adding to the end of the array.

I'm going to extract out a `chunkBy` function into utils in a follow up so I'll make this a bit easier to read then

Comment (SebNegahbanMoJ) [reply server/services/conditionService.ts:145]: Ah I misread `...(acc[c.code] || [])` as `(...acc[c.code] || [])`, having the spread operator outside of the parentheses makes a lot more sense!

File: server/views/partials/additionalLicenceConditionsSummary.njk [hunk 5]
```diff
@@ -35,29 +35,24 @@
         }), additionalLicenceConditions) %}
     {% endif %}
 
-    {% for condition in additionalConditions %}
-        {% if condition.uploadSummary.length %}
-            {% set actionLink =  "/licence/create/id/" + licence.id + "/additional-licence-conditions/condition/" + condition.code + "/file-uploads" %}
```
SebNegahbanMoJ, line 40: Is removing this intentional? Will the new code not mean that clicking the 'change' link won't appear for MEZ with multiple maps? (and if there's only one upload, the change link will no longer go to `/file-uploads`?)

andrewrlee: this list of additionalConditions will never include any conditions that have upload information, those are dealt with the `conditionsWithUploads` loop below

SebNegahbanMoJ (approved):

PR #647, opened 2023-09-01
CVSL-1287 Extracting out condition routes into separate section

andrewrlee:
Although there are a lot of files this change is just extracting out a load of routes into a separate area.
This section of the app is used across create,vary and updating based on policy changes. 

The urls are gnarly but that's a different problem... also could do with moving the views... but will leave that for a follow up


server/services/conditionService.ts
```diff
@@ -1,34 +1,34 @@
...
-import ElectronicMonitoringTypes from '../routes/creatingLicences/types/additionalConditionInputs/electronicMonitoringTypes'
-import ElectronicMonitoringPeriod from '../routes/creatingLicences/types/additionalConditionInputs/electronicMonitoringPeriod'
-import ApprovedAddress from '../routes/creatingLicences/types/additionalConditionInputs/approvedAddress'
-import AlcoholMonitoringPeriod from '../routes/creatingLicences/types/additionalConditionInputs/alcoholMonitoringPeriod'
-import AlcoholRestrictionPeriod from '../routes/creatingLicences/types/additionalConditionInputs/alcoholRestrictionPeriod'
-import CurfewTerms from '../routes/creatingLicences/types/additionalConditionInputs/curfewTerms'
-import CurfewAddress from '../routes/creatingLicences/types/additionalConditionInputs/curfewAddress'
-import NoContactWithVictim from '../routes/creatingLicences/types/additionalConditionInputs/noContactWithVictim'
-import ReportToApprovedPremises from '../routes/creatingLicences/types/additionalConditionInputs/reportToApprovedPremises'
+import RegionOfResidence from '../routes/manageConditions/types/additionalConditionInputs/regionOfResidence'
```
andrewrlee, line 2: These service layer modules referencing route level modules are a bit whiffy.. violate direction of dependencies, they should probably be pulled into the service layer

gurps-singh (approved):

PR #649, opened 2023-09-01
CVSL-1182 Adding versioning to support page

SebNegahbanMoJ:
Slightly different to the ticket ACs as the logs work slightly differently to the ticket implies. Instead, the licence version has been added to the table of licences for a person in prison or on probation. The version number will also be a link to the PDF, if the licence is in a suitable state for pdf-viewing.

Relies on the [API counterpart PR](https://github.com/ministryofjustice/create-and-vary-a-licence-api/pull/304) being merged first


andrewrlee (approved):
LGTM

PR #651, opened 2023-09-05
CVSL-1287 Tidying up MEZ routes

andrewrlee:


server/routes/manageConditions/handlers/additionalLicenceConditionUploadsRemovalHandler.test.ts
```diff
@@ -1,11 +1,11 @@
 import { Request, Response } from 'express'
 import LicenceService from '../../../services/licenceService'
-import AdditionalLicenceConditionDeletionHandler from './additionalLicenceConditionDeletionHandler'
+import AdditionalLicenceConditionUploadsRemovalHandler from './additionalLicenceConditionUploadsRemovalHandler'
```
andrewrlee, line 3: Giving a better name - this is only used when removing file uploads

server/routes/manageConditions/handlers/deleteConditionsByCodeHandler.test.ts
```diff
@@ -1,10 +1,10 @@
 import { Request, Response } from 'express'
 import LicenceService from '../../../services/licenceService'
-import AdditionalLicenceTypesHandler from './additionalLicenceTypesHandler'
+import DeleteConditionsByCodeHandler from './deleteConditionsByCodeHandler'
```
andrewrlee, line 3: Improving name - this is used to delete all conditions by code

server/routes/manageConditions/index.ts
```diff
@@ -73,30 +72,18 @@ export default function Index({ licenceService, conditionService }: Services): R
   }
 
   // START MEZ
-  // remove area from map list in form builder without confirmation (delete single conditions) - DELETE NOT USED
-  {
-    const controller = new AdditionalLicenceConditionRemoveUploadRoutes(licenceService)
-    get('/condition/id/:conditionId/remove-upload', controller.GET)
-  }
```
andrewrlee, line 80: Removed this as it wasn't being used - only reference was setting a variable that was not used

server/routes/manageConditions/index.ts
```diff
@@ -73,30 +72,18 @@ export default function Index({ licenceService, conditionService }: Services): R
...
     const controller = new AdditionalLicenceConditionUploadsRoutes(licenceService, conditionService)
     get('/additional-licence-conditions/condition/:conditionCode/file-uploads', controller.GET)
     post('/additional-licence-conditions/condition/:conditionCode/file-uploads', controller.POST)
   }
 
-  // delete MEZ condition (delete all conditions)
-  {
-    const controller = new AdditionalLicenceTypesRoutes(licenceService)
-    get('/additional-licence-types/condition/:conditionCode/delete', controller.DELETE)
-  }
```
andrewrlee, line 93: this is a bit more generic than just MEZ so moved it down

server/views/formBuilder.njk
```diff
@@ -255,11 +255,6 @@
             </ul>
         </div>
     </div>
-    <div class="govuk-body">
-        <div>
-            {% set removeFileUploadLink = "/licence/create/id/"+ licenceId +"/condition/id/" + condition.id + "/remove-upload?_csrf=" + csrfToken %}
-        </div>
```
andrewrlee, line 261: unused variable, I guess it was intended to be used but rather than fix now, would be better to just add support for what it was doing after this area stabilises (if still necessary)

louisenorris (approved):
LGTM

PR #653, opened 2023-09-05
CVSL-1179 update additional licence conditions with secondary button

nishanthmahasamudram:


server/views/partials/additionalLicenceConditionsSummary.njk
```diff
@@ -1,36 +1,29 @@
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
 
 {% macro additionalLicenceConditions(licence, additionalConditions, isEditable, validationErrors, title, hideHeading) %}
     {% set errorMessage = validationErrors | findError('additionalLicenceConditions') %}
+    {% set conditionCount = additionalConditions.length or 0 %}
```
andrewrlee, line 6: length of an empty array would be 0 anyway?

nishanthmahasamudram: If the additionalConditions are undefined then the count is coming as empty.
I found it in one of the test cases.
So I added it for protecting undefined stuff.

andrewrlee (approved):
LGTM

PR #654, opened 2023-09-05
CVSL-1287 Splitting out MEZ and normal condition templates

andrewrlee:
Split out the existing routes so MEZ has its own template on
`/additional-licence-conditions/condition/:conditionId`

and then that template posts to its own special handlers for setting data and deletions:
```
/additional-licence-conditions/condition/:conditionId/file-upload-input
/additional-licence-conditions/condition/:conditionId/file-upload-removal
``` 

This reduces complexity in normal condition endpoints and also allows the ability to customize the MEZ journey a bit more freely

### DIFF
--- a/server/routes/manageConditions/handlers/additionalLicenceConditionInput.ts
+++ b/server/routes/manageConditions/handlers/additionalLicenceConditionInput.ts
@@ -42,20 +42,9 @@ export default class AdditionalLicenceConditionInputRoutes {
     const { licenceId } = req.params
     const { conditionId } = req.params
     const { user, licence } = res.locals
-    const isFileUploadRequest = req.file && req.file.fieldname === 'outOfBoundFilename'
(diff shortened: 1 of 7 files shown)


server/routes/manageConditions/handlers/additionalLicenceConditionInput.ts
```diff
@@ -42,20 +42,9 @@ export default class AdditionalLicenceConditionInputRoutes {
     const { licenceId } = req.params
     const { conditionId } = req.params
     const { user, licence } = res.locals
-    const isFileUploadRequest = req.file && req.file.fieldname === 'outOfBoundFilename'
```
andrewrlee, line 45: one big benefit of this approach is that we don't need to have logic like this or `'outOfBoundArea' in req.body`

PR #662, opened 2023-09-07
CVSL-1287 Fixing navigation bug

andrewrlee:
Previously if you click delete condition from the list view it would drop out of the create flow. Now it should know whether in the create flow or review flow and preserve the journey


justin-gardiner (approved):
lgtm

SebNegahbanMoJ (approved):

PR #665, opened 2023-09-12
CVSL-1357 Tidying up bank-holiday usage

andrewrlee:
When looking at updating eslint it picked up this cyclical dependency.


justin-gardiner (approved):
lgtm

PR #666, opened 2023-09-12
CVSL-1364 Simplifying expire-licences job and removing call to NOMIS

SebNegahbanMoJ:
The job has been failing for an unknown period of time due to a call to NOMIS that was requesting too many records at once. This call was found to be entirely unnecessary, as we store the required dates within the CVL database.


andrewrlee (approved):
LGTM

PR #668, opened 2023-09-14
CVSL-999 add highlighting, set active tab and make changes based on feedback

gurps-singh:
This final PR for CVSL-999 includes the addition of highlighting in the table of results and setting the active tab based on the number of results retrieve. This also includes amendments based on testing feedback and feedback from previous PR's related to 999.


assets/js/search/highlightSearchTerm.js
```diff
@@ -0,0 +1,23 @@
...
+    return
+  }
+
+  for (let result of searchResults) {
+    result.innerHTML = caseInsensitiveHighlighting(result.innerHTML, searchTerm)
+  }
+}
+
+function caseInsensitiveHighlighting(html, searchTerm) {
+  var reg = new RegExp(searchTerm, 'gi')
```
andrewrlee, line 19: This should probably be a `const`

assets/js/search/setActiveTab.js
```diff
@@ -0,0 +1 @@
+window.location.hash = document.getElementById('activeTab').value
```
andrewrlee, line 1: I said in that thread but I'm a bit concerned about accessibility issues around this... 
Changing page content dynamically could be confusing

I've asked someone else for their opinion but we'll leave in for now and can redesign if necessary

server/routes/search/handlers/probationSearch.ts
```diff
@@ -1,42 +1,54 @@
 import { Request, Response } from 'express'
 import SearchService from '../../../services/searchService'
 import statusConfig from '../../../licences/licenceStatus'
+import { ProbationSearchResult } from '../../../@types/licenceApiClientTypes'
```
andrewrlee, line 4: when you're just importing things for their types, you can `import type` or if you have a mixture of types or non types: `import { type bob, jim, type sally}`.

This means that the transpiled code doesn't actually import the module unless it needs to, and it won't trigger any side effects

server/views/pages/search/probationSearch/probationSearch.njk
```diff
@@ -24,27 +28,69 @@
...
+                            href: backLink,
+                            attributes: { 'data-qa': 'return-to-caselist' }
+                        }) }}
+                    </p>
+                </div>
+            </div>
         {% else %}
-            {% set caption = "On Probation Search Results" %}
-            <h2 class="govuk-heading-l" id = "tab-heading-probation">{{ countText }}</h2>
+            <div class="govuk-grid-row">
```
andrewrlee, line 58: There's quite a lot of duplication in content here.. bit of spot the difference, but I can't see whats the difference between the if/else block?

assets/js/search/setActiveTab.js
gurps-singh: I did ask Steph about this following the thread and she mentioned it was something we can look at following feedback on search.

assets/js/search/highlightSearchTerm.js
gurps-singh: Done

server/routes/search/handlers/probationSearch.ts
gurps-singh: Done

server/views/pages/search/probationSearch/probationSearch.njk
gurps-singh: De-duped.

server/routes/search/handlers/probationSearch.ts
```diff
@@ -1,42 +1,54 @@
...
+      searchResponse = await this.searchService.getProbationSearchResults(queryTerm, deliusStaffIdentifier)
     }
 
-    if (searchResponse.onProbationCount === 0 || searchResponse.onProbationCount > 1) {
-      onProbationCountText = `People on probation (${searchResponse.onProbationCount} results)`
+    const backLink = this.getBackLink(previousCaseloadPage)
+
+    let activeTab: string
+
+    if (searchResponse.onProbationCount === 0 || searchResponse.inPrisonCount >= searchResponse.onProbationCount) {
```
andrewrlee, line 30: is the `searchResponse.onProbationCount === 0 ` clause necessary? 

There's generally quite a lot of mutable references in the PR, doesn't require action now but it's generally worth trying to avoid them wherever possible.. there's a few ways around this - ternarys, "or" ing on falsy values, or extracting functions, or removing the conditional entirely

andrewrlee (approved):
LGTM

PR #670, opened 2023-09-18
Cvsl 1327 dps header update for cvl

nishanthmahasamudram:


server/app.ts
```diff
@@ -48,6 +49,7 @@ export default function createApp(services: Services): express.Application {
   }
 
   app.use(authorisationMiddleware)
+  app.get('*', getFrontendComponents(services))
```
andrewrlee, line 52: I think this will mean we load the header and footer for the prisoner image endpoint

nishanthmahasamudram: This is to load header and footer components from URL(https://frontend-components-dev.hmpps.service.justice.gov.uk) and assign to res.locals. So, that as soon as our app loads we can fetch and populate.

andrewrlee (approved):
LGTM

server/app.ts
nishanthmahasamudram: Raised a ticket to tidy up this https://dsdmoj.atlassian.net/browse/CVSL-1374

PR #677, opened 2023-09-21
Revert "Cvsl 1327 dps header update for cvl"

andrewrlee:
Reverts ministryofjustice/create-and-vary-a-licence#670


nishanthmahasamudram (approved):
Reverting header and footer changes as we are not able to navigate back to DPS when in service.

PR #680, opened 2023-09-21
CVSL-1367 Allowing 404 errors to return null on getLatestHdcStatus endpoint

SebNegahbanMoJ:
The method that calls `getLatestHdcStatus` already has handling for if the return value is null, so no additional logic was needed there.


andrewrlee (approved):
LGTM

PR #683, opened 2023-09-26
CVSL-1250 add call to new endpoint in retriever

gurps-singh:
This PR makes changes to the existing bank holiday feed service to utilise the new bank holiday endpoint from the API.


integration_tests/mockApis/licence.ts
```diff
@@ -1168,4 +1168,18 @@ export default {
       },
     })
   },
+
+  stubGetBankHolidays: (): SuperAgentRequest => {
```
andrewrlee, line 1172: ah.. not actually passing dates through to jsonBody here

gurps-singh: Yup, spotted that yesterday. Done.

andrewrlee (approved):
LGTM

PR #685, opened 2023-09-26
CVSL-1384 Adding explicit check to determine which licence to load for CA view

SebNegahbanMoJ:
This shouldn't functionally make any difference, as `_.head(c.licences)` would likely always return the `APPROVED` licence anyway. This is due to `c.licences` being sorted by `id`, which would typically put the APPROVED licence before any subsequent edits.

I thought it better to solidify this logic though, both for readability and for preventing potential edge-cases.


SebNegahbanMoJ:
Needs some more investigation/work, so closing for now

PR #686, opened 2023-09-28
CVSL-999 enable new PP search in all environments

gurps-singh:
This PR is to enable the use of the new PP search across all environments following successful QA.


andrewrlee (approved):
LGTM

PR #700, opened 2023-10-12
CVSL-1343-CVL - error searching on no prison numbers

nishanthmahasamudram:


server/listeners/eventHandlers/prisonEvents/datesChangedEventHandler.ts
```diff
@@ -18,27 +18,38 @@ export default class DatesChangedEventHandler {
 
     const nomisId =
       offenderIdDisplay ||
-      (await this.prisonerService.searchPrisonersByBookingIds([bookingId])).map(o => o.prisonerNumber).pop()
+      (!!bookingId &&
```
andrewrlee, line 21: could you push this check into the client?

server/routes/creatingLicences/handlers/confirmCreate.ts
```diff
@@ -50,7 +50,7 @@ export default class ConfirmCreateRoutes {
     const { answer } = req.body
     const backLink = req.session.returnToCase
 
-    if (answer === YesOrNo.YES) {
+    if (answer === YesOrNo.YES && !!nomisId) {
```
andrewrlee, line 53: can this be null? wouldn't the path fail to render if this was missing?

server/listeners/eventHandlers/prisonEvents/datesChangedEventHandler.ts
nishanthmahasamudram: This is event handler, so I added it here.

server/routes/creatingLicences/handlers/confirmCreate.ts
nishanthmahasamudram: Yes, if it is null then it will redirect to caseload page.
This is POST method is triggered from front end button click, so it should have the nomisId.

server/routes/support/handlers/offenderAudit.ts
```diff
@@ -27,7 +27,7 @@ export default class OffenderAuditRoutes {
     res.render('pages/support/offenderAudit', {
       prisonerDetail: {
         id: nomsId,
-        name: convertToTitleCase(`${prisonerDetail.firstName} ${prisonerDetail.lastName}`),
+        name: (!!prisonerDetail && convertToTitleCase(`${prisonerDetail.firstName} ${prisonerDetail.lastName}`)) || '',
```
andrewrlee, line 30: I think these could be undefined still

andrewrlee: if no prisoner is found

andrewrlee (approved):
LGTM

server/routes/support/handlers/offenderAudit.ts
nishanthmahasamudram: Yes, if it is undefined it will return ''(empty string instead of undefined undefined)

PR #701, opened 2023-10-12
CVSL-1162 Updating curfew input page and summary page to match new designs

SebNegahbanMoJ:
Also changing routing as necessary


server/routes/manageConditions/handlers/curfew/curfewRoutes.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    const formResponses = conditionInstances.reduce((acc, instance, index) => {
+      instance.data.map(conditionData => {
+        const key = index === 0 ? conditionData.field : `${conditionData.field}${index + 1}`
+        acc[key] = conditionData.value
+        return conditionData.value
+      })
+      return acc
+    }, {})
```
SebNegahbanMoJ, line 39: Instances of the curfew condition are stored as separate conditions in the DB, but all need to be displayed on one page. This reducer takes all instances and creates one object that houses all of the input data in a way that is compatible with the formbuilder components

server/routes/manageConditions/handlers/curfew/curfewRoutes.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (inputs.numberOfCurfews === 'Two curfews') {
+      await this.addCurfewCondition(licence, user, conditionCode, inputs.curfewStart, inputs.curfewEnd, inputs)
+      await this.addCurfewCondition(licence, user, conditionCode, inputs.curfewStart2, inputs.curfewEnd2, inputs)
+    } else if (inputs.numberOfCurfews === 'Three curfews') {
+      await this.addCurfewCondition(licence, user, conditionCode, inputs.curfewStart, inputs.curfewEnd, inputs)
+      await this.addCurfewCondition(licence, user, conditionCode, inputs.curfewStart2, inputs.curfewEnd2, inputs)
+      await this.addCurfewCondition(licence, user, conditionCode, inputs.curfewStart3, inputs.curfewEnd3, inputs)
+    } else {
+      await this.addCurfewCondition(licence, user, conditionCode, inputs.curfewStart, inputs.curfewEnd, inputs)
+    }
```
SebNegahbanMoJ, line 60: The reverse to the GET method, where a single page of inputs potentially needs to be split into multiple instances of the condition.
I thought about making a method that parsed the selected number of curfews, or relied on null/undefined checks to determine how many instances of the condition to add, but in the end I thought it's a minor efficiency saving at the cost of a significant amount of readability.

That said, I'm happy to refactor into a more succinct method if desired.

server/routes/manageConditions/handlers/curfew/curfewRoutes.ts
```diff
@@ -0,0 +1,116 @@
...
+    )
+
+    const request = {
+      conditionCode,
+      conditionCategory: condition.categoryShort || condition.category,
+      conditionText: condition.text,
+      conditionType: type,
+      expandedText: condition.tpl,
+      sequence,
+    } as AddAdditionalConditionRequest
```
SebNegahbanMoJ, line 92: Taken from `fileListUploadRoutes.ts`

server/routes/manageConditions/types/additionalConditionInputs/curfewTerms.ts
```diff
@@ -27,4 +79,28 @@ class CurfewTerms {
...
+    return enteredValues[enteredValues.length - 3]
+  }
+  if (numberOfCurfews === 'Two curfews') {
+    return enteredValues[enteredValues.length - 2]
+  }
+  if (numberOfCurfews === 'Three curfews') {
+    return enteredValues[enteredValues.length - 1]
+  }
+  return undefined
+}
```
SebNegahbanMoJ, line 104: With the `SimpleTime` objects built (via `splitTimes` above), all but one of the created objects will be invalid, as it will have been built with `undefined` entries. As the arrays of `SimpleTime`s are in order though, we can always tell which one we should use based on the selected number of curfews.
For example, when `Two curfews` is selected (and entered), `enteredValues` will look like this:
image
So we can use `enteredValues.length - 2` to consistently select the `ValidSimpleTime`. The same logic applies to the other cases.

integration_tests/integration/createLicence.cy.ts
```diff
@@ -61,6 +61,7 @@ context('Create a licence', () => {
         .nextCondition()
         .checkBoxes()
         .nextCondition(false) // aria-expanded attribute causes issues with Axe
+        .selectRadio('One curfew')
```
andrewrlee, line 64: Would be good to have better test coverage for multiple conditions... lack of test coverage really hampers ability to maintain code mid to long term. Restricts ability to safely refactor code

server/routes/manageConditions/handlers/curfew/curfewRoutes.test.ts
```diff
@@ -0,0 +1,252 @@
...
+
+      expect(licenceService.deleteAdditionalConditionsByCode).toHaveBeenCalledWith(
+        CURFEW_CONDITION_CODE,
+        res.locals.licence,
+        res.locals.user
+      )
+    })
+
+    it('adds a number of instances of the curfew condition equal to the user-selection', async () => {
+      jest.spyOn(handler, 'addCurfewCondition')
```
andrewrlee, line 166: This pattern of spying the test under class and verifying its internal implementation is not great practice.
When doing this style of test its better to test the interactions with the unit under test's collaborators. 

Can you verify the licence service was called appropriately?

server/routes/manageConditions/handlers/curfew/curfewRoutes.test.ts
```diff
@@ -0,0 +1,252 @@
...
+      req.query.fromReview = 'true'
+      await handler.POST(req, res)
+      expect(res.redirect).toHaveBeenCalledWith(
+        '/licence/create/id/1/additional-licence-conditions/callback?fromReview=true'
+      )
+    })
+  })
+
+  describe('addCurfewCondition', () => {
+    conditionService.getAdditionalConditionType.mockResolvedValue(LicenceType.AP)
```
andrewrlee, line 189: better to do this within a before each - this is only called once, when the module is imported which means its easily to inadvertently add dependencies in the order in which tests are run.

server/routes/manageConditions/handlers/curfew/curfewRoutes.ts
andrewrlee: this is a bit mind melty.. did you mean to use a `foreach`? you seem to be not using the resullts of the `map`?

I wonder if a `flatmap` to produce an array of `[key, value]` then an `Object.fromEntries` would be more elegant..

server/routes/manageConditions/handlers/curfew/curfewRoutes.ts
andrewrlee: Ideally we'd just have a `replace` endpoint on the backend..


I guess you could have done:
```ts
await this.addCurfewCondition(licence, user, conditionCode, inputs.curfewStart, inputs.curfewEnd, inputs)

if (two or three) {
   await this.addCurfewCondition(licence, user, conditionCode, inputs.curfewStart2, inputs.curfewEnd2, inputs)
}

if (three) {
     await this.addCurfewCondition(licence, user, conditionCode, inputs.curfewStart3, inputs.curfewEnd3, inputs)
}
```

server/routes/manageConditions/handlers/curfew/curfewRoutes.test.ts
SebNegahbanMoJ: Absolutely I can
I went with this approach because I'm testing `addCurfewCondition` below separately, and by testing calls to `licenceService` from here it could break if `addCurfewCondition` is refactored. That said, if this type of spying is poor practice, I'm happy to change it over.

justin-gardiner (approved):
lgtm

andrewrlee (approved):
LGTM

integration_tests/pages/additionalConditionInput.ts
```diff
@@ -86,8 +86,8 @@ export default class AdditionalConditionsInputPage extends Page {
   }
 
   nextCondition = (runAxe = true): AdditionalConditionsInputPage => {
-    cy.get(this.continueButtonId).click()
     cy.task('stubGetLicenceWithConditionToComplete', this.additionalConditionsToInput.shift())
+    cy.get(this.continueButtonId).click()
```
SebNegahbanMoJ, line 90: From what I can tell, this was previously not actually working correctly. The 'correct' condition wasn't being loaded before the input page rendered, but because all of the input pages were built the same and we have a lot of stubbing going on, it all worked. Now that I've had to add specific routed for Curfew though, it broke, as it tried to load the normal input page before recognising that the condition in question was curfew.
By switching these calls, it now correctly recognises which condition it's currently looking at

justin-gardiner (dismissed):
lgtm

andrewrlee (dismissed):
LGTM

andrewrlee (approved):
LGTM

gurps-singh (approved):

PR #721, opened 2023-10-19
CVSL-1162 Changing licenceComparator to generically handle multi-instance conditions

SebNegahbanMoJ:
ACO-view
image

COM-view
image

### DIFF
--- a/server/utils/licenceComparator.test.ts
+++ b/server/utils/licenceComparator.test.ts
@@ -18,13 +18,11 @@ describe('Licence Comparator', () => {
             code: '1',
             category: 'category1',
             expandedText: 'testCondition1',
-            uploadSummary: [],
           },
           {
             code: '2',
             category: 'category2',
             expandedText: 'testCondition2',
-            uploadSummary: [],
           },
         ],
         bespokeConditions: [{ text: 'bespoke1' }, { text: 'bespoke2' }],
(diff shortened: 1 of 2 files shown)


server/utils/licenceComparator.test.ts
```diff
@@ -467,7 +440,7 @@ describe('Licence Comparator', () => {
       licenceConditionsRemoved: [
         {
           category: 'Freedom of movement',
-          condition: 'Exclusion zones have been removed',
```
SebNegahbanMoJ, line 470: This is the only substantial (ie spec-impacting) change of this file, the others are only because `uploadSummary` is no longer used in this method

server/utils/licenceComparator.ts
```diff
@@ -82,68 +103,14 @@ const compareAdditionalConditionSet = (
...
-  if (originalExclusionsZoneConditionsSet.length === 0 && variedExclusionZoneConditionsSet.length > 0) {
-    variedConditionsBuilder.recordConditionAdded({
-      category: 'Freedom of movement',
-      condition: `${newZones}`,
-    })
-  } else if (changesToMultipleExclusionZones(variedExclusionZoneConditionsSet, originalExclusionsZoneConditionsSet)) {
-    if (newZones.length === 0) {
-      variedConditionsBuilder.recordConditionRemoved({
-        category: 'Freedom of movement',
-        condition: `Exclusion zones have been removed`,
```
SebNegahbanMoJ, line 101: This used to be a MEZ-specific method. Have confirmed with Tom that it's fine to remove this and instead to just list the text of the removed condition instances

justin-gardiner (approved):
lgtm

andrewrlee (approved):
LGTM

PR #724, opened 2023-10-19
CVSL-1415: Only set lastApprovedVersion on link if licence status is …

justin-gardiner:
…SUBMITTED.


nishanthmahasamudram (approved):
LGTM

PR #728, opened 2023-10-24
Cvsl 1238 friday release change weekend bank hols alert

nishanthmahasamudram:


server/@types/licenceApiImport/index.d.ts
```diff
@@ -1971,6 +1971,8 @@ export interface components {
        * @example 1.3
        */
       licenceVersion?: string
+      /** @description If CRD falls on Friday/Bank holiday/Weekend then it contains Earliest possible release date or CRD */
```
andrewrlee, line 1974: is this doc a bit wrong? should it be `it returns CRD unless the CRD falls on a friday or bank holiday in which case blah blah..` ?

If its wrong then obvs would need fixing in the backend PR.

nishanthmahasamudram: It is correct If CRD falls on Friday/bank holiday/weekend then we are calculating and returning early possible date else it will return CRD.

nishanthmahasamudram: `it returns CRD unless the CRD falls on a friday or bank holiday `  same as `If CRD falls on Friday/bank holiday/weekend then we are calculating and returning early possible date else it will return CRD`

nishanthmahasamudram: Here CRD means ARD/CRD.

nishanthmahasamudram: Updated to ### If ARD||CRD falls on Friday/Bank holiday/Weekend then it contains Earliest possible release date, otherwise contains ARD||CRD

andrewrlee (commented):
LGTM

andrewrlee (approved):
LGTM

PR #734, opened 2023-10-30
CVSL-1455 Adding OMU link from prisoner details

andrewrlee:


server/routes/support/handlers/omuEmailAddress.ts
```diff
@@ -22,6 +22,30 @@ export default class OmuEmailAddressRoutes {
...
+    }
+
+    return res.render('pages/support/manageOmuEmailAddress', {
+      prisonIdCurrent: prisonId,
+      prisonIdEdit: prisonId,
+      prisonIdDelete: prisonId,
+      emailCurrent,
+      email: emailCurrent,
+    })
+  }
```
andrewrlee, line 47: There's no test coverage for the support area, I left it as it was😬

louisenorris: 😬

PR #743, opened 2023-11-08
CVSL-868 add calendar picker to initial appointment page

louisenorris:
Screenshot 2023-11-09 at 4 55 53 pm


server/validators/dateIsAfterExpectedReleaseDate.ts
```diff
@@ -2,19 +2,20 @@ import { registerDecorator, ValidationArguments, ValidationOptions } from 'class
...
-  const dateIsAfterExpectedReleaseDate = async (date: SimpleDate, { object }: ValidationArguments) => {
+  const dateIsAfterExpectedReleaseDate = async (date: SimpleDate | DateString, { object }: ValidationArguments) => {
     const bankHolidays = await bankHolidayService.getEnglishAndWelshHolidays()
 
     const getWorkingDayOnOrBefore = (d: Moment): Moment => {
       return bankHolidays.isBankHolidayOrWeekend(d) ? getWorkingDayOnOrBefore(d.subtract(1, 'day')) : d
     }
 
-    const dateAsMoment = date.toMoment()
+    const dateAsMoment = typeof date === 'string' ? moment(date, 'DD/MM/YYYY') : date.toMoment()
```
andrewrlee, line 18: would be tempted to wrap the string in a DateString class, with a `toMoment`method and `@Expose` on the string.
Then these could work on things with `toMoment`s on it - rather than do extra parsing here.

server/utils/utils.ts
```diff
@@ -38,6 +39,33 @@ const hasAuthSource = (user: Express.User, source: string): boolean => user?.aut
...
+  const ampm = momentDt.format('a') === 'am' ? AmPm.AM : AmPm.PM
+  const simpleTime = new SimpleTime(momentDt.format('hh'), momentDt.format('mm'), ampm)
+  return DateTime.fromDateAndTime(dateString, simpleTime)
+}
+
+/**
+ * Converts a DateTime display value to a JSON string format dd/mm/yyyy hh:mm
+ * @param dt: DateTime
+ */
+const dateTimeToJson = (dt: DateTime): string | undefined => {
```
andrewrlee, line 61: would this be better as a function on DateTime?

andrewrlee (approved):
LGTM, maybe you could address comments in a follow up?

server/utils/utils.ts
louisenorris: sortd in a follow up PR

server/validators/dateIsAfterExpectedReleaseDate.ts
louisenorris: sorted in a follow up PR: https://github.com/ministryofjustice/create-and-vary-a-licence/pull/751

PR #750, opened 2023-11-14
CVSL-868 filename fix

louisenorris:
Decided to rename the script path for now rather than make the filename datePicker as the filename change wasn't being picked up as a git change. Will prob make the filename datePicker for consistency in a follow up PR when I address the other comments in the last PR: https://github.com/ministryofjustice/create-and-vary-a-licence/pull/743

This is a quick fix to get the script working in dev for now.


andrewrlee (approved):
lgtm

PR #751, opened 2023-11-15
CVSL-868 move toJson utils method to DateTime class

louisenorris:
Addressing comments made in https://github.com/ministryofjustice/create-and-vary-a-licence/pull/743 to move the dateTimeToJson method from 'utils' to be a `toJson` function on `DateTime`. Did the same with the `jsonToDateTime` function as thought that made sense. Have also wrapped the date string in a DateString class.


andrewrlee (approved):
LGTM!

PR #753, opened 2023-11-17
CVSL-1431 Adding skip button to appointment date/time and relevant conditions

SebNegahbanMoJ:
Depends on the [corresponding API PR](https://github.com/ministryofjustice/create-and-vary-a-licence-api/pull/373) to be merged first


server/routes/manageConditions/handlers/additionalLicenceConditionInput.ts
```diff
@@ -83,4 +83,19 @@ export default class AdditionalLicenceConditionInputRoutes {
 
     return res.redirect(redirect)
   }
+
+  SKIP = async (req: Request, res: Response): Promise<void> => {
+    const { user, licence } = res.locals
+    const { conditionId } = req.params
+
+    const condition = licence.additionalLicenceConditions.find(c => c.id === parseInt(conditionId, 10))
+    const conditionData = { conditionSkipped: '[DATE REQUIRED]' }
```
andrewrlee, line 92: meh... could call this `DATE_REQUIRED` and then it would be easier to migrate to an enum?

andrewrlee (dismissed):
Integration tests?

server/routes/manageConditions/handlers/additionalLicenceConditionInput.ts
SebNegahbanMoJ: I'm not sure what you mean by being easier to migrate to an enum, but I left it as `conditionSkipped` to make it slightly easier should we wish to add skipping functionality to any other conditions in future (that aren't necessarily date-specific) while leaving it generic in the DB.

andrewrlee: yes, totally agree with the variable named `conditionSkipped` - just meant if the value of the variable was `DATE_REQUIRED` then if we wanted to map it to an enum in the future then it would be trivial to do so. 
 
on the kotlin side: `val conditionSkipped = ConditionSkipReason.DATE_REQUIRED` 

(UPPER_CAMEL_CASE is a common way to represent codes/constants across programming languages, `[UPPER CASE]` is a bit bespoke)

andrewrlee (approved):
LGTM

gurps-singh (approved):

PR #756, opened 2023-11-22
CVSL-1433-CVL Amend rules Nomis overrides take precedence

nishanthmahasamudram:


SebNegahbanMoJ (commented):
We don't seem to be populating PRRD, PED, or NPD on licence creation, only on date change event.
We're also not explicitly reading any of the overrides on licence creation, is that due to limitations of offender search?

nishanthmahasamudram:
> We don't seem to be populating PRRD, PED, or NPD on licence creation, only on date change event. We're also not explicitly reading any of the overrides on licence creation, is that due to limitations of offender search?

Yes, this is because of the limitation of offender search as they are returning override value in a separate param.

nishanthmahasamudram:
> We don't seem to be populating PRRD, PED, or NPD on licence creation, only on date change event. We're also not explicitly reading any of the overrides on licence creation, is that due to limitations of offender search?

As per Andy's comments(Lets hold off setting the other dates in the database-  as we  would really need to do some kind of migration to back fill existing records which would be loads of work.) we are not adding PRRD, PED, or NPD to Licence table now and not reading them in both licence creation and data change events.

Also, just to confirm we don't need to explicitly read override dates on licence creation as Prison search API is returning override dates if available.

andrewrlee (approved):
LGTM

PR #757, opened 2023-11-22
CVSL-1475 add async error handling around validators

louisenorris:


server/middleware/validationMiddleware.ts
```diff
@@ -10,56 +11,61 @@ export type FieldValidationError = {
...
+        })
+      }
 
-    return res.redirect('back')
+      req.flash('validationErrors', JSON.stringify(flattenErrors(errors)))
+      req.flash('formResponses', JSON.stringify(req.body))
+
+      return res.redirect('back')
+    } catch (error) {
+      logger.error(error, `Failed to validate licence details for: ${req.params.nomisId}`)
```
andrewrlee, line 66: if we log the error then we'll be producing two error events in app insights as the error handler logs it too

Could wrap the error [using the cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#rethrowing_an_error_with_a_cause)

( wouldn't want to rethrow, just pass to `next`)

andrewrlee (approved):
LGTM

PR #778, opened 2023-12-21
Cvsl 1512 hard stop db changes support appointment changes in db

nishanthmahasamudram:


server/services/licenceService.ts
```diff
@@ -168,7 +169,8 @@ export default class LicenceService {
 
   async updateAppointmentTime(id: string, formData: DateTime, user: User): Promise<void> {
     const appointmentTime = DateTime.toJson(formData)
-    const requestBody = { appointmentTime } as AppointmentTimeRequest
+    const appointmentTimeType = AppointmentTimeType.SPECIFIC_DATE_TIME
```
nishanthmahasamudram, line 172: Hardcoded this to 'SPECIFIC_DATE_TIME' as we have only one option now.
Once the new functionality is implemented with three radio button options we can remove this here and send selected value from handlers.

nishanthmahasamudram:
Handled the default in backend.

PR #783, opened 2024-01-03
CVSL-1514 Moving cronjob values to new pattern and enabling in preprod and prod

SebNegahbanMoJ:
Disabled in dev and test environments because the dates of these cases are often unreliable, so could lead to mass-deactivation of useful test cases.


andrewrlee (approved):
LGTM

PR #786, opened 2024-01-04
CVSL-1506 Allowing CAs to edit initial appointment info

SebNegahbanMoJ:
Should only be merged after the [corresponding API PR](https://github.com/ministryofjustice/create-and-vary-a-licence-api/pull/419)


server/routes/viewingLicences/handlers/viewLicence.ts
```diff
@@ -10,8 +10,15 @@ export default class ViewAndPrintLicenceRoutes {
...
     const { licence, user } = res.locals
-
     let warningMessage
+
+    const releaseDateString = licence.actualReleaseDate || licence.conditionalReleaseDate
+    const releaseDate = parse(releaseDateString, 'dd/MM/yyyy', new Date())
+    const isInHardStopWindow = isWithinInterval(startOfDay(new Date()), {
+      start: sub(releaseDate, { days: 2 }),
+      end: releaseDate,
+    })
```
SebNegahbanMoJ, line 20: This can be removed/replaced when the hard-stop logic is fully implemented on the licence representation

server/routes/initialAppointment/handlers/initialMeetingTime.ts
```diff
@@ -17,6 +21,7 @@ export default class InitialMeetingTimeRoutes {
       appointmentTimeType,
       releaseIsOnBankHolidayOrWeekend: licence.isEligibleForEarlyRelease,
       skipUrl: this.getNextPage(licence.id.toString(), licence.typeCode, req),
+      canSkip: this.userType === UserType.PROBATION,
```
andrewrlee, line 24: I think this could be based solely on whether the value is populated or not? 
We don't want to let them skip if they've previously provided a value (as it doesn't do anything and is confusing for the user)

server/routes/initialAppointment/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    const controller = new InitialMeetingTimeRoutes(licenceService, UserType.PROBATION)
+    get('/create/id/:licenceId/initial-meeting-time', controller.GET)
+    post('/create/id/:licenceId/initial-meeting-time', controller.POST, DateTime)
+  }
+
+  {
+    const controller = new InitialMeetingTimeRoutes(licenceService, UserType.PRISON)
+    get('/view/id/:licenceId/initial-meeting-time', controller.GET)
+    post('/view/id/:licenceId/initial-meeting-time', controller.POST, DateTime)
+  }
```
andrewrlee, line 92: nice!

server/routes/initialAppointment/handlers/initialMeetingTime.ts
SebNegahbanMoJ: I can add that check, but I think it'll need to be in addition to the existing check as I don't believe we're allowing CAs to ever skip this option (eg if they go through the hard-stop licence creation, we don't want them to be able to skip)

server/routes/initialAppointment/handlers/initialMeetingTime.ts
```diff
@@ -17,6 +21,7 @@ export default class InitialMeetingTimeRoutes {
       appointmentTimeType,
       releaseIsOnBankHolidayOrWeekend: licence.isEligibleForEarlyRelease,
       skipUrl: this.getNextPage(licence.id.toString(), licence.typeCode, req),
+      canSkip: this.userType === UserType.PROBATION && !licence.appointmentTimeType,
```
SebNegahbanMoJ, line 24: Added a check of `!licence.appointmentTimeType` rather than checking `appointmentTime` directly, as in if the user selected one of the non-time-specific options (eg to report immediately), `appointmentTime` will be null but the question should still be non-skippable.

It seems like there was a migration run to backfill the `appointmentTimeType` for previously-entered dates, so I don't think this should pose an issue for licences that existed before `appointmentTimeType` was added

andrewrlee: test?!

andrewrlee (approved):
LGTM

PR #795, opened 2024-01-16
Cvsl 1444 hard stop split the tabs

nishanthmahasamudram:


server/@types/hardstopCutoffDate.ts
```diff
@@ -0,0 +1,3 @@
+export default class HardStopCutoffDate {
```
andrewrlee, line 1: just need to regenerate the types and expose rather than create this
running: `./generate-licence-api-types.sh` and update `licenceApiClientTypes.ts`

helm_deploy/create-and-vary-a-licence/values.yaml
```diff
@@ -54,6 +54,7 @@ generic-service:
     COMMON_COMPONENTS_ENABLED: 'true'
     MANAGE_USERS_API_URL: 'https://manage-users-api-dev.hmpps.service.justice.gov.uk'
     FRIDAY_RELEASE_POLICY: 'https://www.gov.uk/government/publications/discretionary-fridaypre-bank-holiday-release-scheme-policy-framework'
+    USE_NEW_CASELOAD_VIEW_FOR_HARD_STOP: "false"
```
andrewrlee, line 57: could set to true in `values-dev`? or `values-test1`?

integration_tests/mockApis/community.ts
```diff
@@ -383,4 +383,18 @@ export default {
...
+  stubGetCutOffDateForLicenceTimeOut: (): SuperAgentRequest => {
+    return stubFor({
+      request: {
+        method: 'GET',
+        urlPattern: `/licence/getCutOffDate`,
+      },
+      response: {
+        status: 200,
+        headers: { 'Content-Type': 'application/json;charset=UTF-8' },
+        jsonBody: {},
```
andrewrlee, line 396: shouldn't this return a date or is this in progress?

server/licences/licenceStatus.ts
```diff
@@ -45,7 +45,7 @@ const statusConfig: Record<LicenceStatus, LicenceStatusConfig> = {
   NOT_STARTED: {
     label: 'Not started',
     description: 'Not started',
-    colour: 'grey',
+    colour: 'yellow',
```
andrewrlee, line 48: is this used across prisons and probation? 
This presumably would be active even if the feature was disabled - so released immediately?

Should we hide this change behind the toggle too?

server/licences/licenceStatus.ts
```diff
@@ -45,7 +46,7 @@ const statusConfig: Record<LicenceStatus, LicenceStatusConfig> = {
   NOT_STARTED: {
     label: 'Not started',
     description: 'Not started',
-    colour: 'grey',
+    colour: (!!config.HardStopEnabled && 'yellow') || 'grey',
```
andrewrlee, line 49: meh.. ternary? `config.HardStopEnabled ? 'yellow' : 'grey'`

server/views/pages/index.njk
```diff
@@ -36,13 +36,13 @@
...
-                    <div class="govuk-grid-column-one-third card-group__item">
+                    <div class="govuk-grid-column-one-third card-group__item govuk-grid-column-one-third card-group__item">
                         {{ card({
                             "id": "viewLicenceCard",
                             "href": "/licence/view/cases",
                             "clickable": "true",
-                            "heading": "Print a licence",
-                            "description": "Print standard determinate licences and post sentence supervision orders created by community offender managers.",
+                            "heading": "Print or generate a licence",
+                            "description": "Print licences and post sentence supervision orders or generate standard licences.",
```
andrewrlee, line 45: content changes should be behind feature toggle?

server/views/partials/viewCases.njk
```diff
@@ -0,0 +1,130 @@
...
+    {% endmacro %}
+    {% for case in cases %}
+        {% if case.probationPractitioner %}
+            {% set comDetailsHtml = '<a class="govuk-link" href="/licence/view/probation-practitioner/staffCode/' + case.probationPractitioner.staffCode + '">' + case.probationPractitioner.name + '</a>' %}
+        {% else %}
+            {% set comDetailsHtml = 'Unallocated' %}
+        {% endif %}
+        {% if case.licenceVersionOf and (case.licenceStatus === 'IN_PROGRESS' or case.licenceStatus === 'SUBMITTED') %}
+            {% set licenceStatusHtml = statusBadge({ status: case.licenceStatus, size: 'small' }, statusConfig) +
+                                    "<div class='govuk-!-margin-top-2'><svg class='moj-banner__icon' fill='currentColor' role='presentation'" +
```
andrewrlee, line 32: not sure if it existed before but would suggest extracting this svg into its own macro with a good name rather than embedding within a string within a set

andrewrlee: is there a reason why you aren't using the component? 
https://design-patterns.service.justice.gov.uk/components/banner/

(it's very difficult to understand what's going on here due to the complexity of the nesting - might help to break it into separate macros)

server/config.ts
```diff
@@ -322,4 +322,5 @@ export default {
     'https://www.gov.uk/government/publications/discretionary-fridaypre-bank-holiday-release-scheme-policy-framework',
     requiredInProduction
   ),
+  HardStopEnabled: get('HARD_STOP_ENABLED', 'false', requiredInProduction) === 'true',
```
andrewrlee, line 325: should be `lowerCamelCase` rather than `UpperCamelCase`?

server/utils/nunjucksSetup.ts
```diff
@@ -119,6 +119,10 @@ export function registerNunjucks(app?: express.Express): Environment {
     return appointmentTimeType[type]
   })
 
+  njkEnv.addFilter('beforeOrAfterHardStop', (licences: Record<string, unknown>[], isHardstop: boolean) => {
+    return licences.filter(c => c.hardStop === isHardstop)
+  })
```
andrewrlee, line 124: I think we've talk before about the disadvantages of using boolean flags: 

`data | beforeOrAfterHardStop(true)`
vs
`data | beforeOrAfterHardStop(false)`

How are you supposed to know what each call means? 

Would probably just split it out into two separate methods

server/utils/utils.test.ts
```diff
@@ -510,3 +511,13 @@ describe('Get prisoner release date from Nomis', () => {
...
   })
 })
+
+describe('Check if release date before cutoff date', () => {
+  it('should return true if release date is before cutoff date', () => {
+    expect(isReleaseDateBeforeCutOffDate('06/12/2023', '05/12/2023')).toBeTruthy()
+  })
+
+  it('should return false if release date is after cutoff date', () => {
+    expect(isReleaseDateBeforeCutOffDate('04/12/2023', '05/12/2023')).toBeFalsy()
```
andrewrlee, line 521: The tests should demo the behaviour and we're missing an important comparison - what if the dates are the same?

server/views/pages/view/cases.njk
```diff
@@ -83,100 +87,197 @@
 {% if probationView %} {% set probationViewUrlQuery = '?view=probation' %} {% endif %}
 
 {% block content %}
-    <div class="govuk-grid-row">
-        <div class="govuk-grid-column-three-quarters">
-            <h1 class="govuk-heading-xl">Select a person to view or print licences and supervision orders</h1>
+    {% if HardStopEnabled %}
```
andrewrlee, line 90: I'm finding this really difficult to read, maybe splitting out some macros would help?
```
 {% if HardStopEnabled %}
   { call hardstop macro }
else
   { call non-hardstop macro }
end
```

andrewrlee (approved):
LGTM

PR #803, opened 2024-01-25
CVSL-1444-date-fix

nishanthmahasamudram:


server/utils/utils.ts
```diff
@@ -196,7 +196,7 @@ const selectReleaseDate = (nomisRecord: Prisoner) => {
 }
 
 const isReleaseDateBeforeCutOffDate = (cutOffDate: string, releaseDate: string): boolean => {
-  return isBefore(new Date(releaseDate), new Date(cutOffDate))
+  return moment(releaseDate, 'DD/MM/YYYY').diff(moment(cutOffDate, 'DD/MM/YYYY')) <= 0
```
andrewrlee, line 199: think we're trying to avoid using moment where possible, is it not possible to use the date-fns version?

andrewrlee (approved):
LGTM

PR #805, opened 2024-01-25
CVSL-1587 Removing skip button from initial appointment date entry in hard stop window

SebNegahbanMoJ:
**Currently set to merge into #786. This will need testing and merging before this can be targeted onto main and merged**


server/routes/initialAppointment/handlers/initialMeetingTime.ts
```diff
@@ -21,7 +23,11 @@ export default class InitialMeetingTimeRoutes {
       appointmentTimeType,
       releaseIsOnBankHolidayOrWeekend: licence.isEligibleForEarlyRelease,
       skipUrl: this.getNextPage(licence.id.toString(), licence.typeCode, req),
-      canSkip: this.userType === UserType.PROBATION && !licence.appointmentTimeType,
+      canSkip:
+        this.userType === UserType.PROBATION &&
+        !licence.appointmentTimeType &&
+        licence.kind !== LicenceKind.VARIATION &&
+        !(config.hardStopEnabled && licence.isInHardStopPeriod),
```
andrewrlee, line 30: are people going to be able to access this screen when a licence is being varied when a value would not be set?
 
I think this could be simplified solely to `!licence.appointmentTimeType` - we discussed this whilst you were off and UX were happy with that.

SebNegahbanMoJ: For completeness, access to the `initialMeetingTime` check is now controlled more tightly to cover user type, licence kind, and hard-stop status. This means we no longer need these additional checks here, so the `canSkip` check can be reduced to just `!licence.appointmentTimeType`

andrewrlee (approved):
LGTM

PR #813, opened 2024-02-06
CVSL-1587 Adding redirects to access-denied from initial appointment pages

SebNegahbanMoJ:
Prison users should only be able to access initial appt pages during the hard stop window.
Probation users shouldn't be able to access the initial appt pages during the hard stop window.

Rather than just relying on showing/hiding the `change` buttons, we thought it better to explicitly prevent access (to prevent issues relating to bookmarks etc)


andrewrlee (dismissed):
LGTM, seems like the sort of thing though seems like the sort of thing that should be in some shared middleware

PR #817, opened 2024-02-16
CVSL-1675 Prevent PP from editing licence in hard stop period

SebNegahbanMoJ:
Light content changes will follow. As this work is hidden behind the hard stop feature flag, it's safe to be released without the new content.

PR #819, opened 2024-02-19
Splitting out addition of isInHardStopPeriod utils function to fix live issue

SebNegahbanMoJ:
Moving some of the changes from #817 to fix an oversight that is current causing a live issue


andrewrlee (approved):
LGTM

PR #826, opened 2024-02-23
Cvsl 1446 hard stop licence details pages displayed to case admin when licence in progress submitted or approved

nishanthmahasamudram:


server/routes/initialAppointment/handlers/hardStop/initialMeetingPlace.ts
```diff
@@ -23,6 +23,10 @@ export default class InitialMeetingPlaceRoutes {
     await this.licenceService.updateAppointmentAddress(licenceId, req.body, user)
 
     flashInitialApptUpdatedMessage(req, licence, this.userType)
-    res.redirect(`/licence/hard-stop/create/id/${licenceId}/initial-meeting-contact`)
+    if (req.originalUrl?.includes('edit')) {
```
andrewrlee, line 26: rather than check originalUrl it might be better to check path as that will only include path and not any query parameters?

server/routes/initialAppointment/handlers/hardStop/index.ts
```diff
@@ -47,20 +49,49 @@ export default function Index({ licenceService, conditionService }: Services): R
...
+  }
   {
     const controller = new InitialMeetingPlaceRoutes(licenceService, UserType.PRISON)
     get('/create/id/:licenceId/initial-meeting-place', controller.GET)
     post('/create/id/:licenceId/initial-meeting-place', controller.POST, Address)
   }
+  {
+    const controller = new InitialMeetingPlaceRoutes(licenceService, UserType.PRISON)
+    get('/edit/id/:licenceId/initial-meeting-place', controller.GET)
+    post('/edit/id/:licenceId/initial-meeting-place', controller.POST, Address)
```
andrewrlee, line 65: if not parameterising these then may be you could do `(create|edit)/id/:licenceId/initial-meeting-place`

server/views/pages/view/view.njk
```diff
@@ -28,56 +29,81 @@
...
-                   attributes: { 'data-qa': 'return-to-view-list' }
-                }) }}
+                {% if licence.kind === 'HARD_STOP' %}
+                    {{ hardStopInductionMeetingDetails(licence, isEditableByPrison, [], null, isPrisonUser) }}
+                {% else %}
+                    {{ inductionMeetingDetails(licence, isEditableByPrison, [], null, isPrisonUser) }}
+                {% endif %}
+
+                {% if licence.kind === 'HARD_STOP' %}
+                    <h2 class="govuk-heading-m">Licence conditions</h2>
```
andrewrlee, line 63: would it make sense for this text block to be a macro? think it might be used in a few places?

server/views/pages/standardLicenceConditions.njk
```diff
@@ -0,0 +1,12 @@
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+
+{%  macro standardLicenceConditions() %}
```
andrewrlee, line 3: can we call this immediate release conditions? 
Just that standard licence conditions sounds like the standard conditions you get on any licence

andrewrlee (dismissed):
LGTM

server/routes/viewingLicences/handlers/viewLicence.ts
```diff
@@ -69,6 +73,23 @@ export default class ViewAndPrintLicenceRoutes {
...
+
+    const pduHeads = await this.communityService.getPduHeads(licence.probationPduCode).then(p =>
+      p.map(c => {
+        return {
+          name: `${c.staff.forenames} ${c.staff.surname}`,
+          email: c.email,
+        }
+      })
+    )
+    await this.licenceService.submitVariation(licenceId, pduHeads, user)
```
SebNegahbanMoJ, line 88: This is for CAs creating hard-stop licences, correct?
Should we be contacting the PDU heads for these creations? PDU heads are the heads of probation, and I thought are typically only contacted to approve variations

server/views/pages/immediateReleaseConditions.njk
```diff
@@ -0,0 +1,12 @@
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+
+{%  macro immediateReleaseConditions() %}
+    <p class="govuk-body">By default, the licence will automatically contain the following restrictions:</p>
+    <ul class="govuk-list govuk-list--bullet standard-conditions">
+        <li>standard conditions</li>
+        <li>the following additional condition:</li>
+        {{ govukInsetText({
+            text: "Not to approach or communicate with any victims of your offences without the prior approval of your supervising officer."
```
SebNegahbanMoJ, line 9: Have we confirmed with UCD what the behaviour should be when a licence is PSS-only, or are we covering that in a separate ticket?

server/views/partials/inductionMeetingDetailsSummary.njk
```diff
@@ -2,12 +2,12 @@
...
     {% set apptDateTimeEntered = licence.appointmentTime or licence.appointmentTimeType | length > 0 %}
-    {% set routePath = 'view' if isPrisonUser else 'create'%}
-    {% set routeAppend = '?fromReview=true' if not isPrisonUser %}
+    {% set routePath = 'view' if isPrisonUser else 'create' %}
+    {% set routeAppend = '?fromReview=true' if hardStopEnabled or not isPrisonUser %}
 
     <div class="govuk-grid-row check-answers-header">
         <div class="govuk-grid-column-three-quarters">
-            <h2 class="govuk-!-margin-bottom-3 govuk-heading-m">Initial appointment details</h2>
+            <h2 class="govuk-!-margin-bottom-3 govuk-heading-m">Initial appointment details{{isPrisonUser}}</h2>
```
SebNegahbanMoJ, line 10: Was this left in by mistake?

nishanthmahasamudram: Sorry.
It was by mistake.

server/routes/viewingLicences/handlers/viewLicence.ts
nishanthmahasamudram: Shall I pass CA details instead?

SebNegahbanMoJ: I don't think you need to submit anything
I suppose it depends what the ticket says as to whether need to be emailed, but I would assume you can just use `submitLicence` rather than `submitVariation`. They both call the same endpoint in the API, and looking at that endpoint it seems like the only time the second param (`NotifyRequest` is used is when the licence kind is `VARIATION`, so it's currently redundant anyway.

If you want to email the prison approver, it seems like you'd need to make changes to the current API endpoint regardless. I think we already send emails through some mechanism when a licence is submitted though, so it may be worth checking if just using `submitLicence` would result in an email being sent anyway, in which case you probably wouldn't need to add anything.

andrewrlee: that makes sense - emails is a separate ticket also 👍

server/views/pages/immediateReleaseConditions.njk
andrewrlee: Separate ticket

andrewrlee: It's CVSL-1691

SebNegahbanMoJ (approved):

PR #832, opened 2024-02-29
CVSL-1731 Fixing local stack issue

andrewrlee:
Also:
* Removing usages of deprecated test method
* Moving to use project version of node for npm security and outdated checks
* Fixing little issue on support page where the page wouldn't load if there was no associated com for an offender

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@7
+  hmpps: ministryofjustice/hmpps@8
   codecov: codecov/codecov@3.2.5
 
 parameters:
@@ -240,10 +240,12 @@ workflows:
     jobs:
       - hmpps/npm_outdated:
           slack_channel: << pipeline.parameters.alerts-slack-channel >>
+          node_tag: << pipeline.parameters.node-version >>
           context:
             - hmpps-common-vars
       - hmpps/npm_security_audit:
           slack_channel: << pipeline.parameters.alerts-slack-channel >>
+          node_tag: << pipeline.parameters.node-version >>
           context:
             - hmpps-common-vars
       - hmpps/trivy_latest_scan:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Stage: base image
-FROM node:20.8-bullseye-slim as base
+FROM node:20.11-bookworm-slim as base
 
 ARG BUILD_NUMBER=1_0_0
 ARG GIT_REF=not-available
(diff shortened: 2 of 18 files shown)


.circleci/config.yml
```diff
@@ -240,10 +240,12 @@ workflows:
     jobs:
       - hmpps/npm_outdated:
           slack_channel: << pipeline.parameters.alerts-slack-channel >>
+          node_tag: << pipeline.parameters.node-version >>
```
andrewrlee, line 243: these were defaulting to node 16 rather than node 20 which is what we're currently on!

Dockerfile
```diff
@@ -30,9 +30,6 @@ FROM base as build
 ARG BUILD_NUMBER=1_0_0
 ARG GIT_REF=not-available
 
-RUN apt-get update && \
-    apt-get install -y make python g++
```
andrewrlee, line 34: python etc.. shouldn't be needed

package.json
```diff
@@ -26,8 +26,9 @@
     "integrationTest": "jest --runInBand --testPathPattern=/integration_tests/.*",
     "security_audit": "better-npm-audit audit",
     "int-test": "cypress run --config video=false",
-    "int-test-ui": "cypress open",
-    "clean": "rm -rf dist build node_modules stylesheets coverage"
+    "int-test-ui": "cypress open --e2e --browser chrome",
+    "clean": "rm -rf dist build node_modules stylesheets coverage",
+    "rebuild": "npm run clean && npm i && npm run build"
```
andrewrlee, line 31: this was inherited from the template project commit

server/middleware/fetchLicenceMiddleware.test.ts
```diff
@@ -82,7 +82,7 @@ describe('fetchLicenceMiddleware', () => {
     licenceService.getLicence.mockRejectedValue('Error')
 
     await middleware(req, res, next)
-    expect(next).toBeCalledWith('Error')
+    expect(next).toHaveBeenCalledWith('Error')
```
andrewrlee, line 85: replacing deprecated method in lots of places

server/routes/support/handlers/offenderDetail.ts
```diff
@@ -78,7 +78,7 @@ export default class OffenderDetailRoutes {
         name: probationPractitioner
           ? `${probationPractitioner.staff.forenames} ${probationPractitioner.staff.surname}`
           : '',
-        staffCode: probationPractitionerContact.staffCode,
+        staffCode: probationPractitionerContact?.staffCode,
```
andrewrlee, line 81: Fix issue with showing error page when no COM is assigned - it's a bit ugly ATM as it just shows a load of empty fields but better than blowing up

PR #838, opened 2024-03-06
CVSL-1740: Pull in changes from template project

andrewrlee:


Dockerfile
```diff
@@ -1,49 +1,53 @@
...
 WORKDIR /app
 
-# Cache breaking
-ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
-ENV GIT_REF ${GIT_REF:-xxxxxxxxxxxxxxxxxxx}
-    
+# Cache breaking and ensure required build / git args defined
+RUN test -n "$BUILD_NUMBER" || (echo "BUILD_NUMBER not set" && false)
+RUN test -n "$GIT_REF" || (echo "GIT_REF not set" && false)
+RUN test -n "$GIT_BRANCH" || (echo "GIT_BRANCH not set" && false)
```
andrewrlee, line 21: changes from template - the info endpoint will be a closer to spring boot and include info about git branch (which will generally be main for us!)

helm_deploy/create-and-vary-a-licence/values.yaml
```diff
@@ -28,8 +29,6 @@ generic-service:
     hosts:
       - app-hostname.local # override per environment
     tlsSecretName: create-and-vary-a-licence-cert
-    v1_2_enabled: true
-    v0_47_enabled: false
```
andrewrlee, line 32: This is default behaviour now

helm_deploy/create-and-vary-a-licence/values.yaml
```diff
@@ -17,6 +17,7 @@ generic-service:
   nameOverride: create-and-vary-a-licence
   replicaCount: 4
   serviceAccountName: create-and-vary-a-licence
+  productId: "DPS011"
```
andrewrlee, line 20: This will be exposed by the info endpoint

feature.env
```diff
@@ -16,7 +16,7 @@ SYSTEM_CLIENT_SECRET=clientsecret
 SQS_POLLING_WAIT_TIME_MS=2000
 REDIS_HOST=localhost
 HARD_STOP_ENABLED=true
-
+TOKEN_VERIFICATION_ENABLED=true
```
andrewrlee, line 19: tests now use token verification, the health checks now check partial failure works as expected

integration_tests/mockApis/auth.ts
```diff
@@ -103,29 +104,9 @@ const token = (authorities: string[], authSource: string) =>
       },
     },
   })
-
-const stubUserEmail = () =>
```
andrewrlee, line 107: moved this into dedicated manage-users api file

jobs/deactivateReleaseDatePassedLicences.ts
```diff
@@ -3,10 +3,10 @@
  * Do appinsights first to instrument the logger
  */
 import 'reflect-metadata'
-import { initialiseAppInsights, buildAppInsightsClient, flush } from '../server/utils/azureAppInsights'
+import { initialiseAppInsights, flush } from '../server/utils/azureAppInsights'
+import applicationInfo from '../server/applicationInfo'
 
-initialiseAppInsights()
-buildAppInsightsClient('create-and-vary-a-licence-deactivate-release-date-passed-licences-job')
+initialiseAppInsights(applicationInfo('create-and-vary-a-licence-deactivate-release-date-passed-licences-job'))
```
andrewrlee, line 9: simplified this set up a bit - the buildAppInsightsClient was really "configure default client that is used for everything" and we didn't actually need a reference to it, so merged the two.

server.ts
```diff
@@ -3,16 +3,20 @@
  * Do appinsights first as it does some magic instrumentation work, i.e. it affects other 'require's
  * In particular, applicationinsights automatically collects bunyan logs
  */
-import { initialiseAppInsights, buildAppInsightsClient } from './server/utils/azureAppInsights'
+import { initialiseAppInsights } from './server/utils/azureAppInsights'
+import createApplicationInfo from './server/applicationInfo'
 
-initialiseAppInsights()
-buildAppInsightsClient()
+const applicationInfo = createApplicationInfo()
```
andrewrlee, line 9: creating this here and then using dependency injection to pass through to where needed

server/applicationInfo.ts
```diff
@@ -2,11 +2,30 @@ import fs from 'fs'
 import path from 'path'
 import config from './config'
 
-const packageJson =
```
andrewrlee, line 5: there was some hackiness here as during the relative location of package.json would vary dependent on whether we were running tests (compiled on fly) vs running the app (compiled to dist folder)

This is no longer needed due to use of DI

server/applicationInfo.ts
```diff
@@ -2,11 +2,30 @@ import fs from 'fs'
...
+const getNameFromPackageJson = () => {
+  const packageJson = path.join(__dirname, '../../package.json')
+  const { name } = JSON.parse(fs.readFileSync(packageJson).toString())
+  return name
+}
 
-export default { applicationName, buildNumber, gitRef, gitShortHash: gitRef.substring(0, 7) }
+export default (overrideName?: string): ApplicationInfo => {
+  return {
+    applicationName: overrideName || getNameFromPackageJson(),
```
andrewrlee, line 24: little optimisation to only load name from the file system if no override was specified

server/data/healthCheck.test.ts
```diff
@@ -29,7 +29,7 @@ describe('Service healthcheck', () => {
       fakeServiceApi.get('/ping').reply(200, 'pong')
 
       const output = await healthcheck()
-      expect(output).toEqual('OK')
+      expect(output).toEqual('UP')
```
andrewrlee, line 32: new health endpoint shows UP rather than OK to match spring boot

server/data/index.ts
```diff
@@ -1,13 +1,3 @@
-/* eslint-disable import/first */
-/*
- * Do appinsights first as it does some magic instrumentation work, i.e. it affects other 'require's
- * In particular, applicationinsights automatically collects bunyan logs
- */
-import { initialiseAppInsights, buildAppInsightsClient } from '../utils/azureAppInsights'
-
-initialiseAppInsights()
```
andrewrlee, line 8: This is not needed as instrumentation happens in `server.ts` before anything
(this probably wasn't required before either)

nishanthmahasamudram (approved):
LGTM

PR #851, opened 2024-03-22
CVSL-1767 Allow prison users to not have email

andrewrlee:
This provides access to the serviceto people who have unverified email addresses. Reasoning that it's better to not store an unverified email address

### DIFF
--- a/server/middleware/populateCurrentUser.ts
+++ b/server/middleware/populateCurrentUser.ts
@@ -51,7 +51,6 @@ export default function populateCurrentUser(userService: UserService, licenceSer
             )
 
             const { email } = await userService.getUserEmail(user)
-            if (!email) throw new Error(`Failed to get email for: ${user.username}`)
             await licenceService.updatePrisonUserDetails({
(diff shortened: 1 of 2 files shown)


server/middleware/populateCurrentUser.ts
```diff
@@ -51,7 +51,6 @@ export default function populateCurrentUser(userService: UserService, licenceSer
             )
 
             const { email } = await userService.getUserEmail(user)
-            if (!email) throw new Error(`Failed to get email for: ${user.username}`)
             await licenceService.updatePrisonUserDetails({
               staffUsername: prisonUser.username,
               staffEmail: email,
```
nishanthmahasamudram, line 56: Will this pass undefined to api?

nishanthmahasamudram (approved):
LGTM

server/middleware/populateCurrentUser.ts
andrewrlee: yes, but that should be ok

PR #855, opened 2024-03-28
CVSL-1691-Hard Stop Display different content to describe restriction…

nishanthmahasamudram:
…s on a PSS only licence


server/services/licenceService.ts
```diff
@@ -594,4 +595,8 @@ export default class LicenceService {
   async reviewWithoutVariation(licenceId: number): Promise<void> {
     return this.licenceApiClient.reviewWithoutVariation(licenceId)
   }
+
+  async getPrisonerDetail(nomsId: string, user: User): Promise<CaseloadItem> {
```
andrewrlee, line 599: I'd keep this on prisoner service.. it's part of the prisoner domain

andrewrlee (approved):
LGTM

server/services/licenceService.ts
nishanthmahasamudram: will move it to prisoner service in my next PR.

PR #863, opened 2024-04-04
CVSL-1715 Content changes for CYA page for licences in hard stop period

SebNegahbanMoJ:
Added handling for an error response from `getOmuEmail` as this would be called on every `Check your answers` page, so is high-traffic and high-impact.


server/routes/creatingLicences/handlers/checkAnswers.ts
```diff
@@ -34,6 +34,7 @@ export default class CheckAnswersRoutes {
 
     const conditionsToDisplay = await this.conditionService.getAdditionalAPConditionsForSummaryAndPdf(licence, user)
     const bespokeConditionsToDisplay = await this.conditionService.getbespokeConditionsForSummaryAndPdf(licence, user)
+    const omuEmail = (await this.licenceService.getOmuEmail(licence.prisonCode, user))?.email
```
andrewrlee, line 37: isn't the error on the kotlin side, I thought you'd get a 500 response?

SebNegahbanMoJ: The usual error is when the OMU email is missing which I thought returned a 404, but now I'm doubting myself so will give it another test

Regardless, this should handle a 500 as `getOmuEmail` returns `null` for >=400 and < 500, or the error if it's outside of that range
In either case, <response>?.email should return null, and thereby be handled, no?

andrewrlee (approved):
LGTM

louisenorris (approved):
LGTM

PR #866, opened 2024-04-09
CVSL-1786 Adding standard parsing methods for dates

andrewrlee:
Added 3 helper functions for parsing dates and moved over existing code to use them:
* `parseIsoDate` => `2023-01-23`
* `parseCvlDate` => `23/01/2023`
* `parseCvlDateTime` 
  * `{ withSeconds: true }`  => `23/01/2023 13:45:50`
  * `{ withSeconds: false }`  => `23/01/2023 13:45`

### DIFF
--- a/server/routes/approvingLicences/handlers/approvalCases.ts
+++ b/server/routes/approvingLicences/handlers/approvalCases.ts
@@ -1,8 +1,8 @@
 import { Request, Response } from 'express'
 import _ from 'lodash'
-import { format, getUnixTime, isAfter, isValid, parse, subDays } from 'date-fns'
+import { format, getUnixTime, isAfter, isValid, subDays } from 'date-fns'
 import PrisonerService from '../../../services/prisonerService'
-import { convertToTitleCase, selectReleaseDate } from '../../../utils/utils'
+import { convertToTitleCase, parseCvlDateTime, selectReleaseDate } from '../../../utils/utils'
 import ApproverCaseloadService from '../../../services/approverCaseloadService'
 
 export default class ApprovalCaseRoutes {
@@ -31,7 +31,7 @@ export default class ApprovalCaseRoutes {
         let approvedDate
         if (_.head(c.licences).approvedDate) {
           approvedDate = format(
-            parse(_.head(c.licences).approvedDate, 'dd/MM/yyyy HH:mm:ss', new Date()),
+            parseCvlDateTime(_.head(c.licences).approvedDate, { withSeconds: true }),
             'dd MMMM yyyy'
           )
         }
@@ -41,7 +41,7 @@ export default class ApprovalCaseRoutes {
           prisonerNumber: c.nomisRecord.prisonerNumber,
           probationPractitioner: c.probationPractitioner,
           submittedByFullName: _.head(c.licences).submittedByFullName,
-          releaseDate,
+          releaseDate: releaseDate ? format(releaseDate, 'dd MMM yyyy') : 'not found',
           urgentApproval,
           approvedBy: _.head(c.licences).approvedBy,
           approvedOn: approvedDate,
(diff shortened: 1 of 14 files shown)


server/routes/approvingLicences/handlers/approvalCases.ts
```diff
@@ -41,7 +41,7 @@ export default class ApprovalCaseRoutes {
           prisonerNumber: c.nomisRecord.prisonerNumber,
           probationPractitioner: c.probationPractitioner,
           submittedByFullName: _.head(c.licences).submittedByFullName,
-          releaseDate,
+          releaseDate: releaseDate ? format(releaseDate, 'dd MMM yyyy') : 'not found',
```
andrewrlee, line 44: selectReleaseDate now returns a date or null which means we can now: `string -> date -> string` rather than
`string -> date -> string -> date -> string`

server/routes/approvingVariations/handlers/varyApproveList.ts
```diff
@@ -22,11 +22,11 @@ export default class VaryApproveListRoutes {
         const licence = _.head(c.licences)
 
         const releaseDate = c.nomisRecord.releaseDate
-          ? format(parse(c.nomisRecord.releaseDate, 'yyyy-MM-dd', new Date()), 'dd MMM yyyy')
+          ? format(parseIsoDate(c.nomisRecord.releaseDate), 'dd MMM yyyy')
```
andrewrlee, line 25: we shouldn't be doing parsing of dates that we've retrieved from backend data sources at all 😱
Once we've removed some caseload logic bloat from the frontend we can look at pulling this down into the data layer

server/routes/approvingLicences/handlers/approvalCases.ts
andrewrlee: this weird `not found` string is existing logic that has been pulled out of `selectReleaseDate`

server/utils/utils.ts
```diff
@@ -124,9 +122,19 @@ const jsonDtTo12HourTime = (dt: string): string => {
   return momentTime.isValid() ? momentTime.format('hh:mm a') : null
 }
 
-const toDate = (date: string) => {
```
andrewrlee, line 127: moved this over to `parseIsoDate` which is a bit more tightly defined

server/utils/utils.ts
```diff
@@ -183,40 +191,19 @@ const selectReleaseDate = (nomisRecord: Prisoner) => {
...
 
 const isReleaseDateBeforeCutOffDate = (cutOffDate: string, releaseDate: string): boolean => {
-  const rDate = parse(releaseDate, 'dd/MM/yyyy', new Date())
-  const cDate = parse(cutOffDate, 'dd/MM/yyyy', new Date())
+  const rDate = parseCvlDate(releaseDate)
+  const cDate = parseCvlDate(cutOffDate)
   return isBefore(rDate, cDate) || isEqual(rDate, cDate)
 }
 
-const isPassedArdOrCrd = (licence: LicenceSummary, prisoner: Prisoner | PrisonApiPrisoner): boolean => {
```
andrewrlee, line 206: this wasn't being used so removed

server/utils/utils.ts
```diff
@@ -124,9 +122,19 @@ const jsonDtTo12HourTime = (dt: string): string => {
...
-  return new Date(`${year}-${month}-${day}`)
+const parseIsoDate = (date: string) => {
+  return date ? parse(date, 'yyyy-MM-dd', new Date()) : null
+}
+
+const parseCvlDate = (date: string) => {
+  return date ? parse(date, 'dd/MM/yyyy', new Date()) : null
+}
+
+const parseCvlDateTime = (date: string, { withSeconds }: { withSeconds: boolean }) => {
```
andrewrlee, line 133: specifically not added a default for `withSeconds` - probably safer to always have to specify and have to think/check about what is safe to use

nishanthmahasamudram (approved):
LGTM

nishanthmahasamudram (approved):
LGTM

PR #867, opened 2024-04-09
Moving default gotenberg port

andrewrlee:
Most new typescript repos have an additional server exposing prometheus metrics, running on port 3002 which means a lot of having to stop gotenberg. This moves to another port to stop clashing


louisenorris (approved):
LGTM

PR #868, opened 2024-04-09
CVSL-1786 moving to use licence as a source of hardstop dates

andrewrlee:


integration_tests/integration/createLicence.cy.ts
```diff
@@ -16,6 +16,7 @@ context('Create a licence', () => {
     cy.task('stubGetActivePolicyConditions')
     cy.task('stubGetBankHolidays', dates)
     cy.task('stubAddAdditionalCondition')
+    cy.task('stubFeComponents')
```
andrewrlee, line 19: I'll add this to all cypress tests as a follow up - it adds loads of errors/noise to the logs if this is not stubbed out

server/data/licenceApiClient.ts
```diff
@@ -89,7 +90,7 @@ export default class LicenceApiClient extends RestClient {
     }
   }
 
-  async createLicence(licence: CreateLicenceRequest, user: User): Promise<LicenceSummary> {
+  async createLicence(licence: CreateLicenceRequest, user: User): Promise<LicenceCreationResponse> {
```
andrewrlee, line 93: This is a cut down representation that only includes the licenceId rather than the full object - which is the only bit we actually use

server/routes/creatingLicences/handlers/caseload.test.ts
```diff
@@ -39,6 +37,8 @@ describe('Route Handlers - Create Licence - Caseload', () => {
             id: 1,
             type: LicenceType.AP,
             status: LicenceStatus.IN_PROGRESS,
+            hardStopDate: addDays(parseIsoDate('2022-10-12'), -2),
```
SebNegahbanMoJ, line 40: Not that it particularly matters, just a query as to using `addDays` with a negative value as opposed to `subDays`? Is there a reason that `subDays` isn't suitable, or just habit? 😁

server/routes/views/CaseloadViewModel.ts
```diff
@@ -31,10 +29,8 @@ export default (
           licence.status !== LicenceStatus.NOT_IN_PILOT &&
           licence.status !== LicenceStatus.OOS_RECALL &&
           licence.status !== LicenceStatus.OOS_BOTUS,
-        showHardStopWarning:
-          config.hardStopEnabled &&
-          releaseDate <= moment(hardStopDates.hardStopWarningDate) &&
-          releaseDate > moment(hardStopDates.hardStopCutoffDate),
+        hardStopDate: format(hardStopDate, 'dd MMM yyyy'),
+        showHardStopWarning: config.hardStopEnabled && hardStopWarningDate <= now && now < hardStopDate,
```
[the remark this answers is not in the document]
SebNegahbanMoJ: Looks right to me!

server/routes/creatingLicences/handlers/caseload.test.ts
andrewrlee: I only found subDays halfway through and didn't go back to fix 😂
Can replace in a follow up?

server/services/caseloadService.test.ts
```diff
@@ -1678,66 +1678,4 @@ describe('Caseload Service', () => {
       expect(CaseloadService.isEligibleEDS(yesterday, tenDaysFromNow, tenDaysFromNow, nineDaysFromNow)).toBe(false)
     })
   })
-
-  describe('getHardStopReferenceDate', () => {
```
SebNegahbanMoJ, line 1682: Is this not still in use for `NOT_STARTED` licences? If so, I think we should keep the specs for it

server/views/pages/create/caseload.njk
```diff
@@ -33,7 +33,7 @@
     {% endif %}
 
     {% if offender.showHardStopWarning %}
-        {% set releaseDate = '<span class="urgent-highlight">' + offender.releaseDate + '</span><span class="urgent-highlight urgent-highlight-message">Submit licence by ' + hardStopCutoffDate +'</span>' %}
+        {% set releaseDate = '<span class="urgent-highlight">' + offender.releaseDate + '</span><span class="urgent-highlight urgent-highlight-message">Submit licence by ' + offender.hardStopDate +'</span>' %}
```
SebNegahbanMoJ, line 36: Is `offender.hardStopDate` formatted as `DD MMM YYYY`? I believe that's what it was before when it was passed through explicitly

server/routes/creatingLicences/handlers/caseload.test.ts
SebNegahbanMoJ: I thought that might be the case but wanted to make sure there wasn't a known issue with `subDays` that I wasn't aware of 😂 
No bother, if you remember to fix it then sure, but I'm pretty certain I've seen `addDays` with negative values elsewhere so don't stress it!

server/views/pages/create/caseload.njk
andrewrlee: yes, in caseloadViewModel: 
```
  hardStopDate: format(hardStopDate, 'dd MMM yyyy'),
```

It's annoying that moment and date fns use different date format syntax!

integration_tests/integration/createLicence.cy.ts
andrewrlee: I've included this now

server/routes/creatingLicences/handlers/caseload.test.ts
andrewrlee: I've moved these over to use subDays now

SebNegahbanMoJ (approved):

PR #877, opened 2024-04-15
Cvsl 1799 hard stop highlight the tab with search results for prison admin

nishanthmahasamudram:


server/views/pages/attentionNeededTab.njk
```diff
@@ -0,0 +1,17 @@
+{% from "pages/attentionNeededTabDescription.njk" import attentionNeededTabDescription %}
```
andrewrlee, line 1: Like that you've extracted these all out. 

Guess these are all `/components` rather than `/pages`?

and you could maybe bring the `attentionNeededTabDescription` in to this file. 

maybe as a follow up?

andrewrlee (approved):
LGTM

PR #878, opened 2024-04-15
CVSL-1795 set TIMED_OUT as a licence type where we can update sentence dates

gurps-singh:
This PR is to add the `TIMED_OUT` licence status as a licence where we can update the sentence dates based on the relevant event from NOMIS.


andrewrlee (approved):
LGTM

PR #882, opened 2024-04-15
CVSL-1730 Making NOT_STARTED licences in hard stop period TIMED_OUT

SebNegahbanMoJ:
I believe `isInHardStopPeriod` already accounts for the logic surrounding earliest releases, by picking the relevant hard stop date based on proximity of ARD to CRD. This should mean that we don't need any additional handling in the frontend for this, and we can simply check the flag.
If this is wrong, please let me know and I can fix it!


server/services/caseloadService.ts
```diff
@@ -219,6 +219,8 @@ export default class CaseloadService {
       } else if (this.isRecall(offender)) {
         // Offender is subject to an active recall - not clickable
         licenceStatus = LicenceStatus.OOS_RECALL
+      } else if (offender.cvlFields.isInHardStopPeriod) {
```
andrewrlee, line 222: This is right assumption but think we need to make sure hardstop flag not enabled?

andrewrlee (approved):
LGTM

PR #890, opened 2024-04-19
CVSL-1818 Fix sort issue

andrewrlee:


server/routes/viewingLicences/handlers/viewCases.test.ts
```diff
@@ -519,172 +541,118 @@ describe('Route handlers - View and print case list', () => {
...
+              prisonerNumber: 'A1234AD',
+              releaseDate: '01 May 2022',
+            }),
+            expect.objectContaining({
+              prisonerNumber: 'A1234AE',
+              releaseDate: '01 Jun 2022',
+            }),
+          ],
+          probationView: false,
+        })
```
andrewrlee, line 595: actual issue was a template problem but whilst looking into this I made the tests more explicit,

server/views/pages/view/cases.njk
```diff
@@ -73,7 +73,7 @@
         {
             attributes: {
                 id: 'com-' + loop.index,
-                "data-sort-value": case.probationPractitioner
+                "data-sort-value": case.probationPractitioner.name
```
andrewrlee, line 76: This was breaking sort on PP as this would return `[Object object]` for everyone

server/views/partials/viewCases.njk
```diff
@@ -51,7 +47,7 @@
                     {
                         text: "Release date",
                         attributes: {
-                            "aria-sort":  "none" if params.tabType === 'attentionNeeded' else ("descending"  if probationView else  "ascending")
+                            "aria-sort":  "none" if params.tabType === 'attentionNeeded' else ("descending"  if params.probationView else  "ascending")
```
andrewrlee, line 50: this was always evaluating to false as it needed passing into macro

server/views/partials/viewCases.njk
```diff
@@ -183,6 +179,6 @@
         {% endif %}
         {% set licences = (licences.push(caseData), licences) %}
     {% endfor %}
-    {{ casesTable({licences: licences, tabType: params.tabType}) }}
+    {{ casesTable({licences: licences, tabType: params.tabType, probationView: false }) }}
```
andrewrlee, line 182: This macro is only used by the hardstop tabs - so always used for the prison view

SebNegahbanMoJ: is `probationView` not another name for the `peopleOnProbation` tab rather than based on who is doing the viewing? Or am I misunderstanding your comment?

andrewrlee: Sorry I meant:
> This macro is only used by the tabs introduced by hardstop - so always used for the "in prison" section

SebNegahbanMoJ (approved):

PR #891, opened 2024-04-19
CVSL-1812 - Add licence create, edit, and submit route protection

SebNegahbanMoJ:


integration_tests/mockApis/licence.ts
```diff
@@ -1545,12 +1545,12 @@ export default {
             prisonId: 'BAI',
             sentenceStartDate: '2017-03-01',
             releaseDate: '2024-07-19',
-            confirmedReleaseDate: '2022-11-20',
+            confirmedReleaseDate: '2022-11-10',
             sentenceExpiryDate: '2028-08-31',
             licenceExpiryDate: '2028-08-31',
-            conditionalReleaseDate: '2022-11-21',
+            conditionalReleaseDate: '2022-11-10',
```
SebNegahbanMoJ, line 1551: Changed these dates as the previous values were making the licence appear as being eligible for an early release. This caused a different confirm-create screen to appear, which in turn meant that the createLicence integration test had to click one fewer buttons to progress through certain pages, which caused a break in the tests.

integration_tests/pages/confirmCreate.ts
```diff
@@ -11,13 +11,13 @@ export default class ConfirmCreatePage extends Page {
   }
 
   selectYes = (): ConfirmCreatePage => {
+    cy.task('stubPostLicence')
```
SebNegahbanMoJ, line 14: Moving this here to avoid future potential headaches.
For normal licence creation, stubbing the post method on `clickContinue` was fine. However, if the licence is an early release, `selectYes` _is_ the continue button. This meant that the licence was submitting when it was expecting to be selecting a 'Yes' radio button, and was failing the test due to the POST endpoint not having been mocked yet.
While this isn't currently used, hopefully it will help prevent any issues in future if tests are added explicitly for the early release warning page.

integration_tests/pages/index.ts
```diff
@@ -25,7 +25,7 @@ export default class IndexPage extends Page {
     cy.task('stubGetOffendersByCrn')
     cy.task('searchPrisonersByNomisIds')
     cy.task('stubGetExistingLicenceForOffenderNoResult')
-    cy.task('stubGetPrisonerDetail')
+    cy.task('stubGetCaseloadItem')
```
SebNegahbanMoJ, line 28: Swapping to use licence-API-enriched prisoner-search endpoint.

server/middleware/hardStopCheckMiddleware.ts
```diff
@@ -1,20 +1,20 @@
...
 
 export default function hardStopCheckMiddleware(userType: UserType): RequestHandler {
   return async (req, res, next) => {
     const { licence } = res.locals
     // Prison should only be able to access initial appointment update pages in hard stop, and probation should only be able to access
     // outside of hard stop
-    if (licence.kind === LicenceKind.VARIATION) {
-      return res.redirect('/access-denied')
-    }
+
```
SebNegahbanMoJ, line 11: Removed from here as this is now a piece of middleware that's shared with other routes, where pages that are `VARIATION`s should be accessible _after_ the hard stop period.
eg:
- Before hard stop, PP can access, CA can't
- During hard stop, PP can't access, CA can
- After hard stop (variation or otherwise), PP can access, CA can't

server/routes/creatingLicences/handlers/checkAnswers.ts
```diff
@@ -18,7 +18,7 @@ export default class CheckAnswersRoutes {
 
   GET = async (req: Request, res: Response): Promise<void> => {
     const { licence, user } = res.locals
-    const backLink = req.session.returnToCase || '/licence/create/caseload'
+    const backLink = req.session?.returnToCase || '/licence/create/caseload'
```
SebNegahbanMoJ, line 21: There may be a better solution to this, but in the new routing specs (`crdLicenceInHardStop.test.ts`) there is no session to the request. The easiest solution I could find was to handle a null session in a couple of places, as it was blowing up the tests otherwise.

server/routes/creatingLicences/handlers/confirmCreate.ts
```diff
@@ -11,35 +10,38 @@ import LicenceKind from '../../../enumeration/LicenceKind'
...
     const [nomisRecord, deliusRecord, bankHolidays] = await Promise.all([
-      this.prisonerService.getPrisonerDetail(nomisId, user),
+      this.licenceService.getPrisonerDetail(nomisId, user),
       this.communityService.getProbationer({ nomsNumber: nomisId }),
       this.ukBankHolidayFeedService.getEnglishAndWelshHolidays(),
     ])
 
+    if (nomisRecord.cvl.isInHardStopPeriod) {
+      return res.redirect('/access-denied')
+    }
```
SebNegahbanMoJ, line 30: Unfortunately the hard stop middleware only works when we have a licence object to check `isInHardStopPeriod` on, which we don't have at the point that the middleware runs. Here in the handler however, we have access to the enriched NOMIS record that allows us to check the field. For this reason this route doesn't apply the middleware, but does still prevent access in the hard stop period.

server/routes/initialAppointment/index.ts
```diff
@@ -32,6 +34,7 @@ export default function Index({ licenceService, conditionService }: Services): R
       routePrefix(path),
       roleCheckMiddleware(['ROLE_LICENCE_CA', 'ROLE_LICENCE_RO']),
       fetchLicence(licenceService),
+      licenceKindCheckMiddleware(LicenceKind.VARIATION),
```
SebNegahbanMoJ, line 37: This was previously covered by `hardStopCheckMiddleware`, but has now been split into its own middleware.

server/routes/creatingLicences/handlers/checkAnswers.ts
andrewrlee: 👍  
I might try and fix this in a follow up

server/routes/creatingLicences/handlers/confirmCreate.ts
andrewrlee: add some logging?

andrewrlee (dismissed):
LGTM

andrewrlee (approved):
LGTM

PR #898, opened 2024-04-22
Cvsl 1822 hard stop apply hardstop changes to pp search results

nishanthmahasamudram:


server/utils/nunjucksSetup.ts
```diff
@@ -340,6 +345,44 @@ export function registerNunjucks(app?: express.Express): Environment {
...
+  njkEnv.addFilter('createOffenderLink', (licence: FoundProbationRecord): string => {
+    const isTimedOutLicence = licence?.licenceStatus === 'TIMED_OUT'
+    const isHardStopLicence = licence?.kind === 'HARD_STOP'
+
+    if (isTimedOutLicence && licence.versionOf) {
+      return `/licence/create/id/${licence.licenceId}/licence-changes-not-approved-in-time`
+    }
+
+    if (
+      (isTimedOutLicence && !isHardStopLicence) ||
```
andrewrlee, line 357: you're never going to get a timed out licence that is a hardstop licence so can remove the `!isHardStopLicence`

server/utils/nunjucksSetup.ts
```diff
@@ -340,6 +345,44 @@ export function registerNunjucks(app?: express.Express): Environment {
...
+
+  njkEnv.addFilter('getlicenceStatusForSearchResults', (licence: FoundProbationRecord): LicenceStatus => {
+    if (licence.isReviewNeeded) {
+      return LicenceStatus.REVIEW_NEEDED
+    }
+    const isHardStopLicence = licence?.kind === 'HARD_STOP'
+    return isHardStopLicence ? LicenceStatus.TIMED_OUT : <LicenceStatus>licence.licenceStatus
+  })
+
+  njkEnv.addFilter('getreleaseDateForSearchResults', (releaseDate: string): string => {
```
andrewrlee, line 382: typo but could we not do this in the controller?

andrewrlee: we apparantly call this `DateShort` (e.g: `jsonDtToDateShort`) so `cvlDateAsDateShort` ?

But better in controller if possible 

(we should be parsing dates as close to receiving them as possible - ideally in the client but definitely not when rendering)

andrewrlee: looking through the code and I can see why you've done it that way, maybe just call it `cvlDateToDateShort` for now

(we need to readdress this at some point!)

nishanthmahasamudram: If we handle `'not found'` logic in the backend then we can use `cvlDateAsDateShort`.
Ya sure, will make it `cvlDateToDateShort` for now.

andrewrlee (approved):
LGTM

PR #906, opened 2024-04-24
CVSL-1835 Removing superfluous rule

andrewrlee:
Tests cover this case and did not cover clause I've removed


louisenorris (approved):
LGTM

PR #914, opened 2024-04-25
CVSL-1847 filter out any timed out licences in the vary view

andrewrlee:


server/routes/varyingLicences/handlers/caseload.test.ts
```diff
@@ -306,15 +306,80 @@ describe('Route Handlers - Vary Licence - Caseload', () => {
       caseloadService.getStaffVaryCaseload.mockResolvedValue([
         {
           licences: [
+            {
+              id: 1,
+              type: LicenceType.AP,
+              status: LicenceStatus.ACTIVE,
+            },
             {
```
andrewrlee, line 314: This test check to retrieve the non-active licence if a variation exist. 
Swapped the order here as it wasn't testing the code - it would have passed without special handling as it falls back to just return the first record otherwise

server/routes/varyingLicences/handlers/caseload.ts
```diff
@@ -48,7 +48,9 @@ export default class CaseloadRoutes {
     const caseloadViewModel = cases
       .map(c => {
         const licence =
-          c.licences.length > 1 ? c.licences.find(l => l.status !== LicenceStatus.ACTIVE) : _.head(c.licences)
+          c.licences.length > 1
+            ? c.licences.filter(l => l.status !== LicenceStatus.TIMED_OUT).find(l => l.status !== LicenceStatus.ACTIVE)
+            : _.head(c.licences)
```
andrewrlee, line 53: We should never show timed out licences on the vary case load

nishanthmahasamudram (approved):
LGTM

PR #917, opened 2024-04-25
CVSL-1842 adjust logic for backlink

gurps-singh:
This PR addresses a bug where when a hard stop licence was created, the `Return to caselist` button did not work. This now redirects to the view caseload on clicking of the button rather than the create caseload when a hard stop licence is created and submitted.


andrewrlee (approved):
LGTM

PR #932, opened 2024-04-26
CVSL-1829 fix missing early release warning on hardstop create screen

andrewrlee:
Moving to use the backend as a source of this flag


nishanthmahasamudram (approved):
LGTM

PR #933, opened 2024-04-26
CVSL-1832 add updatedByFullName to mapping licence to offender

gurps-singh:
This PR is to address a bug where `Last Worked On By` was not populated.


andrewrlee (approved):
LGTM

PR #937, opened 2024-04-29
CVSL-1851 add logic to change backlink button

gurps-singh:
This PR addresses a bug where the `Return to caselist` button redirected to the incorrect tab.


server/views/partials/viewCases.njk
```diff
@@ -90,7 +90,7 @@
     {% for case in cases %}
         {% set caseData = [] %}
         {% if case.probationPractitioner %}
-            {% set comDetailsHtml = '<a class="govuk-link" href="/licence/view/probation-practitioner/staffCode/' + case.probationPractitioner.staffCode + '">' + case.probationPractitioner.name + '</a>' %}
+            {% set comDetailsHtml = '<a class="govuk-link" href="/licence/view/probation-practitioner/staffCode/' + case.probationPractitioner.staffCode + '?activeTab=' + params.activeTab+ '">' + case.probationPractitioner.name + '</a>' %}
```
SebNegahbanMoJ, line 93: This may not be possible with the flow of the code, but what happens if `activeTab` is still set to the default `''` when this link is clicked?

gurps-singh: I don't think that would be possible. Tested this locally by manually changing the URL and it successfully shows the COM details and upon clicking the return to caselist button, you are redirected to the caseload with the default `Releases in next 2 working days` tab.

PR #940, opened 2024-04-29
CVSL-1863 temporary remove incorrect label

andrewrlee:


server/views/pages/view/cases.njk
```diff
@@ -83,7 +83,7 @@
                 id: 'release-date-' + loop.index,
                 "data-sort-value": case.releaseDate | dateToUnix
             },
-            html: '<p>' + case.releaseDateLabel + ': <br>' + case.releaseDate + '</p>' 
+            html: '<p>' + (case.releaseDateLabel + ': <br>' if not probationView)  + case.releaseDate + '</p>'
```
andrewrlee, line 86: This is a bit horrible but temporarily removes a potentially mislead label 

it is only misleading if the dates in CVL don't match the dates in nomis, which can only happen after release. 
But this is a risky assumption to rely on and it shouldn't be too difficult to fix properly after our release

nishanthmahasamudram (approved):

PR #953, opened 2024-05-02
Cvsl 1866 b adding hard stop warning to pp search

nishanthmahasamudram:


server/utils/nunjucksSetup.ts
```diff
@@ -389,6 +390,21 @@ export function registerNunjucks(app?: express.Express): Environment {
...
+    'shouldShowHardStopWarning',
+    (licence: { kind: LicenceKind; hardStopWarningDate: string; hardStopDate: string }): boolean => {
+      const now = startOfDay(new Date())
+      return (
+        config.hardStopEnabled &&
+        licence.kind !== LicenceKind.VARIATION &&
+        !!licence.hardStopWarningDate &&
+        parseCvlDate(licence.hardStopWarningDate) <= now &&
+        !!licence.hardStopDate &&
+        now < parseCvlDate(licence.hardStopDate)
```
SebNegahbanMoJ, line 403: I think I'd probably prefer to move the checks of `hardStopWarningDate` and `hardStopDate` definition to outside of the big return statement, just for tidiness and to reduce the size of that single check
```
if (!licence.hardStopWarningDate || !licence.hardStopDate) { return false }
return ...
```
Happy to have a discussion about it if you disagree!

server/routes/views/CaseloadViewModel.ts
```diff
@@ -30,8 +28,9 @@ export default (caseload: ManagedCase[], search: string) => {
           licence.status !== LicenceStatus.NOT_IN_PILOT &&
           licence.status !== LicenceStatus.OOS_RECALL &&
           licence.status !== LicenceStatus.OOS_BOTUS,
-        hardStopDate: hardStopDate && format(hardStopDate, 'dd MMM yyyy'),
-        showHardStopWarning: config.hardStopEnabled && hardStopWarningDate <= now && now < hardStopDate,
+        hardStopDate: format(hardStopDate, 'dd/MM/yyyy'),
+        hardStopWarningDate: format(hardStopWarningDate, 'dd/MM/yyyy'),
```
SebNegahbanMoJ, line 32: Nish and I had a chat about this earlier, and while we (as a service) are trying to move away from repeatedly formatting dates, as it currently stands this seems to be the neatest solution for formatting the dates correctly on all pages that use the shared `shouldShowHardStopWarning` filter.

It might be worth flagging this to Andy when he's back to make sure that he doesn't think there's a particular reason for us to refactor this, but for now I think it's fine to go.

SebNegahbanMoJ: As a side comment, do we need to be checking `hardStopDate && format(hardStoPDate, 'dd/MM/yyyy')` as per the previous version of this code (and the same for hardStopWarningDate)?
I believe that's a check that I added, but I can't remember if I did it because it actually caused an error if you don't or whether it was just for rendering purposes that are now covered by later functions.

nishanthmahasamudram: Yes, they are for rendering purpose and are covered by `cvlDateToDateShort` filter.

PR #954, opened 2024-05-02
CVSL-1771 add functionality to update PPRD on sentence date update

gurps-singh:
This PR is to process PRRD data changes when sentence dates are updated following an event in the front end.


louisenorris (approved):
LGTM

PR #957, opened 2024-05-03
CVSL-1878 Refactoring initial appointment time validator to compare against LED or TUSED depending on licence type

SebNegahbanMoJ:
I couldn't find anywhere that the `DateTime` validator type is used apart from the initial appointment routes. All integration tests also continue to work without requiring changes.


nishanthmahasamudram (approved):
LGTM

PR #963, opened 2024-05-15
CVSL-1738 add inactivate recall job to FE

gurps-singh:
This PR is to add the job to inactivate active licences and their variations where a PRRD has been entered, meaning the person has been recalled.


andrewrlee (approved):
LGTM

PR #982, opened 2024-06-04
CVSL-1910 extracting out CA caseload

andrewrlee:


jobs/promptLicenceCreationService.ts
```diff
@@ -5,18 +5,18 @@ import config from '../server/config'
...
 import CommunityService from '../server/services/communityService'
 import logger from '../logger'
 import { LicenceApiClient } from '../server/data'
+import CaCaseloadService from '../server/services/caCaseloadService'
 
 export default class PromptLicenceCreationService {
   constructor(
     private readonly licenceService: LicenceService,
-    private readonly caseloadService: CaseloadService,
+    private readonly caseloadService: CaCaseloadService,
```
andrewrlee, line 19: going to remove the dependency here in a follow up, similar to the caseloads it would benefit from something simpler and more specific.

server/services/caCaseloadService.ts
```diff
@@ -0,0 +1,440 @@
...
+    return offender.nomisRecord?.imprisonmentStatus && offender.nomisRecord?.imprisonmentStatus === 'BOTUS'
+  }
+
+  /**
+   * Parole Eligibility Date must be set and in the future
+   * If the date is in the past, it's no longer parole eligible
+   * Parole eligibility excludes the offender, so a truthy return here is an exclusion from CVL
+   * @param ped
+   */
+  public static isParoleEligible(ped: string): boolean {
```
andrewrlee, line 408: all of this is copied across as is and then trimmed back a bit. 
Going to extract out some of these static helpers into somewhere central to share across the caseload builders - these can be reimplemented in the backend

louisenorris (approved):
LGTM

server/services/caseloadService.ts
```diff
@@ -73,47 +72,6 @@ export default class CaseloadService {
       .then(caseload => this.mapResponsibleComsToCases(caseload))
   }
 
-  async getOmuCaseload(user: User, prisonCaseload: string[]): Promise<OmuCaselist> {
```
louisenorris, line 76: Assume this and the `pairNomisRecordsWithDelius` are no longer being used or have already been extracted out to somewhere else? 😅 sorry if I've missed this

louisenorris: Maybe these are included as the static helpers you refer to in your comment above actually 😅

PR #993, opened 2024-06-20
Cvsl 1910 cvl caseload move ca caseload from frontend to backend

nishanthmahasamudram:


server/services/lists/caCaseloadService.ts
```diff
@@ -122,6 +163,37 @@ export default class CaCaseloadService {
...
+    return cases.map(c => {
+      const releaseDate = c.licence?.releaseDate || selectReleaseDate(c.nomisRecord)
+      const tabType = determineCaViewCasesTab(c.licence, c.nomisRecord, c.cvlFields)
+      return {
+        kind: c.licence?.kind,
+        licenceId: c.licence?.id,
+        licenceVersionOf: c.licence?.versionOf,
+        name: convertToTitleCase(`${c.nomisRecord.firstName} ${c.nomisRecord.lastName}`.trim()),
+        prisonerNumber: c.nomisRecord.prisonerNumber,
+        probationPractitioner: c.probationPractitioner,
```
SebNegahbanMoJ, line 184: **question:** Is probationPractitioner always defined? What happens if the case is unallocated?

server/services/lists/caCaseloadService.ts
```diff
@@ -122,6 +163,37 @@ export default class CaCaseloadService {
...
+        releaseDate: releaseDate ? format(releaseDate, 'dd MMM yyyy') : 'not found',
+        releaseDateLabel: c.nomisRecord.confirmedReleaseDate ? 'Confirmed release date' : 'CRD',
+        licenceStatus: c.licence.status,
```
SebNegahbanMoJ, line 187: **nitpick:** licence is treated as nullable elsewhere, but isn't here. I think there should always be a representation of a licence (even if it's `NOT_STARTED`) so non-nullable _should_ work, but we should probably be consistent one way or the other

nishanthmahasamudram: Ya, there won't be a situation with status null because even if we don't have a licence we are making it NOT_STARTED. So nullable isn't required here.

PR #999, opened 2024-06-28
CVSL-1910 Refactoring CA caseload

SebNegahbanMoJ:


server/services/lists/caCaseloadService.ts
```diff
@@ -59,137 +63,188 @@ export type ManagedCase = {
...
+          lastWorkedOnBy: licence.updatedByFullName,
+          isDueForEarlyRelease: licence.isDueForEarlyRelease,
+          isInHardStopPeriod: licence.isInHardStopPeriod,
+          tabType: determineCaViewCasesTab(p.prisoner, p.cvl, licence),
+          probationPractitioner: {
+            staffUsername: licence.comUsername,
+          },
+        }
+      })
+      .filter(c => c)
```
andrewrlee, line 170: **question:** `c` is never going to be falsy is it?

SebNegahbanMoJ: I'm pretty sure you're right and it's left over from a previous iteration
Will confirm and remove it if so!

server/services/lists/caCaseloadService.ts
```diff
@@ -521,4 +425,30 @@ export default class CaCaseloadService {
       return view === 'prison' ? crd1 - crd2 : crd2 - crd1
     })
   }
+
+  splitCasesByComDetails(cases: CaCase[]) {
+    const groupedCases = cases.reduce((acc, c) => {
+      const groups = acc
+      if (!groups.staffCode) {
```
andrewrlee, line 432: **suggestion:** rather than doing this on each iteration would just initialise your `acc`?

`{ staffCode = [],  staffUsername = [] ...etc.. } as GroupedByCom`

also should those be pluralized?

SebNegahbanMoJ: Goddamn it I looked for ages to see if I could set a default/initial value for `acc` and found nothing. I was surprised that I seemingly couldn't, but turns out that's because I can 😅 

To be honest I'm thinking that I don't like the names anyway, as it implies that the array is a list of `staffCode`s, `staffUsername`s etc, whereas it's actually cases _with_ a staff code or _with_ a staff username.
I might just rename them while I'm changing it anyway

server/services/lists/caCaseloadService.ts
```diff
@@ -219,287 +278,132 @@ export default class CaCaseloadService {
...
 
-  private pairDeliusRecordsWithNomis = async (
-    managedOffenders: Array<DeliusRecord>,
-    user: User
-  ): Promise<Array<ManagedCase>> => {
-    const caseloadNomisIds = managedOffenders
-      .filter(offender => offender.otherIds?.nomsNumber)
-      .map(offender => offender.otherIds?.nomsNumber)
+  private async mapCasesToComs(cases: CaCase[]): Promise<CaCase[]> {
+    const splitCases = this.splitCasesByComDetails(cases)
```
andrewrlee, line 319: **suggestion:** I found this a little confusing, but like the idea of it. 

Maybe you call this `cases` and prefix the fields `with`.
e.g:
`cases.withNoComId.map(c => ...`

🤷

SebNegahbanMoJ: Good shout!

server/services/lists/caCaseloadService.ts
```diff
@@ -59,137 +63,186 @@ export type ManagedCase = {
...
-              LicenceStatus.TIMED_OUT,
-              LicenceStatus.NOT_IN_PILOT,
-              LicenceStatus.OOS_RECALL,
-              LicenceStatus.OOS_BOTUS,
-            ].some(status => c.licences.find(l => l.status === status))
-          )
-      })
+    if (!prisonersApproachingRelease.length) {
+      return []
+    }
```
andrewrlee, line 130: **quibble:(non-blocking)** There's a lot of defensive coding around checking whether arrays have lengths..

Here we're just avoiding filtering over an empty array which would then be filtered out anyway.
Generally good to avoid these "edges" ([obligatory Michael Feathers link again](https://www.youtube.com/watch?v=vbLyfh6QdI0)) and use empty arrays as sensible null values that can pass through subsequent functions. 

There obviously are places where you need to handle these, we don't really want to do api calls to retrieve no information but these can generally be implementation details of the clients (and probably are in most if not all places)

server/services/lists/caCaseloadService.ts
```diff
@@ -24,18 +22,24 @@ export type CaCaseLoad = {
   showAttentionNeededTab: boolean
 }
 
+export type GroupedByCom = {
```
andrewrlee, line 25: **quibble:** guess maybe we don't need to export this?

andrewrlee: not sure we need to export all the other types either

andrewrlee (dismissed):
LGTM

server/services/lists/caCaseloadService.ts
SebNegahbanMoJ: Great point, not sure why I originally thought it was necessary to put this one in
If I end up having to make any other changes I'll remove this and do a scan through to remove any others that don't really need to be there

andrewrlee (approved):
LGTM

gurps-singh (approved):

PR #1002, opened 2024-07-03
CVSL-1906 point the approver caseload to use the new api endpoints

gurps-singh:
This PR changes the existing approver caseload functions to point to the new api endpoints we have created for the approver caseload.


server/services/lists/approverCaseloadService.test.ts
```diff
@@ -1,124 +1,56 @@
...
-      isDueToBeReleasedInTheNextTwoWorkingDays: false,
-    } as LicenceSummaryApproverView
+      urgentApproval: false,
+      probationPractitioner: {
+        staffCode: 'X1234',
+        name: 'Joe Bloggs',
+      },
+    } as ApprovalCase
 
     it('builds the approval needed caseload', async () => {
```
gurps-singh, line 43: **thought:** ‏- don't know if these tests now have any value as we are just testing whether the mock is returned rather than any other functionality. Get rid?

integration_tests/integration/approveLicence.cy.ts
```diff
@@ -22,7 +22,7 @@ context('Approve a licence', () => {
     })
 
     cy.task('stubGetCompletedLicence', { statusCode: 'SUBMITTED', typeCode: 'AP_PSS' })
-    cy.task('stubGetLicencesToApprove')
```
gurps-singh, line 25: **question:** ‏Now that we have the caseload on the backend, I don't believe the get functions for licences to approve or recently approved are needed anymore (added a TODO where they are used)?

server/data/licenceApiClient.ts
```diff
@@ -251,6 +252,7 @@ export default class LicenceApiClient extends RestClient {
     )) as ComReviewCount
   }
 
+  // TODO: don't believe these two are used now we have moved the caseload to the backend - remove these from where they are used and the tests?
```
gurps-singh, line 255: **note:** ‏Something to think about

server/services/lists/approverCaseloadService.test.ts
andrewrlee: **suggestion(non-blocking):**  that would make sense - think we should remove these as part of the overall story

andrewrlee (approved):
LGTM

PR #1005, opened 2024-07-09
CVSL-1911: Replace Container type with Array in COM case service. Rem…

justin-gardiner:
…ove unnecessary mocks in some tests.


andrewrlee (approved):
LGTM

PR #1006, opened 2024-07-09
CVSL-1906 remove old endpoints and tests

gurps-singh:
This PR is to remove the `getLicencesForApproval` and `getLicencesRecentlyApproved` functions from the frontend following the adding of the new approver caseload functionality in the backend.


server/@types/licenceApiClientTypes.ts
```diff
@@ -20,7 +20,6 @@ export type Licence =
   | components['schemas']['VariationLicence']
   | components['schemas']['HardStopLicence']
   | components['schemas']['CrdLicence']
-export type LicenceSummaryApproverView = components['schemas']['LicenceSummaryApproverView']
```
gurps-singh, line 23: **note:** ‏Listing noted that this is no longer used - double checked and no longer in use so removed.

SebNegahbanMoJ (approved):

PR #1012, opened 2024-07-24
CVSL-2050 Fixing 500 error when searching when case had an undefined COM name

SebNegahbanMoJ:


justin-gardiner (approved):
LGTM

SebNegahbanMoJ:
This is being fixed in a matter of days by moving the caCaseloadService to the API

PR #1013, opened 2024-07-25
CVSL-1910-date-format-fix

nishanthmahasamudram:


server/routes/viewingLicences/handlers/viewCases.ts
```diff
@@ -63,7 +63,7 @@ export default class ViewAndPrintCaseRoutes {
         }
       }),
       CaViewCasesTab,
-      showAttentionNeededTab,
+      showAttentionNeededTab: cases.some(c => c.tabType === 'ATTENTION_NEEDED'),
```
gurps-singh, line 66: **question:** ‏Probably lack of TS/JS understanding but will some return only some of them and do we not need all of them?

nishanthmahasamudram: Some in JS will return true if any of the element matches the condition.
It's a boolean used to show and hide attention needed tab.

gurps-singh (approved):

PR #1016, opened 2024-07-26
CVSL-1911: Refactor comCaseloadService so that it returns a much simp…

justin-gardiner:
…ler return type with only required data. Separate out any view related properties into the view handlers.


server/@types/managedCase.ts
```diff
@@ -30,23 +38,8 @@ export type Licence = {
...
-  dateCreated?: string
-  approvedBy?: string
-  approvedDate?: string
-  versionOf?: number
-  updatedByFullName?: string
-  submittedByFullName: string
-  isDueForEarlyRelease: boolean
-  releaseDate: Date
-  isDueToBeReleasedInTheNextTwoWorkingDays: boolean
+  previouslyApprovedId?: number
```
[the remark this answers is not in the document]
justin-gardiner: yes it's no longer needed, I will remove

server/routes/creatingLicences/handlers/caseload.ts
```diff
@@ -35,17 +38,60 @@ export default class CaseloadRoutes {
...
+          comCase.prisonerNumber
+        ),
+        releaseDate: format(parseCvlDate(comCase.releaseDate), 'dd MMM yyyy'),
+        hardStopDate: comCase.hardStopDate && format(parseCvlDate(comCase.hardStopDate), 'dd/MM/yyyy'),
+        hardStopWarningDate:
+          comCase.hardStopWarningDate && format(parseCvlDate(comCase.hardStopWarningDate), 'dd/MM/yyyy'),
+        isClickable:
+          comCase.probationPractitioner !== undefined &&
+          comCase.licenceStatus !== LicenceStatus.NOT_IN_PILOT &&
+          comCase.licenceStatus !== LicenceStatus.OOS_RECALL &&
```
SebNegahbanMoJ, line 60: **nitpick:** No harm in the `OOS`s being here but I believe (and I may be getting confused as we've had various conversations about it) that we're removing `OOS_RECALL` and `OOS_BOTUS` as statuses, as we never actually use them for anything

justin-gardiner: ok will remove

server/routes/support/handlers/probationStaff.ts
```diff
@@ -31,7 +30,7 @@ export default class ProbationTeamRoutes {
         : await this.comCaseloadService.getStaffVaryCaseload({ ...user, deliusStaffIdentifier: staff.staffIdentifier })
 
     return res.render('pages/support/probationStaff', {
-      caseload: createCaseloadViewModel(caseload),
+      caseload,
```
SebNegahbanMoJ, line 33: **question:** Do we not need to add the additional fields added in `handlers/caseload.ts` to the caseload object here to match the COM caseload view?

server/services/lists/comCaseloadService.ts
```diff
@@ -348,4 +376,102 @@ export default class ComCaseloadService {
       'licence status is not one of ACTIVE, VARIATION_IN_PROGRESS, VARIATION_SUBMITTED, VARIATION_APPROVED, VARIATION_REJECTED, REVIEW_NEEDED'
     )
   }
+
+  createCaseloadViewModel = (caseload: ManagedCase[]): ComCase[] => {
+    return caseload
+      .map(c => {
+        const licence = this.findLicenceToDisplay(c)
+        const releaseDate = c.nomisRecord.releaseDate || c.nomisRecord.conditionalReleaseDate
```
SebNegahbanMoJ, line 384: **suggestion:** I think here we should be using the licence release date if it exists, then falling back to NOMIS if it doesn't (eg if the licence is `NOT_STARTED`

server/services/lists/comCaseloadService.ts
```diff
@@ -348,4 +376,102 @@ export default class ComCaseloadService {
...
+
+  createVaryCaseloadViewModel = (caseload: ManagedCase[]): ComCase[] => {
+    return caseload.map(managedCase => {
+      const licences = managedCase.licences.filter(l => l.status !== LicenceStatus.TIMED_OUT)
+      const licence =
+        licences.length > 1
+          ? licences.find(l => l.status !== LicenceStatus.ACTIVE && l.status !== LicenceStatus.REVIEW_NEEDED)
+          : _.head(licences)
+
+      const releaseDate = managedCase.nomisRecord.releaseDate || managedCase.nomisRecord.conditionalReleaseDate
```
SebNegahbanMoJ, line 417: **suggestion:** As above, I think this should come directly from the licence and probably *shouldn't* fall back to NOMIS (though happy to discuss this)
This could mean that the releaseDate is null, but would also highlight that dates are missing in CVL, which we would need to fix!
Thinking on it further, IS91 and remand cases (though remand are currently OOS) would have no dates in NOMIS but would (hopefully) have dates in CVL, so we'd definitely want to be at least prioritising the licence, if not only reading from it.

server/routes/creatingLicences/handlers/caseload.test.ts
```diff
@@ -22,130 +18,84 @@ describe('Route Handlers - Create Licence - Caseload', () => {
...
+        name: 'John Roberts',
+        releaseDate: '12/10/2022',
+        sortDate: '2022-10-12',
+        prisonerNumber: '123',
+        licenceId: 1,
+        licenceType: LicenceType.AP,
+        licenceStatus: LicenceStatus.IN_PROGRESS,
+        hardStopDate: '10/10/2022',
+        hardStopWarningDate: '10/10/2022',
+        kind: LicenceKind.CRD,
```
SebNegahbanMoJ, line 31: **praise:** I love this return structure, so much cleaner!

server/routes/support/handlers/probationStaff.ts
justin-gardiner: no, we don't use any of those fields on this view. This view is quite limited as it displays both vary and create cases and only uses fields common to both of them

server/services/lists/comCaseloadService.ts
justin-gardiner: done, updated to use licence.releaseDate if available

server/services/lists/comCaseloadService.ts
```diff
@@ -185,8 +215,11 @@ export default class ComCaseloadService {
               id: licence.licenceId,
               status: licence.isReviewNeeded ? LicenceStatus.REVIEW_NEEDED : <LicenceStatus>licence.licenceStatus,
               type: <LicenceType>licence.licenceType,
-              comUsername: licence.comUsername,
               kind: <LicenceKind>licence.kind,
+              crn: licence.crn,
+              nomisId: licence.nomisId,
+              name: convertToTitleCase(`${offender.nomisRecord.firstName} ${offender.nomisRecord.lastName}`.trim()),
```
SebNegahbanMoJ, line 221: **nitpick (non-blocking):** I think we can also take name from the licence, but don't worry about it if you don't need to make any other code changes

justin-gardiner: updated

SebNegahbanMoJ (approved):

PR #1046, opened 2024-08-08
Cvsl 1911 fix sorting and review needed issues

justin-gardiner:


server/routes/support/handlers/probationTeam.ts
```diff
@@ -21,16 +22,22 @@ export default class ProbationTeamRoutes {
...
-          comCase.hardStopWarningDate && format(parseCvlDate(comCase.hardStopWarningDate), 'dd/MM/yyyy'),
-      }
-    })
+    )
+      .map(comCase => {
+        const releaseDate = comCase.releaseDate ? format(parseCvlDate(comCase.releaseDate), 'dd MMM yyyy') : 'not found'
+        return {
+          ...comCase,
+          releaseDate,
+          licenceStatus: comCase.isReviewNeeded ? LicenceStatus.REVIEW_NEEDED : comCase.licenceStatus,
```
SebNegahbanMoJ, line 31: Is this not also needed for the `probationStaff` vary caseload?

server/routes/support/handlers/probationTeam.ts
```diff
@@ -21,16 +22,22 @@ export default class ProbationTeamRoutes {
...
+          sortDate: comCase.releaseDate && parseCvlDate(comCase.releaseDate),
+          hardStopDate: comCase.hardStopDate && format(parseCvlDate(comCase.hardStopDate), 'dd/MM/yyyy'),
+          hardStopWarningDate:
+            comCase.hardStopWarningDate && format(parseCvlDate(comCase.hardStopWarningDate), 'dd/MM/yyyy'),
+        }
+      })
+      .sort((a, b) => {
+        return (a.sortDate?.getTime() || 0) - (b.sortDate?.getTime() || 0)
```
SebNegahbanMoJ, line 39: **thought (non-blocking):** I don't think it matters with the number of cases we're dealing with, but I wonder if there's an efficiency difference between creating a `releaseDateString` that is the formatted `releaseDate`, then comparing `releaseDate` directly here.

server/routes/support/handlers/probationTeam.ts
justin-gardiner: yes, it should be, added

server/routes/support/handlers/probationTeam.ts
justin-gardiner: quite possibly, I guess it would be worth testing if it started to get slow.

SebNegahbanMoJ (approved):

PR #1047, opened 2024-08-09
CVSL-2056 Fixing 500 error during policy upgrade when two deleted conditions existed on a licence

SebNegahbanMoJ:
Should only be deployed after the corresponding [API changes](https://github.com/ministryofjustice/create-and-vary-a-licence-api/pull/575)

### DIFF
--- a/server/routes/varyingLicences/handlers/policyChange.ts
+++ b/server/routes/varyingLicences/handlers/policyChange.ts
@@ -1,8 +1,8 @@
 import { Response, Request } from 'express'
 import LicenceService from '../../../services/licenceService'
 import LicenceType from '../../../enumeration/licenceType'
-import { AdditionalCondition } from '../../../@types/licenceApiClientTypes'
-import ConditionService from '../../../services/conditionService'
+import { AddAdditionalConditionRequest, AdditionalCondition } from '../../../@types/licenceApiClientTypes'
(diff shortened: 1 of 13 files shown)


server/routes/varyingLicences/handlers/policyChange.ts
```diff
@@ -99,62 +99,89 @@ export default class PolicyChangeRoutes {
...
+      this.licenceService.deleteAdditionalConditionsByCode(conditionsToDelete, licence.id, user)
 
       // Add replacement conditions
       if (req.body.additionalConditions?.length > 0) {
-        const replacementArray = await Promise.all(
-          req.body.additionalConditions.map((code: string) => {
-            return this.conditionService.getAdditionalConditionByCode(code, licence.version)
+        await Promise.all(
+          await req.body.additionalConditions.map(async (code: string) => {
+            const conditionToAdd = await this.conditionService.getAdditionalConditionByCode(code, licence.version)
```
nishanthmahasamudram, line 122: **question:** Is it ok to use nested async/await?

SebNegahbanMoJ: I believe with the use of `map` to return the `Promise` from each iteration, then resolving the returned array of Promises with `Promise.all()`, it works!

nishanthmahasamudram: ok.
Thank you.

PR #1050, opened 2024-08-19
CVSL-2084 Preventing policy upgrade notice page from appearing when there are zero changes to review

SebNegahbanMoJ:
Fixing a small bug that's been around for a while.
When a user has already reviewed all policy changes, they still meet the criteria of being a `VARIATION_IN_PROGRESS` with a parent on an older version. If the user doesn't submit the variation immediately, every time they go to edit the variation, they are presented with a screen telling them that there are "Zero policy changes to review"
This screen is redundant for these users, so should just be skipped.


andrewrlee (approved):
LGTM

PR #1057, opened 2024-08-28
CVSL-2102 library updates

andrewrlee:


.circleci/config.yml
```diff
@@ -237,11 +237,6 @@ workflows:
               only:
                 - main
     jobs:
-      - hmpps/npm_outdated:
```
andrewrlee, line 240: removed this as it adds to alert fatigue and is not strictly necessary for repos that aren't the template project

server/config/policyChangeHintText.ts
```diff
@@ -1,4 +1,6 @@
-export default {
+type HintText = { description: string[]; bulletpoints: string[] }
```
andrewrlee, line 1: all subsequent changes are just to make new version of typescript happy

nishanthmahasamudram (approved):
LGTM

PR #1058, opened 2024-08-29
CVSL-2107 Tidying up vary approval caseload

andrewrlee:


server/services/lists/caseloadService.ts
```diff
@@ -1,128 +1,144 @@
...
+        probationPractitioner,
       }
     })
   }
 
-  private async mapResponsibleComsToCases(caseload: ManagedCase[]): Promise<ManagedCase[]> {
-    const comUsernames = caseload
-      .map(
-        offender =>
-          offender.licences.find(l => offender.licences.length === 1 || l.status !== LicenceStatus.ACTIVE).comUsername
```
andrewrlee, line 88: The queries are always limited to return VARIATION_SUBMITTED so no need to check status.

There should only ever be one of these per noms id so I've created an assertion that ensures there's no duplicates.

gurps-singh (approved):

PR #1059, opened 2024-08-30
CVSL-2108 Filtering out inactive teams

andrewrlee:
This is based on the logic in the community api which triggers the 404 response we receive when trying to request one of these teams [here](https://github.com/ministryofjustice/community-api/blob/5a30ec4bab1fc648cf14201972dbe3ad9ca583df/src/main/java/uk/gov/justice/digital/delius/jpa/standard/repository/TeamRepository.java#L15).

### DIFF
--- a/server/middleware/populateCurrentUser.ts
+++ b/server/middleware/populateCurrentUser.ts
@@ -1,6 +1,6 @@
 import { RequestHandler } from 'express'
 import logger from '../../logger'
-import { convertToTitleCase, removeDuplicates } from '../utils/utils'
+import { convertToTitleCase, parseIsoDate, removeDuplicates } from '../utils/utils'
 import CvlUserDetails from '../@types/CvlUserDetails'
 import config from '../config'
 import LicenceService from '../services/licenceService'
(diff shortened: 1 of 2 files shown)


server/middleware/populateCurrentUser.ts
```diff
@@ -72,10 +72,11 @@ export default function populateCurrentUser(userService: UserService, licenceSer
             cvlUser.telephoneNumber = probationUser?.telephoneNumber
             cvlUser.probationAreaCode = probationUser?.probationArea?.code
             cvlUser.probationAreaDescription = probationUser?.probationArea?.description
-            cvlUser.probationPduCodes = probationUser?.teams?.map(team => team?.borough?.code)
-            cvlUser.probationLauCodes = probationUser?.teams?.map(team => team?.district?.code)
-            cvlUser.probationTeamCodes = probationUser?.teams?.map(team => team?.code)
-            cvlUser.probationTeams = probationUser?.teams
+            const teams = probationUser?.teams?.filter(t => !t.endDate || parseIsoDate(t.endDate) > new Date()) || []
+            cvlUser.probationPduCodes = Array.from(new Set(teams.map(team => team?.borough?.code)))
+            cvlUser.probationLauCodes = Array.from(new Set(teams.map(team => team?.district?.code)))
```
andrewrlee, line 77: prevent duplicates appearing in the list when someone exists on multiple teams in the same LAU/PDU

server/middleware/populateCurrentUser.ts
```diff
@@ -72,10 +72,11 @@ export default function populateCurrentUser(userService: UserService, licenceSer
             cvlUser.telephoneNumber = probationUser?.telephoneNumber
             cvlUser.probationAreaCode = probationUser?.probationArea?.code
             cvlUser.probationAreaDescription = probationUser?.probationArea?.description
-            cvlUser.probationPduCodes = probationUser?.teams?.map(team => team?.borough?.code)
-            cvlUser.probationLauCodes = probationUser?.teams?.map(team => team?.district?.code)
-            cvlUser.probationTeamCodes = probationUser?.teams?.map(team => team?.code)
-            cvlUser.probationTeams = probationUser?.teams
+            const teams = probationUser?.teams?.filter(t => !t.endDate || parseIsoDate(t.endDate) > new Date()) || []
```
andrewrlee, line 75: this will also "fix" the issue when someone has no probation teams - rather than receiving an unexpected error they will see no cases.. we might want to add some content here.

PR #1065, opened 2024-09-11
CVSL-1788 Adding and updating validators for policy v3

SebNegahbanMoJ:


server/routes/manageConditions/types/additionalConditionInputs/cashInPossession.ts
```diff
@@ -0,0 +1,11 @@
+import { Expose } from 'class-transformer'
+import { IsNotEmpty, IsNumberString } from 'class-validator'
+
+class CashInPossession {
+  @Expose()
+  @IsNotEmpty({ message: 'Enter a value in £' })
```
andrewrlee, line 6: presumably we got content sign off for these?

SebNegahbanMoJ: They're just placeholders for now, I'll be working with Tom on the content for both the error messages and the policy upgrade journey next. As it's not going to be visible yet, I thought it was alright to get them out as-is, but can hold off if you'd rather

andrewrlee (approved):
LGTM

PR #1066, opened 2024-09-12
CVSL-1788 Adding routing for single instance file upload condition

SebNegahbanMoJ:


server/routes/manageConditions/handlers/fileUploads/index.ts
```diff
@@ -52,12 +52,19 @@ export default function Index({ licenceService, conditionService }: Services): R
   }
 
   {
-    // upload/delete files from input page
-    const controller = new FileUploadInputRoutes(licenceService)
+    // upload/delete files from input page for multi-instance conditions (eg MEZ)
+    const controller = new FileUploadInputRoutes(licenceService, true)
```
andrewrlee, line 56: **nit:(non-blocking)** passing boolean flags is a bit of a code smell.. enums or objects would be more descriptive

PR #1067, opened 2024-09-16
CVSL-2154 Adding integration tests for MEZ 

SebNegahbanMoJ:
Also fixes routing for map deletion during licence creation flow


andrewrlee (approved):
LGTM

PR #1068, opened 2024-09-19
CVSL-2151 Adding correct routing for map upload conditions to redirect back to policy review flow when appropriate

SebNegahbanMoJ:
Also fixes some 'fromReview' routing for MEZ

### DIFF
--- a/server/enumeration/fileUploadType.ts
+++ b/server/enumeration/fileUploadType.ts
@@ -1,6 +1,6 @@
 enum FileUploadType {
-  multiInstance = 'multiInstance',
-  singleInstance = 'singleInstance',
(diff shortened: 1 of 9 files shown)


server/enumeration/fileUploadType.ts
```diff
@@ -0,0 +1,6 @@
+enum FileUploadType {
+  multiInstance = 'multiInstance',
+  singleInstance = 'singleInstance',
```
andrewrlee, line 3: **suggestion:** enums instances are generally UPPER_SNAKE case

andrewrlee (approved):
LGTM

PR #1069, opened 2024-09-19
PI-2534 Migrate from Community API to Probation Integration API

marcus-bcl:
Community API is now deprecated. This PR migrates the CVL UI code to use the [Create and Vary a Licence and Delius integration service](https://ministryofjustice.github.io/hmpps-probation-integration-services/tech-docs/projects/create-and-vary-a-licence-and-delius/index.html) instead.

The data returned by the new API should be the same as before, but please reach out to us on Slack at [#probation-integration-tech](https://moj.enterprise.slack.com/archives/C02HQ4M2YQN) if you hit any problems or need any changes.


server/data/deliusClient.ts
```diff
@@ -0,0 +1,60 @@
...
+  async getManagerEmailAddresses(crns: string[]): Promise<DeliusStaffEmail[]> {
+    return (await this.post({ path: `/probation-case/responsible-community-manager`, data: crns })) as Promise<
+      DeliusStaffEmail[]
+    >
+  }
+}
+
+export function nameToString(name: components['schemas']['Name']): string {
+  return name
+    ? [name.forename, name.middleName, name.surname]
```
andrewrlee, line 55: I think we didn't previously explicitly include the middle name too? 

is this equivalent to `${forenames} ${surname}`?

andrewrlee (approved):
LGTM

PR #1081, opened 2024-10-08
CVSL-2151 Adding policy upgrade jump page hint text

SebNegahbanMoJ:
Not to be merged before policy v3 goes live, as this will affect the existing policy upgrade journey


nishanthmahasamudram (approved):
LGTM

PR #1082, opened 2024-10-10
CVSL-2181 Changing content and layout of MEZ list screen for maps without names (as will be in policy v3)

SebNegahbanMoJ:
Comparison of new and old styling can be seen here:
image

This design has been signed off by UCD


andrewrlee (approved):
LGTM

PR #1083, opened 2024-10-16
CVSL-2193 Hijacking policy review flow to redirect to MEZ input page in order to remove map names

SebNegahbanMoJ:


server/routes/manageConditions/handlers/fileUploads/fileUploadInputRoutes.ts
```diff
@@ -22,6 +24,16 @@ export default class fileUploadInputRoutes {
     }
 
     if (req.query?.fromPolicyReview) {
+      // This hijacks the policy review loop to allow users to review each instance of a changed multi-instance upload condition.
+      if (this.fileUploadType === FileUploadType.MULTI_INSTANCE) {
+        const activePoliceVersion = await this.conditionService.getPolicyVersion()
```
andrewrlee, line 29: would this differ from licence.version? thought we updated that first?

(wondering if you can avoid an extra api call)

SebNegahbanMoJ: No, you're right, it wouldn't
I'll swap that over now, good spot!

andrewrlee (approved):
LGTM

PR #1085, opened 2024-10-17
CVSL-2192 Adding missing policy upgrade hint text for outOfBoundsPremises condition

SebNegahbanMoJ:
Changes were made to the outOfBoundsPremises condition in v2.1 of the policy some time ago, which results in it being recognised as `NEW_OPTIONS` change when going through a policy upgrade.
While this hasn't been raised so far, with the new version of the policy anyone upgrading from v1.0 to v2.0 to v3 would experience the issue of missing content.


nishanthmahasamudram (approved):
LGTM

PR #1095, opened 2024-10-30
CVSL-2233 Updating banner content for after policy v3 release

SebNegahbanMoJ:
Note: The banner is not currently turned on, just updating the content for when we flip the switch


andrewrlee (approved):
LGTM

PR #1096, opened 2024-10-30
CVSL-2145 changes to HDC AP licence template

louisenorris:


server/utils/utils.ts
```diff
@@ -109,7 +109,7 @@ const stringToAddressObject = (address: string): Address => {
 
 const jsonDtToDate = (dt: string): string => {
   const momentDate = moment(dt, 'DD/MM/YYYY HH:mm')
-  return momentDate.isValid() ? momentDate.format('Do MMMM YYYY') : null
+  return momentDate.isValid() ? momentDate.format('D MMMM YYYY') : null
```
louisenorris, line 112: Design confirmed we should be displaying cardinal dates (eg 1 December 2024) rather than ordinal dates (eg 1st December 2024) across the service so altered the date formatting here.

server/utils/utils.ts
```diff
@@ -119,7 +119,7 @@ const jsonDtToDateShort = (dt: string): string => {
 
 const jsonDtToDateWithDay = (dt: string): string => {
   const momentDate = moment(dt, 'DD/MM/YYYY HH:mm')
-  return momentDate.isValid() ? momentDate.format('dddd Do MMMM YYYY') : null
+  return momentDate.isValid() ? momentDate.format('dddd D MMMM YYYY') : null
```
louisenorris, line 122: Design confirmed we should be displaying cardinal dates (eg 1 December 2024) rather than ordinal dates (eg 1st December 2024) across the service so altered the date formatting again here.

server/views/pages/licence/HDC_AP.njk
```diff
@@ -61,7 +55,25 @@
...
+               <span data-qa="appointment-person" class="bold line-space-after">{{ appointmentPerson }}</span>
+               {% for a in (licence.appointmentAddress | formatAddressAsList) %}
+                  <span class="bold line-space-after">{{ a }}</span>
+               {% endfor %}
+         </p>
+         <p data-qa="appointment-time">On <span class="bold">{{ licence.appointmentTime | datetimeToDateWithDay }}</span> at <span class="bold">{{ licence.appointmentTime | datetimeTo12HourTime }}</span></p>
+         <p>Contact telephone number: <span class="bold">{{ licence.appointmentContact }}</span></p>
+      </div>
+   </div>
+
```
louisenorris, line 76: Policy don’t want us to use options like “immediately after release” and "By 2pm next working day" that we use in CVL in AFER so removed this - think we will therefore need to adjust CVL to only show these options if HDC kind

SebNegahbanMoJ (approved):

server/views/pages/licence/HDC_AP.njk
SebNegahbanMoJ: If there's not already a ticket for this, we should probably add one!
Wouldn't want it to slip through the cracks

louisenorris: I've added a comment re this to the Allow creation of HDC cases in CVL ticket

PR #1112, opened 2024-11-21
CVSL-2319-Migrating to shared eslint library

nishanthmahasamudram:


eslint.config.mjs
```diff
@@ -0,0 +1,38 @@
+import hmppsConfig from '@ministryofjustice/eslint-config-hmpps'
+
+export default [
+  ...hmppsConfig({
+    extraPathsAllowingDevDependencies: ['**/__testutils/**'],
+  }),
+  {
+    rules: {
```
nishanthmahasamudram, line 8: There are few rules which we need to fix going forward.

andrewrlee (approved):
I think we should probably remove assets from the rules and look at cutting back some of those disabled rules but this is good for now

LGTM

PR #1115, opened 2024-11-22
CVSL-2241 add dpr component library and dependencies

louisenorris:


server/views/layout.njk
```diff
@@ -121,8 +121,21 @@
 
 {% block bodyEnd %}
   {# Run JavaScript at end of the <body>, to avoid blocking the initial render. #}
-  <script type="module" src="/assets/govukFrontendInit.js"></script>
+  <script type="module" src="/assets/ext/day.js"></script>
+  <script type="module" src="/assets/ext/dayjs/plugin/customParseFormat.js"></script>
+  <script type="module" src="/assets/ext/chart.js"></script>
+  <script type="module" src="/assets/ext/chartjs-datalabels.js"></script>
```
andrewrlee, line 127: Presumably we don't need to load all this on every page across the app? 
Is this just required on pages that show dashboards?

If this is in their documentation can you get them to update it?

louisenorris: The guidance specifies to add the scripts to the nunjucks layout but I'll double check this with them

assets/sass/application.sass
```diff
@@ -5,6 +5,7 @@ $path: "/assets/images/"
 
 @import 'govuk/all'
 @import 'moj/all'
+@import "node_modules/@ministryofjustice/hmpps-digital-prison-reporting-frontend/dpr/all"
```
andrewrlee, line 8: is it possible to add a load-path to the `bin/build-css.sh` so we don't need to reference the file structure here?

server/middleware/setUpStaticResources.ts
```diff
@@ -37,6 +37,30 @@ export default function setUpStaticResources(): Router {
...
+  ;['/node_modules/dayjs/dayjs.min.js'].forEach(dir => {
+    router.use('/assets/ext/day.js', express.static(path.join(process.cwd(), dir), cacheControl))
+  })
+  ;['/node_modules/dayjs/plugin/customParseFormat.js'].forEach(dir => {
+    router.use(
+      '/assets/ext/dayjs/plugin/customParseFormat.js',
+      express.static(path.join(process.cwd(), dir), cacheControl),
+    )
+  })
+  ;['/node_modules/govuk-frontend/dist/govuk/govuk-frontend.min.js'].forEach(dir => {
```
andrewrlee, line 61: **issue:** we already have an entry for govuk-frontend above.

**suggestion:** We don't need to iterate over arrays of one item. 
It'd be better to iterate over all items or if that's not possible then just inline each call

louisenorris: Cool, have updated to no longer iterate where not needed and removed duplication of entry for govuk-frontend - sorry thats me thinking the second was different, introduced via the DPR guidance.

server/views/layout.njk
```diff
@@ -121,8 +121,13 @@
 
 {% block bodyEnd %}
   {# Run JavaScript at end of the <body>, to avoid blocking the initial render. #}
+  <!-- Govuk Lib -->
   <script type="module" src="/assets/govukFrontendInit.js"></script>
+  <!-- MoJ Lib -->
   <script src="/assets/moj/all.js"></script>
-  <script src="/assets/mojFrontendInit.js"></script>
+  <script type="module" src="/assets/mojInit.mjs"></script>
```
andrewrlee, line 128: mmh.. what's the difference between moj frontend and moj all?

server/views/layout.njk
```diff
@@ -121,8 +121,13 @@
...
+
+  <!-- DPR Lib -->
+  <script type="module" src="/assets/dpr/js/all.mjs"></script>
```
andrewrlee, line 131: I guess this only needs to included in pages that have reports too..

louisenorris: Removed this from layout - thank you

andrewrlee (approved):
LGTM

PR #1117, opened 2024-11-28
CVSL-2328 Adding lightweight creation endpoint for HDC licence in CVL for MVP1

SebNegahbanMoJ:
Current access controls are solely based around having an HDCAD.
Once the HDC API work has been done to add a flag indicating those eligible for CVL licence creation, these rules may need to be expanded.

**Known "issue":** The 'Release date' in the prisoner details header shows `ARD || CRD`, which will be incorrect for HDC cases. This should be addressed as a part of the licence start date work.


andrewrlee (approved):
LGTM

PR #1121, opened 2024-12-05
CVSL-2329 Tweaking reporting instruction journey for HDC licences and adding integration test

SebNegahbanMoJ:


server/views/pages/create/initialMeetingTime.njk
```diff
@@ -28,23 +28,28 @@
...
-                    {% if licence.isEligibleForEarlyRelease %}
-                        {{ govukWarningText({
-                            html: "This person is eligible for an earlier release. It may affect their initial appointment. Find more information <br> in the <a class='govuk-link' rel='noreferrer noopener' target='_blank' href="+ fridayReleasePolicy +">Friday or pre-bank holiday release policy framework (opens in new tab).</a>",
-                            iconFallbackText: "Warning"
-                        }) }}
-                    {% endif %}
-
-                    {% set dateTimeHtml %}
+                }) %}
+                    {% if licence.kind == "HDC"%}
```
andrewrlee, line 32: **suggestion:** would be tempted to move these to macros? 
```
{% if licence.kind == "HDC"%}
   ... call hdc macro
{% else % }
   ... call non hdc macro
```

server/views/pages/create/initialMeetingContact.njk
```diff
@@ -18,7 +18,7 @@
         <input type="hidden" name="_csrf" value="{{ csrfToken }}">
         <div class="govuk-grid-row">
             <div class="govuk-grid-column-full">
-                {% if licence.isEligibleForEarlyRelease %}
+                {% if licence.isEligibleForEarlyRelease and not licence.kind == "HDC" %}
```
andrewrlee, line 21: it'd be nice to get this check pushed into the API at somepoint

andrewrlee (dismissed):
LGTM

server/views/pages/create/initialMeetingContact.njk
andrewrlee: (I saw the other use of `isEligibleForEarlyRelease` and it hadn't got the extra checks - but I assume that's ok as it's in the `else` so not HDC.., things like that are why it's good to to centralise logic like this)

SebNegahbanMoJ: Absolutely, my intention is to make `isEligibleForEarlyRelease` just return `false` for HDC licences, but I need to make sure there's no other places that the flag is used and that that's the behaviour UCD want before I commit to it!

andrewrlee (approved):
LGTM

PR #1127, opened 2024-12-18
Cvsl 2339 highlight hdc cases on approver caselist

louisenorris:


server/views/pages/approve/cases.njk
```diff
@@ -17,6 +17,9 @@
     {% elseif case.isDueForEarlyRelease and approvalNeededView %}
         {% set releaseDate = '<span class="urgent-highlight">' + case.releaseDate
             + '</span><span class="urgent-highlight urgent-highlight-message">Early release</span>' %}
+    {% elseif case.kind == 'HDC' and approvalNeededView %}
+        {% set releaseDate = '<span class="urgent-highlight">' + case.releaseDate
+            + '</span><span class="urgent-highlight urgent-highlight-message">HDC release</span>' %}
```
andrewrlee, line 22: mmh.. not sure if we need to do it for both approval needed and the approval given view

server/views/pages/approve/cases.test.ts
```diff
@@ -32,4 +33,19 @@ describe('View and print a licence - case list', () => {
     expect($('#nomis-id-2').text()).toBe('A1234AB')
     expect($('#release-date-2').text()).toBe('1 Sep 2022')
   })
+
+  it('should display HDC release warning label when kind is HDC', () => {
```
andrewrlee, line 37: could add another test for approval needed : false to show it not shown

andrewrlee (approved):
LGTM

server/views/pages/approve/cases.njk
louisenorris: Sorry I'm confused - assume `approvalNeededView` is the view we do want to show it in? and therefore that's what the above is doing? Is approval given view for `'/pdf-print'`? I think I need to keep `and approvalNeededView` 😅

louisenorris: Design confirmed it's needed for both approval needed and recently approved tab views so updated for this

server/views/pages/approve/cases.test.ts
louisenorris: Added another test to cover highlighting HDC cases in both approval needed and recently approved tab views.

andrewrlee (approved):
LGTM

PR #1133, opened 2024-12-31
CVSL-2340 update key dates summary list

louisenorris:


server/utils/nunjucksSetup.ts
```diff
@@ -177,6 +178,10 @@ export function registerNunjucks(app?: express.Express): Environment {
     return jsonDtToDateShort(dt)
   })
 
+  njkEnv.addFilter('datetimeToDateShortMinusDay', (dt: string) => {
```
andrewrlee, line 181: we should probably add examples of date formats in comments to each of these but it's out scope for this PR

andrewrlee: actually it'd probably be nicer to make filters more composable...

`myDate | minusDays(1) | jsonDtToDateShort`

rather than have lots of very specific filters

andrewrlee: (by the name of the filter I'd assumed it would show the date without the day)

server/views/pages/approve/keyDates.njk
```diff
@@ -2,22 +2,37 @@
...
                 value: { html: releaseDateHtml }
             }
         ]
      %}
 
+    {% if licence.kind === 'HDC' %}
+        {% set dateRows = (dateRows.push(
+                 {
+                    key: { text: "HDC end date" },
+                    value: { text: licence.conditionalReleaseDate | datetimeToDateShortMinusDay, classes: "hdc-end-date" }
```
andrewrlee, line 27: infact this should probably be calculated on the backend and part of the HDC licence response

louisenorris: Should I implement that as part of this piece of work? And then remove the filter once it's part of the HDC licence response?

andrewrlee (approved):
LGTM

PR #1139, opened 2025-01-17
CVSL-2336 add logic to change confirmation message based on kind

gurps-singh:
This PR is content change for the confirmation page based on the licence kind being HDC.

Screenshot 2025-01-17 at 12 30 17


nishanthmahasamudram (approved):
LGTM

PR #1144, opened 2025-01-24
CVSL-2329 combine time and date into a single fieldset

gurps-singh:
This follow on PR is based on demo feedback to combine the fields into a single fields for date and time instead of having two separate fieldsets.


andrewrlee (approved):
LGTM

PR #1148, opened 2025-01-30
CVSL-2337 add HDC highlight for cases in the people on probation tab

gurps-singh:
This follow on PR adds the HDC release label to cases in the CA view, where the person is on probation. 

Screenshot 2025-01-30 at 15 16 57


louisenorris (approved):
LGTM

PR #1158, opened 2025-02-11
CVSL-2423 ineligibility reason to always show and say None if missing

louisenorris:


gurps-singh (approved):

server/views/pages/support/offenderDetail.njk
```diff
@@ -262,25 +278,27 @@
         </div>
     </div>
 
-    {% if ineligibilityReasons | length %}
```
louisenorris, line 265: remove check to show only if `ineligibilityReasons` present

PR #1160, opened 2025-02-14
CVSL-2521 add new getIS91Status call to licence api client

louisenorris:


server/views/pages/support/offenderDetail.njk
```diff
@@ -201,6 +201,10 @@
                         <th scope="row" class="govuk-table__header">Recall</th>
                         <td class="govuk-table__cell">{{ prisonerDetail.recall }}</td>
                     </tr>
+                    <tr class="govuk-table__row">
+                        <th scope="row" class="govuk-table__header">IS-91 status</th>
+                        <td class="govuk-table__cell">{{ "Yes" if is91Status else "No" }}</td>
+                    </tr>
```
louisenorris, line 207: Currently displays as:
Screenshot 2025-02-14 at 15 16 11

Maybe makes more sense to set the row title to `IS-91` instead of `IS-91 status`? If the status is either Yes or No 🤔

SebNegahbanMoJ: Possibly something like `Is IS91 case?` would be more descriptive? It doesn't matter much though as this is an internally-used page, so it doesn't have to be perfect!

SebNegahbanMoJ (approved):

PR #1172, opened 2025-03-05
CVSL-2617 Fixing deletion link for exclusion zone with no map

SebNegahbanMoJ:


andrewrlee (approved):
Should have a test!

andrewrlee (approved):
LGTM

PR #1176, opened 2025-03-17
CVSL-2562 content changes for hdc licences on the current licence and licence history page

louisenorris:


server/routes/varyingLicences/handlers/timeline.ts
```diff
@@ -26,14 +24,10 @@ export default class TimelineRoutes {
...
     }
-    if (licence.statusCode === ACTIVE && !licence.isReviewNeeded && licence.kind !== LicenceKind.HDC) {
+    if (licence.statusCode === ACTIVE && !licence.isReviewNeeded) {
       return VIEW_OR_VARY
     }
 
-    if (licence.statusCode === ACTIVE && !licence.isReviewNeeded && licence.kind === LicenceKind.HDC) {
-      return VIEW
-    }
-
```
louisenorris, line 36: no longer needed as `VIEW_OR_VARY` will be used for all licences including HDC

server/views/pages/vary/timeline.njk
```diff
@@ -23,12 +24,15 @@
...
             {{ timelineCta(callToAction, licence) }}
             {{ currentLicenceEvents(timelineEvents) }}
-            {% if licence.kind === 'HDC' %}
-                {{ govukDetails({
-                    summaryText: "How do I vary this licence?",
-                    html: "Email createandvaryalicence@digital.justice.gov.uk to vary this licence.",
-                    attributes: { 'data-qa': 'hdc-vary-licence' }
-                }) }}
+
+            {% if not hdcIntegrationMvp2 %}
```
louisenorris, line 28: **question:** Don't really like the way this reads and not sure it's best implementation. I've added a feature flag specifically for checking HDC Integration work for MVP 2 is implemented. Here we want to not show `govukDetails` for `How do I vary this licence?` for MVP 2. It does show for MVP 1... what happens when we start to implement MVP3... or maybe we just address this when we come to it? 😕

server/routes/varyingLicences/handlers/timeline.ts
```diff
@@ -26,12 +27,18 @@ export default class TimelineRoutes {
     if (licence.statusCode === VARIATION_APPROVED) {
       return PRINT_TO_ACTIVATE
     }
-    if (licence.statusCode === ACTIVE && !licence.isReviewNeeded && licence.kind !== LicenceKind.HDC) {
-      return VIEW_OR_VARY
-    }
+    if (config.hdcIntegrationMvp2Enabled) {
```
SebNegahbanMoJ, line 30: I think this probably works fine, but something about it trips my brain up when I think "What happens if the flag is enabled, but the licence status is not `ACTIVE`?"
I'm pretty sure the answer to that is that it just picks up again on line 44, but that feels like an odd code flow to match an `if` condition yet not return a value from it.

It might be worth adding a couple of unit tests for "If the flag is active and isReviewNeeded is true / licence status is `VARIATION_X`" etc to make sure the other branches are covered. If they come back as passing then it's probably fine to go for now, but I'd like to look at refactoring it to be more easily followable in future. I appreciate that your change here is actually minor and it just looks larger because of indentation, but it's highlighted a bit of unnecessary complexity (in my opinion)

server/views/pages/vary/timeline.njk
SebNegahbanMoJ: I agree with you that if we model MVP3 in the same way (by adding a new feature flag) then we may need to refactor this, but I think for now we can probably progress as-is, because we're not sure what MVP3 is going to look like code-wise yet.
In the worst case, I imagine we could do something like `if not (hdcIntegrationMvp2 or hdcIntegrationMvp3)`

server/views/partials/varyCallToActions.njk
```diff
@@ -69,7 +69,7 @@
                     href: "/licence/vary/caseload",
                     attributes: { 'data-qa': 'return-to-caselist' }
                 }) }}
-
+
```
SebNegahbanMoJ, line 72: Can be ignored if there are no other changes in the PR, but presumably this can be removed

louisenorris: Oooo added as we were missing a line between these cases, just following consistency with where we have a line between each other case in the file.

server/views/pages/vary/timeline.njk
louisenorris: Cool, thanks

server/views/pages/vary/timeline.njk
```diff
@@ -10,6 +10,7 @@
 {% set pageTitle = applicationName + " - Vary a licence - licence history" %}
 {% set pageId = "view-timeline-page" %}
 {% set backLinkHref = "/licence/vary/caseload" %}
+{% set hdcIntegrationMvp2 = hdcIntegrationMvp2Enabled %}
```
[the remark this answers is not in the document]
louisenorris: hmmm was following what we do for the comms banner in `index.njk` will look into this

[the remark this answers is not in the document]
louisenorris: Yep - no need for this, removed the reassignment here and on the confirmation page, will fix the `viewActive.njk` similarly on it's own PR - thank you

server/routes/varyingLicences/handlers/timeline.ts
SebNegahbanMoJ: Had a chat with Louise about this and we agree that leaving the code as-is is the correct decision, with a key benefit being that this is easily un-pickable when we move out of MVP2 into MVP3 and beyond. Whereas a larger refactor that potentially "bakes in" the `hdcInegrationMvp2Enabled` flag would be harder to rework again later on.

andrewrlee (approved):
LGTM

PR #1179, opened 2025-03-19
CVSL-2596 adjust macro to include county and adjust names of variables

gurps-singh:
This PR is to adjust the macro which populates the curfew address having changed the representation of curfew addresses in the backend. This is to match what will be from the HDC API for curfew addresses.


server/views/pages/licence/partials/hdcCurfewTimes.njk
```diff
@@ -1,25 +1,28 @@
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
```
gurps-singh, line 2: This was using `curfewTimes` as the parameter name which was confusing me so changed this to `hdcLicenceData`

louisenorris (approved):
LGTM

PR #1186, opened 2025-03-27
CVSL-2729 adjust ordering of marker display

gurps-singh:
This PR is to fix a bug where the Early marker was being shown instead of the HDC marker. Adjusted now the HDC marker takes precedence.

### DIFF
--- a/server/views/pages/approve/keyDates.njk
+++ b/server/views/pages/approve/keyDates.njk
@@ -2,9 +2,11 @@
 
 {% macro keyDates(licence, isDueForEarlyRelease) %}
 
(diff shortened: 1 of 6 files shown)


server/views/pages/approve/keyDates.njk
```diff
@@ -2,9 +2,11 @@
 
 {% macro keyDates(licence, isDueForEarlyRelease) %}
```
gurps-singh, line 4: **note:** ‏changed this as the LSD for HDC is always set to the HDCAD in the backend so can just use LSD and then add the conditional to display the marker.

PR #1203, opened 2025-04-22
CVSL-2167

Ewan-Donovan:
https://dsdmoj.atlassian.net/browse/CVSL-2167

Previous implementations are left unaffected by these changes. Only when mvp2 flag is on, will these changes render.

The change button renders only for in_progress licences and only for correct users `user.authSource == 'delius` (Confirmed by Seb)

Confirmation pages have already been implemented by Gurps.


server/routes/approvingLicences/handlers/approvalView.ts
```diff
@@ -49,6 +49,8 @@ export default class ApprovalViewRoutes {
           licence.kind !== LicenceKind.HDC_VARIATION &&
           licence.isDueForEarlyRelease,
         hdcLicenceData,
+        isPPUser: false,
```
andrewrlee, line 52: is this template only ever used from here and for prison users? 
If so don't think we need to pass this and can hardcode it where we need it.

server/views/pages/licence/partials/hdcCurfewTimes.njk
```diff
@@ -1,10 +1,29 @@
...
+                    href: "#",
+                    text: "Change",
+                    visuallyHiddenText: "curfew hours"
+                }
+            ]
+        } %}
+        {% set tableActionHtml = '<a href="#" class="govuk-link govuk-link--no-visited-state">Change</a><span class="govuk-visually-hidden">curfew hours</span>'%}
+        {% set tableActionClasses = "govuk-table__cell govuk-!-font-weight-regular govuk-!-text-align-right"%}
+    {% else %}
+        {% set actions = null %}
```
andrewrlee, line 24: is this `actions` used anywhere?

Ewan-Donovan: It was not, the null assignment wasn't needed.

PR #1210, opened 2025-05-09
CVSL-2845 Using isVariation to handle HDC variations alongside CRD variations

SebNegahbanMoJ:


server/utils/nunjucksSetup.ts
```diff
@@ -410,6 +410,7 @@ export function registerNunjucks(app?: express.Express): Environment {
       const now = startOfDay(new Date())
       return (
         licence.kind !== LicenceKind.VARIATION &&
+        licence.kind !== LicenceKind.HDC_VARIATION &&
```
andrewrlee, line 413: `!isVariation(licence)` ?

andrewrlee (approved):
LGTM

server/utils/nunjucksSetup.ts
SebNegahbanMoJ: The `licence` here isn't actually a licence object, it's `licence: { kind: LicenceKind; hardStopWarningDate: string; hardStopDate: string }`
I took a quick look and I think it's down to the object being passed into the function not always being a full licence, so I left it as-is

PR #1211, opened 2025-05-09
CVSL-2846 Treating HDC variations as HDC licences

SebNegahbanMoJ:


server/views/pages/create/caseload.njk
```diff
@@ -31,7 +31,7 @@
         {% set comDetailsHtml = 'Unallocated' %}
     {% endif %}
 
-    {% if offender.kind === "HDC" %}
+    {% if offender.kind === "HDC" or offender.kind === "HDC_VARIATION" %}
```
nishanthmahasamudram, line 34: Can we create a nunjucks filter to validate this?
As it is used in multiple places.

SebNegahbanMoJ: If we continue with HDC licences in CVL then yes, I agree it's probably worth doing.
For now I'll leave it, as we may end up removing all of this reasonably soon anwyay.

PR #1213, opened 2025-05-14
CVSL-2857: Tidy up ACO caseload code to make moving to backend a bit …

justin-gardiner:
…easier.


server/services/lists/acoCaseloadService.ts
```diff
@@ -122,11 +157,46 @@ export default class CaseloadService {
...
-        },
+        probationPractitioner: nameToString(com.name),
       }
     })
   }
+
+  private mapAcoCaseToView(acoCase: AcoCase): AcoCaseView {
+    const licence = _.head(acoCase.licences)
+
+    const releaseDate = acoCase.nomisRecord.releaseDate
```
SebNegahbanMoJ, line 168: I think this should be reading `licence.licenceStartDate`, as the release date field from NOMIS is unreliable for edge cases

SebNegahbanMoJ: Worth noting this will also need `parseCvlDateTime` rather than ISO

server/services/lists/acoCaseloadService.ts
```diff
@@ -122,11 +157,46 @@ export default class CaseloadService {
...
+    const searchString = search?.toLowerCase().trim()
+    if (!searchString) return true
+    return (
+      acoCase.crnNumber?.toLowerCase().includes(searchString) ||
+      acoCase.name.toLowerCase().includes(searchString) ||
+      acoCase.probationPractitioner.toLowerCase().includes(searchString)
+    )
+  }
+
+  private sortByCrd(a: AcoCaseView, b: AcoCaseView): number {
```
SebNegahbanMoJ, line 197: Slight misnomer, perhaps should be `sortByReleaseDate`?
I appreciate this is a lift-and-shift, just while you're here!

justin-gardiner: yep, not sure why I put crd :)

server/services/lists/acoCaseloadService.ts
justin-gardiner: Done

server/services/lists/acoCaseloadService.ts
justin-gardiner: Done

andrewrlee (approved):
LGTM

PR #1216, opened 2025-05-15
CVSL-2874 add feature toggle

louisenorris:


server/routes/creatingLicences/handlers/hdc/confirmCreate.ts
```diff
@@ -25,6 +26,11 @@ export default class ConfirmCreateRoutes {
       this.probationService.getProbationer(nomisId),
     ])
 
+    if (config.hdcLicenceCreationBlockEnabled) {
+      logger.error('Access denied to HDC licence creation GET due HDC licence not to be breated in CVL')
```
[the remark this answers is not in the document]
SebNegahbanMoJ: Also might be worth pulling this check above the `nomisRecord` / `deliusRecord` calls above, as they'll never be used and aren't needed for the check

server/routes/creatingLicences/handlers/hdc/confirmCreate.ts
```diff
@@ -25,6 +26,11 @@ export default class ConfirmCreateRoutes {
       this.probationService.getProbationer(nomisId),
     ])
 
+    if (config.hdcLicenceCreationBlockEnabled) {
+      logger.error('Access denied to HDC licence creation GET due HDC licence not to be created in CVL')
+      return res.redirect('/access-denied')
+    }
```
andrewrlee, line 32: probably not worth changing but this could have been before the api calls as they won't be used but it doesn't really matter

andrewrlee (dismissed):
LGTM

server/routes/creatingLicences/handlers/hdc/confirmCreate.ts
louisenorris: Ah yeah, seen Seb mentioned this too above - I do agree and have moved

andrewrlee (approved):
LGTM

PR #1238, opened 2025-06-12
CVSL-2912 add tabs and empty states for CA search

gurps-singh:
This PR is to introduce the tabs for the CA search so that results can be split into people in prison and people on probation. This PR provides the empty states for both tabs should no search results be returned. 

Screenshot 2025-06-12 at 14 32 05

Screenshot 2025-06-12 at 14 32 11


nishanthmahasamudram (approved):
LGTM

PR #1239, opened 2025-06-12
CVSL-2974 move docker to alpine

andrewrlee:


bin/build-css.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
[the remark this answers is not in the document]
andrewrlee: (this will use the default shell)

package.json
```diff
@@ -145,7 +145,7 @@
   },
   "devDependencies": {
     "@ministryofjustice/eslint-config-hmpps": "^0.0.1-beta.2",
-    "@ministryofjustice/hmpps-precommit-hooks": "^0.0.1-alpha.4",
+    "@ministryofjustice/hmpps-precommit-hooks": "^0.0.3",
```
andrewrlee, line 148: This includes fixes for alpine linux that I want to test out.

Dockerfile
```diff
@@ -1,20 +1,24 @@
 # Stage: base image
-FROM node:22.12-bookworm-slim as base
-
-ARG BUILD_NUMBER
-ARG GIT_REF
-ARG GIT_BRANCH
+FROM node:22-alpine AS base
```
andrewrlee, line 2: alpine is much slimmer than bookworm

PR #1250, opened 2025-06-23
CVSL-2938: Add support pages to view the ACO caseload.

justin-gardiner:


server/data/deliusClient.ts
```diff
@@ -25,10 +25,6 @@ export default class DeliusClient extends RestClient {
     await this.put({ path: `/users/${deliusUsername}/roles` })
   }
```
justin-gardiner, line 27: not used

server/routes/approvingVariations/handlers/varyApproveList.ts
```diff
@@ -1,17 +1,19 @@
...
   GET = async (req: Request, res: Response): Promise<void> => {
     const search = req.query?.search as string
     const regionCases = req.query?.view === 'region'
     const { user } = res.locals
 
+    const { getAcoCaseloadFromBackEnd } = config
     const caseload = regionCases
-      ? await this.caseloadService.getVaryApproverCaseloadByRegion(user, search)
-      : await this.caseloadService.getVaryApproverCaseload(user, search)
+      ? await this.caseloadService.getVaryApproverCaseloadByRegion(user, search, getAcoCaseloadFromBackEnd)
```
andrewrlee, line 15: could we read this from a query parameter so we can easily do a spot the difference?

justin-gardiner: yes, we normally get it from the user details but I'm guessing the support used doesn't have a region and pdu's assigned.

PR #1251, opened 2025-06-23
Cvsl 2947 adding pathfinder question page

nishanthmahasamudram:


server/views/pages/manageConditions/pathfinder/input.njk
```diff
@@ -0,0 +1,72 @@
...
+                        label: {
+                            html: 'What is the name of the pathfinder or programme (this information will not appear on the licence)?'
+                        },
+                        errorMessage: validationErrors | findError('programmeName'),
+                        value: licence.electronicMonitoringProvider.programmeName | fillFormResponse(formResponses.programmeName)
+                    }) }}
+                {% endset %}
+                {% set radioValue = formResponses.isToBeTaggedForProgramme
+                    or (licence.electronicMonitoringProvider.isToBeTaggedForProgramme == true and 'Yes')
+                    or (licence.electronicMonitoringProvider.isToBeTaggedForProgramme == false and 'No')
```
SebNegahbanMoJ, line 30: It may be me misunderstanding this, but if I have selected 'No' previously, would `(licence.electronicMonitoringProvider.isToBeTaggedForProgramme == false and 'No')` not return `true`, thereby making `Yes` preselected if I click the `Change` link?

nishanthmahasamudram: Yes it will return 'No' for this condition (licence.electronicMonitoringProvider.isToBeTaggedForProgramme == false and 'No') if we have previously selected 'No'.

In Nunjucks, the and operator returns the second value if the first is truthy.
So, if the left side is true (i.e., the value is false), the whole expression evaluates to 'No'.
If the left side is false, the whole expression evaluates to false.

gurps-singh (approved):

AledWEvans (commented):
LGTM

PR #1256, opened 2025-06-27
CVSL-2915 add yellow highlight to ca search

louisenorris:


AledWEvans (approved):
LGTM - should this also update an integration test, to check if it goes to the expected page?

PR #1258, opened 2025-06-30
CVSL-3041 search highlight bug fix

louisenorris:


assets/js/search/highlightSearchTerm.js
```diff
@@ -17,7 +17,13 @@ window.onload = function () {
 
 function caseInsensitiveHighlighting(html, searchTerm) {
   const reg = new RegExp(searchTerm, 'gi')
-  return html.replace(reg, function (str) {
-    return `<mark>${str}</mark>`
-  })
+  // avoid changing searchTerm within href content
+  if (
+    !html.match(`/licence/create/nomisId/${searchTerm}/confirm`) &&
+    !html.match(`/licence/create/nomisId/${searchTerm}/prison-will-create-this-licence`)
```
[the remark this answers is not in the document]
louisenorris: Another possible fix could be to check whether the search term is a prisoner number using regex as prisoner numbers are only present in the caseload list when part of an href. We can search for a case using prisoner number as it is present in the html but only in hrefs

```
if (!searchTerm.match(/[A-Z]\d{4}[A-Z]{2}/gi)) {
    return html.replace(reg, function (str) {
      return `<mark>${str}</mark>`
    })
  } else return html
```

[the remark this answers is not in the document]
louisenorris: I think I'm veering towards using the prisoner number regex check

assets/js/search/highlightSearchTerm.js
```diff
@@ -10,14 +10,32 @@ window.onload = function () {
...
+  )
+  const nodesToUpdate = []
+  let node
+  while ((node = walker.nextNode())) {
+    nodesToUpdate.push(node)
+  }
+
+  for (const textNode of nodesToUpdate) {
+    const span = document.createElement('span')
+    const regEx = new RegExp(targetText, 'ig')
```
louisenorris, line 37: added this regex so the replaceAll could find target text to replace ignoring text case

server/views/pages/search/caSearch/caSearch.njk
```diff
@@ -288,7 +288,7 @@
 {% block content %}
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-three-quarters">
-            <input type="hidden" id="searchTerm" value="{{ queryTerm }}">
+            <input type="hidden" id="searchTerm" value="{{ queryTerm | titlecase}}">
```
louisenorris, line 291: added titlecase formatting so that the user input if all lowercase/uppercase is formatted and this then is displayed in the search highlight matching title case displayed on page. Without could display for example `john Smith`

assets/js/search/highlightSearchTerm.js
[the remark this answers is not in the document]
louisenorris: Updated method means this is now outdated.

server/views/pages/search/caSearch/caSearch.njk
andrewrlee: doesn't this look weird for prison numbers?

louisenorris: True, guess I just thought we don't display prison numbers on the page but maybe this could change in the future, I'll think of an alternative way to do this

louisenorris: Removed the titlecase njk filter there

PR #1276, opened 2025-07-10
CVSL-3041 bug fix to focus search highlight on specific identifiers

louisenorris:


server/views/pages/search/caSearch/caSearch.njk
```diff
@@ -94,7 +94,7 @@
     {
         text: "Release date",
         attributes: {
-            "aria-sort": "ascending"
+            "aria-sort": "none"
```
louisenorris, line 97: **note:** disabled initial sorting by release date in the table as already sorted in the backend and the table sorting is currently broken. We think the table column is trying to sort by the first characters it finds which is some cases is the label content of a date and not the date.

AledWEvans: arr that explains the odd order we were seeing

AledWEvans (approved):
LGTM

PR #1277, opened 2025-07-10
CVSL-2961 fix release date ordering

louisenorris:


server/utils/nunjucksSetup.ts
```diff
@@ -213,7 +213,7 @@ export function registerNunjucks(app?: express.Express): Environment {
   })
 
   njkEnv.addFilter('dateToUnix', (date: string) => {
-    return moment(date, 'D MMM YYYY').unix()
+    return moment(date, 'YYYY-MM-DD').unix()
```
louisenorris, line 216: date to unix accepts dates formatted as `YYYY-MM-DD`. This fixes the sorting of dates in nunjucks templates.

AledWEvans (approved):
LGTM,... any way we can test that the order is correct?

server/utils/nunjucksSetup.ts
```diff
@@ -214,7 +214,7 @@ export function registerNunjucks(app?: express.Express): Environment {
   })
 
   njkEnv.addFilter('dateToUnix', (date: string) => {
-    return moment(date, 'D MMM YYYY').unix()
+    return moment(date, ['D MMM YYYY', 'YYYY-MM-DD', 'DD/MM/YYYY']).unix()
```
louisenorris, line 217: **note:** using an array of formats to enable the function to match the input to multiple formats.

AledWEvans: interesting approach

AledWEvans (approved):
LGTM

server/utils/nunjucksSetup.ts
AledWEvans: Yes,... kind of, not sure if defensive programming is the best word to use,... but basically it's best for the caller to define,.. but it's fine... but your not being defensive,.. lol

PR #1285, opened 2025-07-16
CVSL-1268 CVL | Implement 10 and 11 digit phone numbers fix.

AledWEvans:
I have fixed the regular expression to allow a range of other numbers that old one did not support, I have put the test cases in a test so should be easy to update in the future.

### DIFF
--- a/server/routes/initialAppointment/types/telephone.test.ts
+++ b/server/routes/initialAppointment/types/telephone.test.ts
@@ -1,4 +1,4 @@
-import { validate } from 'class-validator'
+import { validate, ValidationError } from 'class-validator'
 import Telephone from './telephone'
 
 describe('Telephone validator test', () => {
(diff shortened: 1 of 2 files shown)


server/routes/initialAppointment/types/telephone.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        const errors = await validate(telephone)
+
+        return { number, errors }
+      }),
+    )
+
+    validations.forEach(({ errors }) => {
+      // Then
+      expect(errors.length).toBe(1)
+      expect(errors[0].constraints).toHaveProperty('matches')
```
nishanthmahasamudram, line 93: Better to check for exact error message what we are throwing.(Enter a phone number in the correct format, like 01632 960901)

Also add a test for **Enter a telephone number'**

PR #1288, opened 2025-07-18
CVSL-3011 added Licence type to front end, also improve code generati…

AledWEvans:
Added new licence type but also improved generation script to add single quotes rather than double and removed duplicates.

### DIFF
--- a/server/@types/licenceApiClientTypes.ts
+++ b/server/@types/licenceApiClientTypes.ts
@@ -22,6 +22,8 @@ export type Licence =
   | components['schemas']['HardStopLicence']
   | components['schemas']['CrdLicence']
   | components['schemas']['HdcLicence']
+  | components['schemas']['PrrdLicenceResponse']
+export type PrrdLicenceResponse = components['schemas']['PrrdLicenceResponse']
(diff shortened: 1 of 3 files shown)


server/@types/licenceApiClientTypes.ts
```diff
@@ -22,6 +22,8 @@ export type Licence =
   | components['schemas']['HardStopLicence']
   | components['schemas']['CrdLicence']
   | components['schemas']['HdcLicence']
+  | components['schemas']['PrrdLicenceResponse']
+export type PrrdLicenceResponse = components['schemas']['PrrdLicenceResponse']
```
nishanthmahasamudram, line 26: If we are not using it anywhere we don't need this.

nishanthmahasamudram: I get the idea — it's handy to have the structure ready for when it's needed. But just to keep things tidy, it’s probably best not to import things we’re not using yet. We can always bring them in when the time comes.

nishanthmahasamudram (dismissed):
LGTM

justin-gardiner (approved):
LGTM

PR #1300, opened 2025-07-21
CVSL-2916 fix ca search to search in a users selected prisons rather than all prisons

louisenorris:


server/routes/search/handlers/caSearch.test.ts
```diff
@@ -81,7 +83,7 @@ describe('Route Handlers - Search - Ca Search', () => {
         lastWorkedOnBy: 'Test Updater',
         isDueForEarlyRelease: false,
         isInHardStopPeriod: true,
-        prisonCode: 'BAI',
+        prisonCode: 'MDI',
```
louisenorris, line 86: Updated some prison codes to accurately reflect the prison description

server/routes/search/handlers/caSearch.ts
```diff
@@ -37,12 +38,10 @@ export default class CaSearch {
       }
     } else {
       const allPrisons = await this.prisonerService.getPrisons()
```
andrewrlee, line 40: do you need to do this api call?

louisenorris: Ah it's specifically for the caseload switcher which we don't show on the search page. I'll removed this call.

PR #1303, opened 2025-07-23
CVSL-3128 add date format filter and edit dateToUnix filter

louisenorris:


server/utils/nunjucksSetup.ts
```diff
@@ -214,11 +215,13 @@ export function registerNunjucks(app?: express.Express): Environment {
   })
 
   njkEnv.addFilter('dateToUnix', (date: string) => {
-    return moment(date, ['D MMM YYYY', 'YYYY-MM-DD', 'DD/MM/YYYY']).unix()
```
louisenorris, line 217: Removed additional date formats as should be defined based on usage rather than blanket cover and handle multiple date formats. Adding the multiple formats to this filter was a change made by myself recently but this would signal to anyone using that filter that it can handle any of those dates, possibly at any time which adds a lot of cognitive overhead. So made more specific.

server/views/pages/search/caSearch/caSearch.njk
```diff
@@ -165,7 +165,7 @@
     {
         attributes: {
             id: 'release-date-' + loop.index,
-            "data-sort-value": offender.releaseDate | dateToUnix
+            "data-sort-value": offender.releaseDate | formatCvlDateAsEurDate | dateToUnix
```
louisenorris, line 168: These two filters together `formatCvlDateAsEurDate | dateToUnix` format the `releaseDate` then convert it to a unix value to enable accurate sorting by release date on the CA and PP search screens.

server/utils/utils.ts
```diff
@@ -143,6 +143,10 @@ const toIsoDate = (date: Date) => {
   return date ? format(date, 'yyyy-MM-dd') : null
 }
 
+const formatCvlDateAsEurDate = (date: string) => {
```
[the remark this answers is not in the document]
louisenorris: Cool, yep I'll adjust, and thanks, wasn't sure on the naming

server/utils/utils.ts
```diff
@@ -50,15 +57,15 @@ const simpleDateTimeToJson = (dt: SimpleDateTime): string | undefined => {
   const { date, time } = dt
   const dateTimeString = [date.year, date.month, date.day, time.hour, time.minute, time.ampm].join(' ')
   const momentDt = moment(dateTimeString, 'YYYY MM DD hh mm a')
```
louisenorris, line 59: **question:** there are other date formats used but just once in the file, should constants be created for these too of not necessary if it's only used once?

server/utils/utils.ts
```diff
@@ -14,6 +14,13 @@ import type {
 } from '../@types/licenceApiClientTypes'
 import LicenceKind from '../enumeration/LicenceKind'
 
+const CVL_DATE = 'DD/MM/YYYY'
+const DATE_SHORT = 'D MMM YYYY'
+const ISO_DATE = 'yyyy-MM-dd'
+const JSON_DATE_TIME = 'DD/MM/YYYY HH:mm'
+const SIMPLE_DATE_TIME = 'D/MM/YYYY HHmm'
+const TWELVE_HOUR_TIME = 'hh:mm a'
```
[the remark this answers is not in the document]
andrewrlee: That's fine -  beyond the scope of this

andrewrlee (approved):
LGTM

PR #1309, opened 2025-07-29
CVSL-1888-CVL-Review-dates-and-date-labels-and-how-these-are-used-thr…

nishanthmahasamudram:
…oughout-CVL

### DIFF
--- a/server/views/partials/additionalLicenceConditionsSummary.njk
+++ b/server/views/partials/additionalLicenceConditionsSummary.njk
@@ -48,12 +48,12 @@
 
 {% macro additionalConditionHtml(condition) %}
     <div class="additional-condition-summary">
-        {{ condition.expandedText }}
+        {{ condition.expandedText | removeOrdinalSuffixes }}
     </div>
     {% if condition.data.length %}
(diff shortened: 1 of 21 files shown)


server/views/partials/additionalLicenceConditionsSummary.njk
```diff
@@ -48,12 +48,12 @@
...
 {% macro additionalConditionHtml(condition) %}
     <div class="additional-condition-summary">
-        {{ condition.expandedText }}
+        {{ condition.expandedText | removeOrdinalSuffixes }}
     </div>
     {% if condition.data.length %}
         <div class="govuk-!-font-weight-bold govuk-!-margin-top-3">
             {% for entry in condition.data | separatedDataByFieldName %}
-                <span class="govuk-!-display-block govuk-!-margin-top-2">{{ entry | formatAddress }}</span>
+                <span class="govuk-!-display-block govuk-!-margin-top-2">{{ entry | formatAddress | removeOrdinalSuffixes }}</span>
```
andrewrlee, line 56: not sure why addresses would have ordinals? and if they were then presumably they would be user defined text which we shouldn't format

PR #1313, opened 2025-07-31
CVSL-3096-add new approver search page with header and search bar

louisenorris:
This PR is to add a new search page as a starting point for using the new search pattern for the Prison Approver:

Screenshot 2025-07-31 at 11 06 18


andrewrlee (approved):
LGTM

PR #1323, opened 2025-08-07
CVSL-3004 add kind to licence table

gurps-singh:
This PR is to surface kind on the licence table in support.

Screenshot 2025-08-07 at 11 55 06


AledWEvans (approved):
LGTM

PR #1324, opened 2025-08-08
CVSL-3148 Preventing the submit licence button from showing when a case enters the hard stop period

SebNegahbanMoJ:


AledWEvans (approved):
LGTM

PR #1326, opened 2025-08-12
Cvsl 2984 cvl postcode lookup manual address entry saved addresses front end

nishanthmahasamudram:


server/views/pages/search/approverSearch/approverSearch.njk
```diff
@@ -85,7 +85,7 @@
     ), searchTableHeadings) %}
 {% endif %}
 
-{% set approvalNeededSearchTableHeadings = searchTableHeadings.toSpliced(-1, 0, {
```
louisenorris, line 88: `toSpliced` is adding the Location column to a new array

server/routes/initialAppointment/handlers/hardStop/manualAddressPostcodeLookup.test.ts
```diff
@@ -65,6 +65,7 @@ describe('Route Handlers - Create a licence - Manual address entry', () => {
           townOrCity: 'Testville',
           county: 'Testshire',
           postcode: 'TE5 7ST',
+          isPreferredAddress: '',
```
justin-gardiner, line 68: is '' a valid value, isn't is a boolean value?

justin-gardiner: ignore, I see it's a string

PR #1328, opened 2025-08-13
CVSL-3203 add location changer to prison approver search results page

louisenorris:
Location changer added to search results page which will redirect back to an updated search results page. Check all is working locally according to ACs in ticket. Finish adding unit tests and add integration test. 

Work done while working on another ticket but not in scope of that so extracted work done so far into this PR. To be picked up when ticket is ready, currently with Design.


louisenorris:
Unfinished work when first raised which has since been completed on another branch.

PR #1329, opened 2025-08-14
CVSL-2969 Removing ability to click into a case that's missing a CRN

SebNegahbanMoJ:
New display (using dummy data):
image


AledWEvans (approved):
LGTM

PR #1334, opened 2025-08-18
Cvsl 1636 remove child pages for conditions which dont have information to enter

nishanthmahasamudram:


integration_tests/integration/createLicence.cy.ts
```diff
@@ -102,6 +103,8 @@ context('Create a licence', () => {
         .nextInput()
         .enterAddress()
         .clickContinue()
+        .checkIfChangeLinkVisible()
+        .checkIfDeleteLinkVisible()
```
andrewrlee, line 107: would be nicer if we could parameterise this with the selected condition.. so it's more obvious..

if you think about the ideal way to express this test:
```
   stubConditions(
        {code: "xxxx1", requiresInput: true},
        {code: "xxxx2", requiresInput: false},
    )
   ... 

    checkIfChangeLinkVisible("xxxx1")
    checkIfDeleteLinkVisible("xxxx2")
```

PR #1336, opened 2025-08-20
CVSL-3098 content changes to the new approver search page

louisenorris:


server/views/pages/search/approverSearch/approverSearch.njk
```diff
@@ -222,9 +255,10 @@
     {
         attributes: {
             id: 'approved-on-' + loop.index,
+            "data-qa": "approvedOnDate",
             "data-sort-value": case.approvedOn | dateToUnix
         },
-        html: '<span data-qa="approvedOnDate">' + case.approvedOn  + '</span>'
+        text: case.approvedOn
```
louisenorris, line 261: **note:** changed to text to stop rendering `null` when value is not present.

server/views/pages/search/approverSearch/approverSearch.njk
```diff
@@ -47,7 +49,7 @@
     {% endif %}
 {% endmacro %}
 
-{% set searchTableHeadings = [
+{% set approvalNeededSearchTableHeadings = [
```
louisenorris, line 52: **note:** decided to make each tab table headings more static as there was a further difference needed between the two. (Approval needed cases are sorted by release date ascending by default and we require the arrow icon to be displayed so need to state this as below). Hopefully having each set set apart from location at the start makes it easier to read overall.

PR #1342, opened 2025-08-22
CVSL-3238 patching CVL

andrewrlee:


docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: '3.1'
```
andrewrlee, line 1: field deprecated so removing

package.json
```diff
@@ -40,10 +40,7 @@
   "jest": {
     "transform": {
       "^.+\\.tsx?$": [
-        "ts-jest",
-        {
-          "isolatedModules": true
```
andrewrlee, line 45: option deprecated.. I've moved to tsconfig obviously that will now apply to production files but don't think that will cause any issues. It just seems to add [additional warnings](https://www.typescriptlang.org/tsconfig/isolatedModules.html).

server/middleware/setUpWebSession.ts
```diff
@@ -1,5 +1,5 @@
 import session from 'express-session'
-import RedisStore from 'connect-redis'
+import { RedisStore } from 'connect-redis'
```
andrewrlee, line 2: breaking change of connect redis

package.json
```diff
@@ -131,20 +128,19 @@
     "moment": "^2.30.1",
     "multer": "^2.0.2",
     "nocache": "^4.0.0",
-    "npm": "^11.4.2",
```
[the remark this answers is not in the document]
justin-gardiner: whoops that may have been me, I remember upgrading to npm 11

justin-gardiner (approved):
LGTM

PR #1348, opened 2025-09-01
CVSL-3196 CVL Timeline page sometimes prioritising inactive parent li…

nishanthmahasamudram:
…cence over active variation when sorting


AledWEvans (approved):
LGTM

PR #1355, opened 2025-09-04
CVSL-3134 remove fetchDprDefinitions middleware and make it a service…

gurps-singh:
This PR is to remove the middleware which fetches the DPR definitions and retrieve the DPR definitions as a service call instead.


andrewrlee (approved):
LGTM

PR #1356, opened 2025-09-05
CVSL-3295: Pass user details when syncing COM allocation so we can re…

justin-gardiner:
…cord user in the audit event.


andrewrlee (approved):
LGTM

PR #1357, opened 2025-09-05
Cvsl 3204 changing order of results from postcode lookup

AledWEvans:


server/services/addressService.ts
```diff
@@ -5,12 +5,65 @@ import LicenceApiClient from '../data/licenceApiClient'
...
+
+  private generateAddressSortingInfo(results: AddressSearchResponse[], priorityStrings: string[]) {
+    const lowercasePriorities = priorityStrings.map(p => p.toLowerCase())
+
+    return results.map((address, index) => this.rankAddressForSorting(address, index, lowercasePriorities))
+  }
+
+  private rankAddressForSorting(address: AddressSearchResponse, index: number, priorities: string[]) {
+    const addressText = [address.firstLine, address.secondLine ?? ''].join(',').toLowerCase()
+
```
AledWEvans, line 40: have done first line and second as probabtion can turn up on the second line and I also cant find anything call Probation St, Probation Rd

PR #1358, opened 2025-09-08
CVSL-3204 Changing order of results from postcode lookup. Changed it …

AledWEvans:
had to change this as I made the assumption that the default order was alphanumeric but it was not.


andrewrlee (approved):
LGTM

PR #1361, opened 2025-09-09
Cvsl 3299 cvl fix the form responses to reload the entered values on refresh

nishanthmahasamudram:


server/utils/nunjucksSetup.ts
```diff
@@ -250,6 +262,27 @@ export function registerNunjucks(app?: express.Express): Environment {
       : undefined
   })
 
+  njkEnv.addGlobal(
+    'lookupCurfewConditionValue',
+    (additionalCondition: AdditionalCondition | AdditionalCondition[], inputName: string): string[] => {
+      if (!additionalCondition) return []
```
andrewrlee, line 268: Would be good to extract this in to a function, maybe in `/handlers/curfew`? and test it in isolation

server/routes/manageConditions/types/additionalConditionInputs/curfewTerms.ts
```diff
@@ -91,15 +91,9 @@ const splitTimes = (value: SimpleTime) => {
...
-  if (numberOfCurfews === 'One curfew') {
-    return enteredValues[enteredValues.length - 3]
-  }
-  if (numberOfCurfews === 'Two curfews') {
-    return enteredValues[enteredValues.length - 2]
-  }
-  if (numberOfCurfews === 'Three curfews') {
-    return enteredValues[enteredValues.length - 1]
-  }
+  if (numberOfCurfews === 'One curfew') return enteredValues[enteredValues.length - 3]
```
andrewrlee, line 94: is there anything we can do here to not key based on the label? is there a code or something we can pass back?

nishanthmahasamudram: No there is not code linked to it.

PR #1368, opened 2025-09-15
CVSL-3311 Producing CSV report for PRRD cases

andrewrlee:


server/views/pages/support/prrdCasesByPrison.njk
```diff
@@ -48,12 +50,13 @@
             classes: "max-width max-width-200",
             text: offender.probationPractitioner
         },
-        {
+                {
```
SebNegahbanMoJ, line 53: **nitpick:** I assume some unintentional spacing

SebNegahbanMoJ:
I had forgotten that I'd said to Simon we'd only be returning pre-release PRRDs, but there's probably no harm in returning them all (as long as we warn him)

PR #1373, opened 2025-09-16
CVSL-3300 CVL | Retrieve phone numbers where multiple for licence check

AledWEvans:
As a Probation Practitioner or Prison case admin creating a licence or editing a licence
I need the all phone numbers to appear on the check licence page
So that the numbers are visible to for me to confirm or edit.


cypress.config.ts
```diff
@@ -15,6 +15,7 @@ import prisonRegister from './integration_tests/mockApis/prisonRegister'
 import gotenbergApi from './integration_tests/mockApis/gotenberg'
 
 export default defineConfig({
+  projectId: 'ckh82p',
```
andrewrlee, line 18: mmh.. what's this for?

gurps-singh (approved):

PR #1375, opened 2025-09-17
CVSL-3225 change content for AP licence

gurps-singh:
Content change for text in AP licence.


AledWEvans (approved):
LGTM alway like IT test though,.. but may be OTT for this one

PR #1378, opened 2025-09-18
CVSL-3157 CVL | Retrieve phone numbers where multiple for print licence

AledWEvans:
As a User printing the licence 
I need the all phone numbers to appear on the relevant pages and documents
So that the numbers are visible to those who need them


gurps-singh (approved):

louisenorris (approved):
LGTM

PR #1382, opened 2025-09-18
CVSL-3157 - update after demoing to Tom, both tel inputs now show at …

AledWEvans:
After talking with tom, he thought it best to show both inputs all the time even if you click the edit alternative link or the edit tel link you will still be presented with both inputs.


integration_tests/integration/editLicence.cy.ts
```diff
@@ -44,12 +44,12 @@ context('Edit a licence before release', () => {
     })
     checkAnswersPage
       .clickChangeTelephoneLink()
-      .enterTelephone('01632960901', null)
+      .enterTelephone('01632960901', '01632960904')
```
nishanthmahasamudram, line 47: shall we add a test with either of the numbers entered?

AledWEvans: Other tests cover this for another PR that has been merged

PR #1395, opened 2025-09-23
CVSL-2916 regenerate types for ca search response and update active tab logic

louisenorris:
PR to update the CA search response type in CVL following an update to the API and also a change to the active tab logic for CA search as it previously did not consider the attention needed tab. Now attention needed cases are not shown in the in prison tab we want to make sure the attention needed tab is the active tab if it yields the most results.


server/routes/search/handlers/caSearch.ts
```diff
@@ -38,17 +38,27 @@ export default class CaSearch {
...
+
+    if (inPrisonResults.length >= onProbationResults.length || inPrisonResults.length >= attentionNeededResults.length)
+      activeTab = '#people-in-prison'
+    if (
+      attentionNeededResults.length > inPrisonResults.length &&
+      attentionNeededResults.length >= onProbationResults.length
+    )
+      activeTab = '#attention-needed'
+    if (onProbationResults.length > inPrisonResults.length && onProbationResults.length > attentionNeededResults.length)
+      activeTab = '#people-on-probation'
```
andrewrlee, line 61: **suggestion:** on co-pilot's behalf:
```ts
const lengths = {
  '#people-in-prison': inPrisonResults.length,
  '#attention-needed': attentionNeededResults.length,
  '#people-on-probation': onProbationResults.length,
};
const greatestLength = Math.max(...Object.values(lengths));

const activeTab = Object.keys(lengths).find(
  key => lengths[key] === greatestLength
);
```

AledWEvans (approved):
LGTM some nickpicks

server/routes/search/handlers/caSearch.ts
louisenorris: I was looking at using Max but does this cover that some tabs need to show if they are more than or equal to another or just if they are more than? 🤔

PR #1399, opened 2025-09-25
CVSL-3329 change hard stop event types to use common event type

gurps-singh:
This PR is to change the instances of HARD_STOP event types to the common event types now they have been removed from the backend.


andrewrlee (approved):
LGTM

PR #1405, opened 2025-09-25
CVSL-1679 Adding time served cases page

andrewrlee:


server/@types/licenceApiImport/index.d.ts
```diff
@@ -5653,8 +5653,6 @@ export interface components {
         | 'SUPERSEDED'
         | 'HARD_STOP_CREATED'
         | 'HARD_STOP_SUBMITTED'
-        | 'HARD_STOP_REVIEWED_WITHOUT_VARIATION'
-        | 'HARD_STOP_REVIEWED_WITH_VARIATION'
```
andrewrlee, line 5657: removing manual modification but thankfully nothing was using it.

server/data/licenceApiClient.ts
```diff
@@ -628,6 +629,15 @@ export default class LicenceApiClient extends RestClient {
     )) as Promise<CaCase[]>
   }
 
+  async getTimeServedCases(prisonCode: string, user?: User): Promise<TimeServedCaseload> {
```
nishanthmahasamudram, line 632: Are we not adding tests for this like how we do earlier?
Ex:   ```
it('should get prisoner details', async () => {
    const username = 'admin-user'
    await licenceApiClient.getPrisonerDetail('G4169UO', { username } as User)
    expect(get).toHaveBeenCalledWith({ path: '/prisoner-search/nomisid/G4169UO' }, { username })
  })
```

**Review by nishanthmahasamudram — APPROVED** [review]
LGTM

File: server/data/licenceApiClient.ts (hunk 2 above)
Comment (andrewrlee) [reply server/data/licenceApiClient.ts:632]: Thanks. I've added this

File: server/routes/support/handlers/timeServedCasesByPrison.ts [hunk 3]
```diff
@@ -0,0 +1,128 @@
...
+    return res.render('pages/support/timeServedCasesByPrison', { caseload, user, prisons, prisonCode })
+  }
+
+  GET_CSV = async (req: Request, res: Response): Promise<void> => {
+    const { user } = res.locals
+    const { prisonCode } = req.params || {}
+
+    const caseload = await this.getCases(user, prisonCode)
+
+    const header = [
```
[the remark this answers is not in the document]
andrewrlee: not sure the extra level of abstraction would be useful  and this is temporary so optimising for deletion

server/views/pages/support/timeServedCasesByPrison.njk
```diff
@@ -0,0 +1,305 @@
...
+
+    {% if otherCases.length > 0 %}
+        <h2>Other upcoming releases</h2>
+        {{ govukTable({
+            attributes: {
+                'data-module': 'moj-sortable-table'
+            },
+            caption: "Other upcoming releases",
+            captionClasses: "govuk-visually-hidden",
+            head: [
```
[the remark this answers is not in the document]
andrewrlee: Thanks Nishanth, these are two separate tables so have different reasons for change so don't think it makes sense to merge. Also conditionally adding/removing is a bit awkward with nunjucks and not sure its worth the extra complexity

PR #1416, opened 2025-09-30
Cvsl 3348 cvl case admin navigation improvements

nishanthmahasamudram:


integration_tests/integration/viewAndPrintLicence.cy.ts
```diff
@@ -320,4 +320,19 @@ context('View and print licence', () => {
     const viewLicencePage = viewCasesList.clickALicence()
     viewLicencePage.checkIfElectronicMonitoringAdditionalInformationDoesNotExist()
   })
+
+  it('Should navigate back to the "Future Release" tab when clicking the "Back" link', () => {
```
louisenorris, line 324: Should this be called `Should navigate back to the current active tab when clicking the "Back" link` as that's the behaviour we want? I read this as it will always go back to the "Future Release" tab but I think we only want that if it's the active tab?

PR #1418, opened 2025-09-30
CVSL-3197: On the timeline page, if a user attempts to access a older…

justin-gardiner:
… version of the licence linked to a different team check if they should be able view it.


andrewrlee (approved):
LGTM

PR #1420, opened 2025-09-30
POC to prevent execution of arbitrary scripts

andrewrlee:
The idea is that we'd wrap this in a package which could initialised a project via `npx @ministryofjustice/hmpps-npm-script-locker` similar to other packages. 

## How it works:
This sets some sensible defaults in .npmrc to prevent execution of scripts and use of exact versions in `package.json` (thought unfortunately not `package-lock.json`)

`npm install` install packages but doesn't not execute any scripts
`npm run setup` runs `npm ci` (without scripts) and then this package to execute allowlisted scripts 

So requires a bit of mind bending to stop using npm install or npm ci  

This isn't a 100% sticking plaster over all of npms foibles, but a possible extra level of defence in depth
Not pitching it as an alternative to developing in containers, ditching node and moving to deno, golang, erlang, pascal etc.

## Doesn't lavamoat do the same thing?

This is heavily inspired by [lavamoat's allowscripts](https://github.com/LavaMoat/LavaMoat/tree/main/packages/allow-scripts).

This works slightly differently: 
* Scripts need to be explicitly configured either to be included or not or the script will fail
* This uses the explicit version in the package-lock.json to key whether a script should run or not, rather than just the name of the package

## Use cases
There's a few different use cases this needs to support: 

* A developer checks out a project and wants to set it up ✅   
  - `npm run setup`
* A developer wants to patch ✅  
  - `(npm update` or `npm audit fix` - or  check-updates or whatevs ) && `npm run setup`
* A developer wants to install a new package ✅ 
  - `npm install X` && `npm run setup`
* CI builds image ✅  
  - `npm run setup`
* A bot (dependabot/ renovate) applies updates ⚠️ 

## TODO

As well as creating the package will require some changes to build script to accomodate

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,13 @@
+# This provides sensible defaults, this can be customised if necessary and changes will not be overwritten
+
+# Prevent preinstall, install, postinstall scripts from running by default
+ignore-scripts = true
+
+# Use exact versions in package.json when using npm install <pkg>
+save-exact = true
+
+# Fail if trying to use incorrect version of npm
+engine-strict = true
+
+# Show any output of scripts in the console 
+foreground-scripts = true
--- a/hmpps-npm-script-locker/hmpps-unlock-scripts.mjs
+++ b/hmpps-npm-script-locker/hmpps-unlock-scripts.mjs
@@ -1,0 +1,56 @@
+#!/usr/bin/env node
+
+import fs from 'fs'
+import npmRunScript from '@npmcli/run-script'
+
+const CONFIG_LOCATION = '../.allowed-scripts.mjs'
+const packageLockJson = JSON.parse(fs.readFileSync('package-lock.json', 'utf-8'))
+
+console.log(`Gathering configuration information from ${CONFIG_LOCATION}\n`)
+
+const configuredAllowlistModule = await import(CONFIG_LOCATION)
+const configuredAllowlist = Object.entries(Object.assign({}, configuredAllowlistModule).default)
+
+const packagesWithScripts = Object.entries(packageLockJson.packages)
+  .map(([name, { version, hasInstallScript }]) => ({ name, version, hasInstallScript }))
+  .filter(pkg => pkg.hasInstallScript)
+  .map(pkg => ({ ...pkg, nameWithVersion: `${pkg.name}@${pkg.version}` }))
+  .map(pkg => {
+    const configuredPackage = configuredAllowlist.find(([name]) => pkg.nameWithVersion === name)
+    const isConfigured = Boolean(configuredPackage)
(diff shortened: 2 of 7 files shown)


.npmrc
```diff
@@ -0,0 +1,13 @@
...
+ignore-scripts = true
+
+# Use exact versions in package.json when using npm install <pkg>
+save-exact = true
+
+# Fail if trying to use incorrect version of npm
+engine-strict = true
+
+# Show any output of scripts in the console 
+foreground-scripts = true
```
andrewrlee, line 13: set sensible defaults for any developer of project

package.json
```diff
@@ -31,7 +31,8 @@
     "rebuild": "npm run clean && npm i && npm run build",
     "precommit:secrets": "gitleaks git --pre-commit --redact --staged --verbose --config .gitleaks/config.toml",
     "precommit:lint": "node_modules/.bin/lint-staged",
-    "precommit:verify": "npm run typecheck"
+    "precommit:verify": "npm run typecheck",
+    "setup": "npm ci && node run hmpps-allow-scripts"
```
andrewrlee, line 35: developers and ci would need to run this instead of the npm install in order to ensure appropriate scripts are installed.
If they forget then new dependencies would be pulled in but the .npmrc config should ensure no new dependencies are run

hmpps-npm-script-locker/index.mjs
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Provides information as to whether a script should be enabled or not.
+ *
+ * @param {Scripts} config
+ * @return {Scripts} an array of eslint config objects
+ */
+export default function configureAllowedScripts(config) {
+  return config
+}
```
andrewrlee, line 23: provides some type safety/autocomplete for config

hmpps-npm-script-locker/hmpps-unlock-scripts.mjs
```diff
@@ -0,0 +1,56 @@
...
+
+const configToPrint = Object.fromEntries(packagesWithScripts.map(pkg => [pkg.nameWithVersion, pkg.status]))
+
+console.log(`Current configuration: ${JSON.stringify(configToPrint, null, 2)}\n`)
+
+const missingPackages = packagesWithScripts.filter(pkg => !pkg.configured).map(pkg => pkg.nameWithVersion)
+if (missingPackages.length) {
+  console.log(`ERROR: Explicitly configure the following in ${CONFIG_LOCATION}:\n  ${missingPackages.join('\n  ')}`)
+  process.exit(1)
+}
```
andrewrlee, line 32: all packages with scripts need to be included for anything to run.. we could also do the opposite check - don't run if the config references package/versions that no longer exit 🤷

PR #1425, opened 2025-10-02
CVSL-3353: On the confirm licence creation page always display the su…

justin-gardiner:
…mmary box instead of radio buttons.


louisenorris (approved):
LGTM

PR #1736, opened 2026-04-15
CVSL-3993 Prog HDC Add Change link to update Curfew details in Check …

nishanthmahasamudram:
Update first night curfew hours and individual or same curfew hours from check your answers page for HDC licences.


server/views/pages/licence/partials/nonUniformCurfewTimesTable.njk
```diff
@@ -26,7 +26,7 @@
       head: [
         { text: "Start time", classes: "govuk-!-width-one-quarter" },
         { text: "End time",  classes: "govuk-!-width-one-quarter" },
-        { text: "",          classes: "govuk-!-width-one-half" }
+        { html: "<a class='govuk-link remove-bold' href='" + baseURL + "/standard-curfew-hours-question" + routeAppend + "'>Change</a>" if isEditable else "", classes: "govuk-!-width-one-half text-align-right" }
```
andrewrlee, line 29: do we need to escape the base url?  what about the routeAppend?

nishanthmahasamudram: **{% set baseURL = "/licence/create/id/" + licence.id + "/hdc" %}
{% set routeAppend = '?fromReview=true' if not isPrisonUser %}**

BaseURL doesn’t need escaping as it’s app‑defined and built from static paths + licence.id (server‑side, not user input).
routeAppend is also hard‑coded ('?fromReview=true') and not derived from user input.
Because both values are fully controlled by the application, they’re safe to use directly in the href. Escaping them wouldn’t add any security benefit here.

andrewrlee (approved):
LGTM

PR #1748, opened 2026-04-29
CVSL-4038 Package updates

andrewrlee:


audit-ci.json
```diff
@@ -3,7 +3,8 @@
   "package-manager": "auto",
   "registry": "https://registry.npmjs.org",
   "allowlist": [
-    "GHSA-5c6j-r48x-rmvq|mocha>serialize-javascript",
```
andrewrlee, line 6: this has now been resolved

audit-ci.json
```diff
@@ -3,7 +3,8 @@
...
-    "GHSA-73rr-hh4g-fpgx|mocha>diff"
+    "GHSA-73rr-hh4g-fpgx|mocha>diff",
+    "GHSA-w5hq-g745-h8pq|cypress-axe>cypress>@cypress/request>uuid",
+    "GHSA-w5hq-g745-h8pq|jest-junit>uuid"
```
andrewrlee, line 8: allowing these as false positives as they're only used by test dependencies. 
There's one remaining moderate issue which we're not able to fix yet which I've left in so we'll stick get warnings for.  
(there is a patch for it but we're still in the cooldown period)

server/listeners/eventHandlers/domainEvents/index.ts
```diff
@@ -30,5 +30,6 @@ export default function buildEventHandler({ licenceService, prisonerService }: S
         }
       }
     })
+    return messages
```
andrewrlee, line 33: The interface has changed - you now return the list of messages you want to acknowledge. 
This should keep the existing logic - acknowledge all the messages by default

package.json
```diff
@@ -160,26 +160,27 @@
...
     "jest-junit": "^16.0.0",
     "jsonwebtoken": "^9.0.3",
-    "lint-staged": "^16.3.3",
+    "lint-staged": "^16.4.0",
     "mocha": "^11.7.5",
     "mocha-junit-reporter": "^2.2.1",
-    "nock": "^14.0.11",
+    "nock": "^14.0.13",
     "nodemon": "^3.1.14",
+    "npm-check-updates": "21.0.3",
```
andrewrlee, line 178: added this as dev dependency so we can manage it's version explicitly and it should pick up the version when running via `npx`

dave-virden (approved):
LGTM

PR #1786, opened 2026-05-29
CVSL-3948: Updates for drinking establishment condition in policy V4.

justin-gardiner:


server/services/conditionService.ts
```diff
@@ -81,19 +81,6 @@ export default class ConditionService {
     return (await this.getLicencePolicy()).version
   }
 
-  async getStandardConditions(
```
justin-gardiner, line 84: only used in tests

andrewrlee (approved):
LGTM

PR #1790, opened 2026-06-02
Update shell quote to 1.8.4

justin-gardiner:


justin-gardiner:
not the right approach

PR #1793, opened 2026-06-03
CVSL-4110 Adding Curfew address variation manual entry screen

SebNegahbanMoJ:
Currently has no effect other than redirect on POST as the receiving endpoint is under construction


andrewrlee (approved):
LGTM

PR #1797, opened 2026-06-08
CVSL-4210 add suppression

gurps-singh:
This PR is to add a suppression for the crit vulnerability in shell quote. This is bought over by DPR and a separate issue has been raised to upgrade the vulnerable dependency.


justin-gardiner (approved):
LGTM

PR #1799, opened 2026-06-08
CVSL-4172 update mobile phone ban condition

gurps-singh:
This PR is to make an update on the mobile phone ban condition for version 4 of the licence conditions. This only impacts the mock v4 policy which has been updated to reflect the change. 

Screenshot 2026-06-08 at 15 18 45


dave-virden (approved):
LGTM

PR #1804, opened 2026-06-10
CVSL-3947: Remove extra wrapping button group div.

dave-virden:
There was an extra (I believe) button group which was causing shorter text bodies to be wrapped shorter than designed/desired, see below (The button group was on L36 as it's not immediately clear from the diff)

Screenshot 2026-06-10 at 09 29 29

And when it's removed it renders in the same way as the prototype

Screenshot 2026-06-10 at 09 26 55


justin-gardiner (approved):
LGTM

PR #1807, opened 2026-06-11
CVSL-4076 Check your answer variation page

nishanthmahasamudram:


server/@types/licenceApiImport/index.d.ts
```diff
@@ -2873,8 +2873,7 @@ export interface components {
     }
     /** @description Request to add HDC curfew address with post-release checks */
     AddHdcCurfewAddressRequest: {
-      /** @description The address to be added as the HDC curfew address */
-      address: components['schemas']['AddAddressRequest']
+      address?: components['schemas']['AddAddressRequest'] | null
```
AledWEvans, line 2876: why is this optional?

nishanthmahasamudram: The address is optional because it may not always be provided at the time of request creation.

server/views/partials/variationSummary.njk
```diff
@@ -1,6 +1,11 @@
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
 
 {% macro variationSummary(licence, isEditable) %}
+
+    {% set isResidential = licence.curfewAddress and licence.curfewAddress.accommodationType == "RESIDENTIAL" %}
+    {% set checksCompleted = licence.curfewAddress.postReleaseResidentialChecksCompleted %}
```
AledWEvans, line 6: so this is use for all types of licence,... and you use the existence of licence.curfewAddress to safe guard.
But how do you protect the other line I would have thought you needed
`licence.curfewAddress and licence.curfewAddress.postReleaseResidentialChecksCompleted`

nishanthmahasamudram: No, this is only for HDC_Variation and curfew address will exists only for this.

AledWEvans (approved):
LGTM

PR #1810, opened 2026-06-15
CVSL-4198 fix type safety of Licence entries

andrewrlee:
This must have been broken for a while! 😱 

I've updated the scripts to automatically remove the extra union on licence which causes an infinite loops resulting in typescript resolving the `Licence` type to `any`.

I've also added some tests to ensure that if this breaks in the future it's detected.

This has unsurfaced quite a few type errors - to help address those I've added a couple of assertion functions. 

```ts
assertIsHdcLicence(licence)
// licence.actualHdcDate etc can now be referenced
```

```ts
assertIsVariation(licence)
// licence.variationOfId etc can now be referenced
```

If an incorrect licence type is passed in then it will throw an error.


renovate.json
```diff
@@ -1,14 +1,6 @@
...
-  "extends": ["github>ministryofjustice/hmpps-renovate-config:node", ":disableDependencyDashboard"],
-  "enabledManagers": ["npm", "nvm", "dockerfile", "docker-compose", "helmv3", "helm-values", "github-actions"],
-  "printConfig": true,
-  "timezone": "Europe/London",
-  "schedule": ["before 9am on the first day of the month", "after 4pm on the 20th day of the month"],
-  "rebaseWhen": "behind-base-branch",
-  "prCreation": "not-pending",
-  "branchPrefix": "renovate-",
-  "rollbackPrs": true,
+  "extends": ["github>ministryofjustice/hmpps-renovate-config:node"],
```
andrewrlee, line 3: I've just sync'd this with the typescript template version.

package.json
```diff
@@ -33,7 +33,7 @@
   },
   "engines": {
     "node": "^24",
-    "npm": "^11"
+    "npm": "^11.10"
```
andrewrlee, line 36: update this to the min version that supports cooldown version - to ensure everyone is using an up to date/safe version of npm

server/utils/utils.ts
```diff
@@ -295,6 +296,20 @@ function isHdcLicence(licence: Licence): licence is HdcLicence | HdcVariationLic
...
+    'Licence must be HDC or HDC Variation',
+  )
+}
+
+function assertIsVariation(licence: Licence): asserts licence is VariationLicence | HdcVariationLicence {
+  assert(
+    licence.kind === LicenceKind.VARIATION || licence.kind === LicenceKind.HDC_VARIATION,
+    'Licence must be Variation or HDC Variation',
+  )
+}
```
andrewrlee, line 311: new assertion functions

generate-licence-api-types.sh
```diff
@@ -36,6 +36,18 @@ node - <<EOF
...
 
-  fs.writeFileSync(file, cleanedFileWithOutDuplicates);
+  // Remove the discriminated union appended to the Licence schema by openapi-typescript.
+  // It creates a circular type reference (Licence -> subtypes -> Licence) which causes
+  // TypeScript to resolve the first union member (PrrdLicenceResponse) to `any`.
+  // The union is redundant because each subtype already extends Licence directly.
+  const cleanedLicenceUnion = cleanedFileWithOutDuplicates.replace(
+    / & \(\s*(?:\|\s*components\['schemas'\]\['[^']+'\]\s*)+\)/g,
+    ''
+  );
```
andrewrlee, line 46: removed dodgy extra union which causes a cyclical dependency

nishanthmahasamudram (approved):
LGTM

PR #1839, opened 2026-06-22
CVSL-4244 tweak margins for curfew times

gurps-singh:
This PR is to amend some of the margins added in a previous PR and instead use the SCSS file for time restrictions to add some custom rules. 

Screenshot 2026-06-22 at 16 26 10

Screenshot 2026-06-22 at 16 35 43


justin-gardiner (approved):
LGTM

PR #1850, opened 2026-06-25
Cvsl 4177 add hdc details to pdu head summary page

nishanthmahasamudram:


server/views/partials/variationDetails.njk
```diff
@@ -1,6 +1,9 @@
 {% from "customComponents/summaryCard/macro.njk" import summaryCard %}
 
 {% macro variationDetails(licence, conversation) %}
+    {% set isResidential = licence.curfewAddress and licence.curfewAddress.accommodationType == "RESIDENTIAL" %}
+    {% set checksCompleted = licence.curfewAddress.postReleaseResidentialChecksCompleted %}
```
SebNegahbanMoJ, line 5: I'm kind of in two minds about whether these should be defined here or defined in the handler and passed through, but that would also be quite a lot of information to pass through when it will be false for the majority (ie non-HDC) licences 🤔 

I think it's probably fine as-is, we can always lift it out in future if needed.

SebNegahbanMoJ (approved):
Approving this PR because I think it's fine as-is, but did we need to show the Residential vs CAS selection here too?

PR #1851, opened 2026-06-25
CVSL-4129 Adding HDC changes to variation summary page

SebNegahbanMoJ:
Also removes `spy`s from `licenceService.test.ts`


nishanthmahasamudram (approved):
LGTM

PR #1853, opened 2026-06-25
CVSL-4083 add missing release date labels

gurps-singh:
This PR is to add some missing release date labels for the COM search. This already exists as a default but there are some instances where release date label is not present. 

Screenshot 2026-06-25 at 12 11 36


nishanthmahasamudram (approved):
LGTM

PR #1856, opened 2026-06-26
CVSL-4266 Adding HDC changes to view submitted variation page

SebNegahbanMoJ:
Also renames some fields of varyRefer, which already have the necessary changes for HDC fields to appear


andrewrlee (approved):
LGTM

PR #1863, opened 2026-06-29
CVSL-4067 remove pss only text

gurps-singh:
This PR is to remove dynamic text from the code for the OMU “Are you sure you want to generate…” screen. As a macro is used, this will also remove the text from other screens which may also be covered in CVSL-4065

Before
Screenshot 2026-06-29 at 13 11 42

Screenshot 2026-06-29 at 13 13 58

After
Screenshot 2026-06-29 at 13 12 07

Screenshot 2026-06-29 at 13 13 29


andrewrlee (approved):
LGTM

PR #1867, opened 2026-07-01
CVSL-4295: Fix licence printing issue for event maps

dave-virden:
There is further work to fix this for V4 which will probably require some design input from UCD


andrewrlee (approved):
LGTM

andrewrlee (changes requested):
ahh

andrewrlee (approved):
LGTM

PR #1872, opened 2026-07-02
CVSL-4302 Prog HDC CVL Removing red highlight from PP vary case list …

nishanthmahasamudram:
…and showing the HDCAD release date label


andrewrlee (approved):
LGTM

PR #1874, opened 2026-07-03
CVSL-4169 Move restriction zones to after exclusion zones.

dave-virden:
During testing it was observed that if a licence had a singular restriction zone and multiple exclusion zones the restriction zone was being displayed before the exclusion zones (as the singular conditions are grouped and displayed first). UCD asked if was possible if the restriction zone could always be displayed after the exclusion zones.

image

See above with two exclusions zones displayed above a single restriction zone


justin-gardiner (approved):
LGTM

PR #1879, opened 2026-07-06
CVSL-4066 remove text from OMU screen

gurps-singh:
This PR is to remove some text from the OMU screens.

Screenshot 2026-07-06 at 15 37 27


justin-gardiner (approved):
LGTM

PR #1881, opened 2026-07-06
CVSL-4066 remove logic for pss

gurps-singh:
This PR is to remove some left over logic from previous PR's as suggested by PR feedback.

### DIFF
--- a/server/views/pages/vary/viewActive.njk
+++ b/server/views/pages/vary/viewActive.njk
@@ -50,7 +50,7 @@
 
     <div class="govuk-grid-row govuk-!-margin-bottom-3">
         <div class="govuk-grid-column-full">
-            {% if ( licence.typeCode === 'AP' or licence.typeCode === 'AP_PSS')
+            {% if ( licence.typeCode === 'AP')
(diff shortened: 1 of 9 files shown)


server/views/pages/vary/viewActive.njk
```diff
@@ -50,7 +50,7 @@
 
     <div class="govuk-grid-row govuk-!-margin-bottom-3">
         <div class="govuk-grid-column-full">
-            {% if ( licence.typeCode === 'AP' or licence.typeCode === 'AP_PSS')
+            {% if ( licence.typeCode === 'AP')
```
andrewrlee: it will always be an AP licence now and never in its PSS period

server/views/pages/vary/viewActive.njk
andrewrlee: this will never be true anymore

gurps-singh: What happens here if it is a historic AP_PSS licence which is varied for the first time? 
Assume  this then will just become a AP one upon doing the variation?

andrewrlee: If it's active it will have already been migrated to AP I believe.. all of our licences that are not INACTIVE will be AP

andrewrlee (approved):
LGTM

PR #1894, opened 2026-07-09
Cvsl 4338 hide residential check questions if they are not answered

nishanthmahasamudram:


server/views/partials/variationSummary.njk
```diff
@@ -65,7 +65,7 @@
                         visuallyHiddenText: "Reason address checks not completed"
                     }]
                 } if isEditable
-            } if isResidential and checksCompleted == false
+            } if isResidential and (checksCompleted == true or checksCompleted == false) and checksCompleted == false
```
andrewrlee: this doesn't look right? 

as written it could be simplified to:
```ts
if isResidential and checksCompleted == false
```
which is what it was? 

can't it just be `checksCompleted === false` ?

nishanthmahasamudram: It will be null by default and if answered it will be true/false.

andrewrlee (approved):
LGTM

PR #1916, opened 2026-07-21
CVSL-4350 populate info banner

gurps-singh:
This PR is to display a banner in CVL which informs users of upcoming changes with licence conditions

Screenshot 2026-07-21 at 15 28 03


dave-virden (approved):
LGTM

PR #1922, opened 2026-07-23
CVSL-4350 change content for comms banner for probation user

gurps-singh:
This PR is to adjust text for the comms banner for upcoming licence condition changes for probation users. 

Screenshot 2026-07-23 at 10 36 00


justin-gardiner (approved):
LGTM

PR #1923, opened 2026-07-23
CVSL-4255 Add new routes for no appointment needed.

dave-virden:
This is a prerequisite for sorting out the check your answer page when no appointment needed is an option.


andrewrlee (approved):
LGTM

PR #1933, opened 2026-07-28
CVSL-4282 update contact number page for omu

gurps-singh:
This PR is to update our existing contact number page so that if an OMU has selected no appointment is needed, we can adjust the contents accordingly. 

Screenshot 2026-07-27 at 17 15 32


AledWEvans (approved):
LGTM

PR #1946, opened 2026-07-29
CVSL-4343 remove deprecated comName and comStaffCode references

louisenorris:


server/views/pages/search/probationSearch/probationSearch.njk
```diff
@@ -18,7 +18,7 @@
       {% else %}
         {% set link = offender | createOffenderLink %}
       {% endif %}
-      {% if offender.probationPractitioner.name or offender.licenceStatus === 'TIMED_OUT' or  offender.kind == "TIME_SERVED" or offender.isOnProbation %}
+      {% if offender.probationPractitioner.name or offender.licenceStatus === 'TIMED_OUT' or offender.kind == "TIME_SERVED" or offender.isOnProbation %}
```
louisenorris, line 21: **question:** Is this a bug? The test name associated with the below reads `should render offender name link when licence is TIMED_OUT and comName is missing` therefore should this actually be `offender.probationPractitioner.name === 'Not allocated' and offender.licenceStatus === 'TIMED_OUT' or offender.kind == "TIME_SERVED" or offender.isOnProbation`

server/views/pages/search/probationSearch/probationSearch.test.ts
```diff
@@ -830,7 +796,7 @@ describe('View Probation Search Results', () => {
     expect($('.govuk-tabs__list a').text()).toContain('People on probation (2 results)')
   })
 
-  it('should render offender name link when licence is TIMED_OUT and comName is missing, kind is not Time served or on prisoner is not on probation', () => {
```
louisenorris, line 833: wording here makes me think there may be a bug in the template code... should the updated name for this be `should render offender name link when licence is TIMED_OUT and probation practitioner name is Not allocated, or kind is not Time served or prisoner is not on probation`?

server/views/pages/search/probationSearch/probationSearch.njk
andrewrlee: think this is an issue - assume in the past when the COM was unallocated they didn't have a name. 

We wouldn't want to do a string comparison though as this is brittle, instead we'd use the `allocated` boolean.

andrewrlee: would you be ok to raise a bug ticket?

andrewrlee (approved):
LGTM

server/views/pages/search/probationSearch/probationSearch.njk
louisenorris: Sure, just to check - would the right logic be if allocated false **and** status TIMED_OUT then keeping the ors for the rest of the line or allocated false or status TIMED_OUT or etc..?

louisenorris: From the below test title I'm assuming the former 😅

PR #1955, opened 2026-07-30
CVSL-4394 update contact address and number urls for no appointment O…

gurps-singh:
…MU licences

### DIFF
--- a/server/routes/initialAppointment/handlers/prisonCreated/hardStop/manualAddressPostcodeLookup.ts
+++ b/server/routes/initialAppointment/handlers/prisonCreated/hardStop/manualAddressPostcodeLookup.ts
@@ -11,11 +11,15 @@ export default class ManualAddressPostcodeLookupRoutes {
 
   GET = async (req: Request, res: Response): Promise<void> => {
     const { licenceId } = req.params
+    const noAppointmentNeeded = res.locals.licence.appointmentPersonType === 'NO_APPOINTMENT_NEEDED'
+
     const action = this.path === PathType.EDIT ? 'edit' : 'create'
     const basePath = `/licence/hard-stop/${action}/id/${licenceId}`
     res.render('pages/initialAppointment/prisonCreated/manualAddressPostcodeLookupForm', {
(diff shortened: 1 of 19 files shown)


server/routes/initialAppointment/handlers/prisonCreated/hardStop/manualAddressPostcodeLookup.ts
```diff
@@ -11,11 +11,15 @@ export default class ManualAddressPostcodeLookupRoutes {
 
   GET = async (req: Request, res: Response): Promise<void> => {
     const { licenceId } = req.params
```
gurps-singh, line 13: **note:** ‏Through testing, found out that if a manual address is selected, the `find an address` link defaulted to the `initial-contact-address` route so made changes to this too.

server/routes/viewingLicences/handlers/viewLicence.ts
```diff
@@ -17,6 +17,7 @@ export default class ViewAndPrintLicenceRoutes {
   GET = async (req: Request, res: Response): Promise<void> => {
     const { licence, user } = res.locals
     let warningMessage
```
gurps-singh, line 19: **note:** ‏Testing showed that the change links were going to the old routes rather than the new no appointment URL's so made this change.

server/views/partials/prisonCreated/inductionMeetingDetailsSummary.njk
```diff
@@ -11,7 +11,14 @@
     } %}
 
     {% set routepath = pathMap[licence.kind] + licence.id %}
-    {% set addressPath = '/manual-address-entry' if (licence.licenceAppointmentAddress and licence.licenceAppointmentAddress.source == 'MANUAL') else '/initial-meeting-place' %}
```
gurps-singh, line 14: **note:** ‏As above, this changes the behaviour of the change links on the OMU check your answer page so that if no appointment is selected, it will now route to the correct URL rather than the default initial-meeting-place route.

dave-virden (approved):

PR #1959, opened 2026-07-31
CVSL-4398 offender name link com allocated bug fix

louisenorris:


server/views/pages/search/probationSearch/probationSearch.test.ts
```diff
@@ -796,7 +796,7 @@ describe('View Probation Search Results', () => {
     expect($('.govuk-tabs__list a').text()).toContain('People on probation (2 results)')
   })
 
-  it('should render offender name link when licence is TIMED_OUT and probation practitioner name is missing, kind is not Time served or on prisoner is not on probation', () => {
+  it('should render offender name link when licence is TIMED_OUT and probation practitioner is not allocated, kind is not Time served and prisoner is not on probation', () => {
```
louisenorris, line 799: Test is testing the `TIMED_OUT` part of the logic but rewording following the bug fix.

server/views/pages/search/probationSearch/probationSearch.njk
```diff
@@ -18,7 +18,7 @@
       {% else %}
         {% set link = offender | createOffenderLink %}
       {% endif %}
-      {% if offender.probationPractitioner.name or offender.licenceStatus === 'TIMED_OUT' or offender.kind == "TIME_SERVED" or offender.isOnProbation %}
```
louisenorris, line 21: Following past work done to return a Probation Practitioner object for all cases which includes the name `Not allocated` if a PP is missing there will now always be a name present so we need to check the allocated boolean on this object instead to see if a PP is present.

gurps-singh (approved):

PR #1963, opened 2026-08-04
CVSL-4418 Get appointment date/time when requiring an appointment.

dave-virden:


server/routes/initialAppointment/handlers/initialMeetingName.test.ts
```diff
@@ -146,6 +146,18 @@ describe('Route Handlers - Create Licence - Initial Meeting Name - Probation use
...
+        res.locals.licence.appointmentPersonType = 'NO_APPOINTMENT_NEEDED'
+        req.body.appointmentPersonType = 'CUSTOM_PERSON'
+        req.query.fromReview = 'true'
+
+        // When
+        await handler.POST(req, res)
+
+        expect(res.redirect).toHaveBeenCalledWith('/licence/create/id/1/initial-meeting-time?fromReview=true')
+      })
+
```
AledWEvans, line 160: Would be good to get a test if fromReview is false or undefined or null, your call

AledWEvans (approved):
LGTM added a comment

server/routes/initialAppointment/handlers/initialMeetingName.test.ts
dave-virden: That's the standard flow and covered by the other tests/integration tests

